PR #40, opened 2023-01-06
Add draft of the SDK Rustdoc

### LINKED ISSUE #20: Document public interfaces for `rustdoc`
The rustdocs can be generated with `cargo doc --no-deps --open`.

This will build the docs locally and open them in a browser.

ok300:
The rustdocs can be generated with `cargo doc --no-deps --open`. This will build the docs locally and open them in a browser.

Fixes #20

### DIFF
--- a/libs/sdk-core/src/lib.rs
+++ b/libs/sdk-core/src/lib.rs
@@ -65,9 +65,21 @@
 //!
 //! ### E. Utilities
 //!
-//! * [binding::list_payments] to get a `Vec` of [Payment] based on from/to timestamps or a [PaymentTypeFilter]
-//! * [binding::list_fiat_currencies] and [binding::fetch_fiat_rates] to get the current exchange rates
+//! Use [binding::parse] to parse generic input. The input can come from the user, from a clicked link or from a QR code.
+//! The resulting [InputType] will tell you what the input is and how to treat it, as well as present relevant payload data
+//! in a structured form.
+//!
+//! The SDK also includes payment-related utilities:
+//!
+//! * [binding::list_fiat_currencies] to get the supported fiat currencies
+//! * [binding::fetch_fiat_rates] to get the current exchange rates
 //! * [binding::recommended_fees] for the recommended mempool fees
+//!
+//! as well as wallet utilities:
+//!
+//! * [binding::list_payments] to get a `Vec` of [Payment] based on from/to timestamps or a [PaymentTypeFilter]
+//! * [binding::list_refundables] for a list of swaps
+//! * [binding::mnemonic_to_seed]
 //!
 //! ### E. Node Management
 //!
--- a/libs/sdk-core/src/lib.rs
+++ b/libs/sdk-core/src/lib.rs
@@ -1,3 +1,100 @@
+//! # Breez SDK
+//!
+//! The Breez SDK allows developers to build Bitcoin and Lightning enabled applications.
+//!
+//!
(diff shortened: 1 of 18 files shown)


libs/sdk-core/src/lib.rs
```diff
@@ -1,3 +1,143 @@
...
+//! Supported BOLT11 operations are
+//!
+//! * [BreezServices::receive_payment] to create an invoice
+//! * [BreezServices::send_payment] to pay an invoice
+//!
+//! Keysend is possible with [BreezServices::send_spontaneous_payment].
+//!
+//! See LNURL section below for LNURL-pay and -withdraw.
+//!
+//! ### C. On-chain Operations
```
roeierez, line 90: I think it is better to have a category for On-chain Receive (swap in) with the following methods:
[BreezServices::receive_onchain]
[BreezServices::list_refundables] for a list of swaps
[BreezServices::refund]

libs/sdk-core/src/models.rs
```diff
@@ -271,13 +287,15 @@ pub struct LnPaymentDetails {
...
+/// Details for a payment which closed a channel
 #[derive(PartialEq, Eq, Debug, Clone, Deserialize, Serialize)]
-pub struct ClosesChannelPaymentDetails {
+pub struct ClosedChannelPaymentDetails {
     pub short_channel_id: String,
     pub state: ChannelState,
     pub funding_txid: String,
 }
 
+/// LN channel managed by the LSP
```
roeierez, line 298: Not managed and doesn't have to be with the LSP peer. I would use just an "LN channel"

libs/sdk-core/src/lib.rs
```diff
@@ -1,3 +1,143 @@
...
+//! ```ignore
+//! let mnemonic = Mnemonic::new(Words12, English);
+//! let seed = Seed::new(&mnemonic, "");
+//!
+//! let creds = BreezServices::register_node(Network::Bitcoin, seed.as_bytes().to_vec()).await?;
+//! let sdk = BreezServices::init_services(
+//!         None,
+//!         seed.to_vec(),
+//!         creds.clone(),
+//!         Box::new(AppEventListener {}),
```
roeierez, line 31: I think in the following sections where you put some great details about the possibilities it worth adding a section for the event listener and which events can be consumed by the user.

roeierez (changes requested):
Some minor comments and suggestions. Looks great!

libs/sdk-core/src/lib.rs
ok300: Done, I added a paragraph with links in the intro section.

libs/sdk-core/src/lib.rs
ok300: Done. I then moved `[BreezServices::sweep]` to the Wallet Utilities section.

libs/sdk-core/src/models.rs

libs/sdk-core/src/models.rs
```diff
@@ -271,13 +287,15 @@ pub struct LnPaymentDetails {
     pub bolt11: String,
 }
 
+/// Details for a payment which closed a channel
```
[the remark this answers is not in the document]
ok300: Done. Not very familiar with the inner workings of this, so if there's a clearer better way to describe it, let me know.

roeierez (approved):

PR #60, opened 2023-01-17
Update rustdoc as per feedback

ok300:


libs/sdk-core/src/binding.rs
```diff
@@ -304,28 +304,28 @@ pub fn parse(s: String) -> Result<InputType> {
...
     block_on(async {
         get_breez_services()?
-            .pay_lnurl(user_amount_sat, comment, req_data)
+            .lnurl_pay(user_amount_sat, comment, req_data)
             .await
     })
 }
 
-/// See [BreezServices::withdraw_lnurl]
+/// See [BreezServices::lnurl_withdraw]
```
roeierez, line 320: I think we should rename to lnurl_withdraw in the bindings.rs as well

ok300: Good catch. Done.

PR #61, opened 2023-01-18
fix lnurl renaming

roeierez:


ok300 (approved):
Thank you, this was caused by my recent renaming.

PR #68, opened 2023-01-19
Fix double minus signal

ademar111190:
Context: https://github.com/breez/c-breez/pull/366


roeierez (approved):
LGTM once CI approves.

ok300:
Minor point, but I'll still mention it:

With the proposed change, as per the screenshot, the outgoing values would look like so

> \- 100
> Fee -2

In the main line, there is an extra space between the minus and the value. For the fee, the minus is just before the number.

Perhaps a better fix would be to keep the value from the SDK negative, but remove the extra "minus and space" prefix probably added in the UI?

That would result in

> \-100
> Fee -2

PR #69, opened 2023-01-19
BreezServices: only return currencies with known rates

ok300:


erdemyerebasmaz (approved):
LGTM

libs/sdk-core/src/fiat.rs
```diff
@@ -64,8 +64,7 @@ fn convert_to_fiat_currency_with_id(id: String, info: CurrencyInfo) -> FiatCurre
 
 #[tonic::async_trait]
 impl FiatAPI for BreezServer {
-    /// Retrieve all available fiat currencies defined in a local configuration file
-    fn list_fiat_currencies(&self) -> Result<Vec<FiatCurrency>> {
+    fn list_fiat_currencies_unfiltered(&self) -> Result<Vec<FiatCurrency>> {
```
roeierez, line 67: I don't think we should rename it. From the point of view of this trait there is no filtering.

roeierez (changes requested):

libs/sdk-core/src/fiat.rs
ok300: I only renamed it to differentiate from the other method, which would be identically named but have a different effect.

Should I then rename the `BreezServices` method? (and all callers)

Or leave both names identical, but risk confusion when someone reads the docs?

ok300: Good point, will update the PR.

ok300: Done.

Ready for review again @roeierez @erdemyerebasmaz

roeierez (approved):
LGTM! I think we should test with c-breez wallet before merge.

ok300:
What's a good way to build the c-breez apk with a specific breez-sdk branch?

PR #73, opened 2023-01-22
Make sync method public

roeierez:
This is needed for users to sync with the node until we will get the remote changes streamed to the sdk via gRPC.


ok300:
@roeierez does the SDK doc have to be updated to mention manually calling this?

If yes, when should `sync` be manually called?

ok300 (approved):

roeierez:
> If no docs update needed, then all good.

I went over the current documentation and it looks good so I don't think any update is needed

ok300:
Thanks, then this LGTM.

PR #74, opened 2023-01-23
Treat amount and fees as u64 (positive integers)

ok300:
Fixes https://github.com/breez/c-breez/issues/400


libs/sdk-core/src/models.rs
```diff
@@ -260,8 +260,8 @@ pub struct Payment {
     pub id: String,
     pub payment_type: PaymentType,
     pub payment_time: i64,
-    pub amount_msat: i64,
-    pub fee_msat: i64,
+    pub amount_msat: u64,
```
roeierez, line 263: existing wallets will fail to load probably because the fee_msat is stored as negative value in the db.
We need an SQL migration for that.

roeierez (changes requested):

libs/sdk-core/src/models.rs
ok300: Done, Erdem helped with that.

ok300:
> @ok300 is this ready for review? Asking because it is still marked as WIP.

@roeierez wanted to first test the new version myself, locally with c-breez. Just now finished my tests, ready for review.

Erdem already reviewed and said, from his side it's ready for merge.

ok300:
FYI the scenarios I tested are:
* new Android image, restore previous wallet seed
  * result: ok (full history shown, incoming and outgoing payments)
* previous master build on an Android image, keep image but install new apk (of this PR), re-open wallet
  * result: ok (same as above)
* with this apk version, make outgoing payment (negative entries in history)
  * result: ok (new negative entries appeared as expected in history)

roeierez (approved):
LGTM! once CI passes can merge.

PR #75, opened 2023-01-23
Add missing sync method

roeierez:
Add missing sync method


ok300 (approved):

roeierez:
@ok300 is this ready for review? Asking because it is still marked as WIP.

PR #85, opened 2023-01-26
Add support for AES successAction

### LINKED ISSUE #82: Add support for LUD-10: `aes` successAction in LNURL-pay
See https://github.com/lnurl/luds/blob/luds/10.md

ok300:
Fixes #82

Unlike other success actions, which can be fully used as soon as we get them, this one can only be decrypted once the LNURL-pay invoice is paid. The preimage is the decryption key.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,8 @@ jobs:
       - name: clippy
         run: |
           cd libs
-          cargo clippy -- -D warnings
-          cargo clippy --tests -- -D warnings
+          # Explicitly allow clippy::uninlined-format-args lint because it's present in the generated breez_sdk.uniffi.rs
+          cargo clippy -- -D warnings -A clippy::uninlined-format-args
+          cargo clippy --tests -- -D warnings -A clippy::uninlined-format-args
           cd ../tools/sdk-cli
           cargo clippy -- -D warnings
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -109,6 +109,27 @@ interface PaymentDetails {
     ClosedChannel(ClosedChannelPaymentDetails data);
 };
 
+dictionary AesSuccessActionDataDecrypted {
+    string description;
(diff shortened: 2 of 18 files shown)


ok300:
@erdemyerebasmaz  do you have access to 

https://github.com/breez/breez-sdk/blob/44e3165346b1e00be1c1d8dbf53eab6dc6464bad/libs/sdk-core/src/lnurl/pay.rs#L225

in Dart, in order to decrypt the payload when needed? Or should I expose it via some other way?

erdemyerebasmaz:
> @erdemyerebasmaz do you have access to
> 
> https://github.com/breez/breez-sdk/blob/44e3165346b1e00be1c1d8dbf53eab6dc6464bad/libs/sdk-core/src/lnurl/pay.rs#L225
> 
> in Dart, in order to decrypt the payload when needed? Or should I expose it via some other way?

No I'm afraid but this is a good opportunity to test this feature of flutter_rust_bridge: https://cjycode.com/flutter_rust_bridge/feature/lang_methods.html

roeierez:
> No I'm afraid but this is a good opportunity to test this feature of flutter_rust_bridge: https://cjycode.com/flutter_rust_bridge/feature/lang_methods.html

This is indeed a nice feature but let's be carfull as we need to support that in uniffi as well and the support is  different there where there is a differentiation between data structs and interfaces.

roeierez:
> in Dart, in order to decrypt the payload when needed? Or should I expose it via some other way?

@ok300  Why not decrypt the payload in rust after the payment and put the decrypted value in the success action?

roeierez:
> @ok300 Why not decrypt the payload in rust after the payment and put the decrypted value in the success action?

@ok300 did you consider this suggestion?

ok300:
Yes, that's what I'm looking into.

I originally wanted to persist the decrypted payload, but its maybe simpler to try to decrypt on the fly.

ok300:
@roeierez looks like even the simplest version was more complicated than I thought :)

Created as a separate PR on top of this one: #88

roeierez (approved):

PR #86, opened 2023-01-29
Monitor refundable swaps until resolution.

roeierez:
This PR fixes an issue when we stop monitoring unconfirmed refunded addresses.
After this fix the swap address will be monitored until the refund tx is confirmed.


roeierez:
@ok300 ping for review.

ok300 (approved):

PR #87, opened 2023-01-31
sdk-bindings adding darwin-universal

lightyear15:
following the blueprint of ios-universal target, this commit changes the darwin-related targets to build a darwin-universal swift-binding


roeierez:
Thank you @lightyear15.
I went over the changes and they looks good and consistent, just need to run some tests on my env and will add my review immediately after.

roeierez (approved):
LGTM!

PR #88, opened 2023-02-01
[WIP] Decrypt LNURL-pay AES successAction on-the-fly

ok300:
Extends #85 with the ablity to decrypt. Added as separate PR since it's a bigger change.

* The simplest approach was to directly return the decrypted payload in the LNURL response inner data object `AesSuccessActionDataDecrypted`
* This means the `BreezServices::lnurl_pay` now returns `SuccessActionProcessed`, which contains the decrypted payload in case of AES, but is otherwise identical to `SuccessAction`
* The decryption happens within `BreezServices::lnurl_pay`
  * For this purpose, `BreezServices::send_payment` now returns the `LnPaymentDetails` of the payment done, which we need to get the preimage (decryption key)
  * The `LnPaymentDetails` are retrieved from the persister via `SqliteStorage::get_payment_by_hash()`. So, to make `send_payment` correctly return the `LnPaymentDetails` during tests, a test persister was added to the mock breez services.
  * Whereas `send_payment` calls a persister lookup, the new payment has to be made known to the persister. This is done via a test-specific global list. When a new test payment is created (`MockNodeAPI::add_dummy_payment_for`), the payment struct is added to this global list. It is then queried within `sync`, such that the persister gets updated with every new generated test payment. This allows us later on to do the persister lookup `get_payment_by_hash`.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -46,6 +46,7 @@ pub trait EventListener: Send + Sync {
 /// Event emitted by the SDK. To listen for and react to these events, use an [EventListener] when
 /// initializing the [BreezServices].
 #[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
 pub enum BreezEvent {
     /// Indicates that a new block has just been found
     NewBlock { block: u32 },
@@ -53,7 +54,7 @@ pub enum BreezEvent {
     InvoicePaid { details: InvoicePaidDetails },
     /// Indicates that the local SDK state has just been sync-ed with the remote components
     Synced,
-    /// Indicates that an outgoing payment has been completed succesfully
+    /// Indicates that an outgoing payment has been completed successfully
     PaymentSucceed { details: Payment },
     /// Indicates that an outgoing payment has been failed to complete
     PaymentFailed { error: String },
@@ -248,7 +249,7 @@ impl BreezServices {
                 Ok(LnUrlPayResult::EndpointSuccess {
                     data: match cb.success_action {
                         Some(sa) => {
-                            let processed = match sa {
+                            let processed_sa = match sa {
                                 // For AES, we decrypt the contents on the fly
                                 Aes(data) => {
                                     let preimage =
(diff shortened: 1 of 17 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -184,11 +189,28 @@ impl BreezServices {
...
         self.start_node().await?;
-        self.node_api.send_payment(bolt11, amount_sats).await?;
+        let payment = self.node_api.send_payment(bolt11, amount_sats).await?;
         self.sync().await?;
-        Ok(())
+
+        // Lookup the payment in our DB and return the associated payment details
+        match self
+            .persister
+            .get_payment_by_hash(&hex::encode(payment.payment_hash))?
```
roeierez, line 204: @ok300 did you consider returning the payment structure directly from the NodeAPI::send_payment?
The conversion method from greenlight payment to our structure exists there and can be used instead of taking it from the persister post payment.

roeierez (changes requested):
Approach looks good, left a suggestion.

libs/sdk-core/src/breez_services.rs
ok300: That's a very good point.

I tinkered with different variants of this and managed to get rid of the persister lookup, instead return the Payment structure directly.

Unfortunately I didn't manage to get rid of the `test_utils::CLOUD_PAYMENTS` global, because I need to tell the `MockNodeAPI` what payments it should "know about" _before_ paying the invoice with `send_payment`.

One concern I have is that this currently only works "on-the-fly", meaning the AES payload is decrypted when the LNURL workflow returns. If the user wants to later view the LNURL-pay AES "outcome", it's currently not possible AFAIK, or am I missing something? Do we explicitly store the success action fields?

roeierez: This approach looks a lot better IMO!
Persisting the success action is a good point. We were thinking to avoid that until we have a cloud backup since a restored node won't be able to get it but maybe it is a good idea to decouple these two issues and implement the persistence as you suggested so at least it works for a new node.
@ok300  How about adding an optional field `LnUrlSuccessAction` in the `LnPaymentDetails`?

ok300: I tried different approaches, the closest to a working solution is #89. Please see the open question there.

ok300:
Marking PR as WIP because persistence is not part of it yet.

roeierez:
@ok300  Note that the change to return Payment from the various send payment functions is already in main. so maybe you want to rebase this PR?

roeierez (approved):

PR #89, opened 2023-02-03
[WIP] Persist LNURL successAction in LnPaymentDetails

ok300:
This PR persists the LNURL successAction in the `LnPaymentDetails` as per https://github.com/breez/breez-sdk/pull/88#discussion_r1095486697 :

https://github.com/breez/breez-sdk/blob/9561894c9206a50d70c0d29c010628a7e93c28ef/libs/sdk-core/src/models.rs#L295-L297

However I don't think this is a viable solution because we don't have the successAction at the point where we want to set it:

https://github.com/breez/breez-sdk/blob/9561894c9206a50d70c0d29c010628a7e93c28ef/libs/sdk-core/src/breez_services.rs#L192-L198

The LNURL successAction is not passed as an arg to the service method `send_payment`.

The tests work because the `MockNodeAPI` has the entire payment structure in advance (incl. expected successAction). However the real `NodeAPI` has to handle it in the `send_payment`.

@roeierez should I add a separate service method `send_payment_lnurl` with an extra arg for `SuccessActionProcessed`?

Or is this general approach too fragile and we should look for a different approach?

---

P.S. In addition, the `udl` doesn't parse and I haven't figured out why, but that's a small issue.

### DIFF
--- a/libs/sdk-core/src/persist/db.rs
+++ b/libs/sdk-core/src/persist/db.rs
@@ -190,6 +190,15 @@ impl SqliteStorage {
              
              DROP TABLE old_swaps;            
             "),
+            M::up(
+                "
+             CREATE TABLE IF NOT EXISTS payments_external_info (
+              payment_id TEXT NOT NULL PRIMARY KEY,
+              lnurl_success_action TEXT,
+              FOREIGN KEY(payment_id) REFERENCES payments(id)
+             ) STRICT;
+            ",
+            ),
         ]);
 
         let mut conn = self.get_connection()?;
--- a/libs/sdk-core/src/persist/transactions.rs
+++ b/libs/sdk-core/src/persist/transactions.rs
@@ -120,15 +120,21 @@ impl SqliteStorage {
             .query_row(
                 "
                 SELECT
-                 id, 
-                 payment_type, 
-                 payment_time, 
-                 amount_msat, 
-                 fee_msat, 
-                 pending, 
-                 description,
-                 details
-                FROM payments where id = ?1",
+                 p.id,
+                 p.payment_type,
+                 p.payment_time,
+                 p.amount_msat,
+                 p.fee_msat,
+                 p.pending,
+                 p.description,
+                 p.details,
+                 e.lnurl_success_action
+                FROM payments p
+                LEFT JOIN payments_external_info e
+                ON
+                 p.id = e.payment_id
+                WHERE
+                 id = ?1",
                 [hash],
                 |row| self.sql_row_to_payment(row),
             )
--- a/libs/sdk-core/src/persist/transactions.rs
+++ b/libs/sdk-core/src/persist/transactions.rs
@@ -1,6 +1,6 @@
 use super::db::SqliteStorage;
-use crate::models::{Payment, PaymentDetails};
-use crate::models::{PaymentType, PaymentTypeFilter};
+use crate::lnurl::pay::model::SuccessActionProcessed;
+use crate::models::*;
 use anyhow::{anyhow, Result};
 use rusqlite::types::{FromSql, FromSqlError, ToSql, ToSqlOutput};
 use rusqlite::OptionalExtension;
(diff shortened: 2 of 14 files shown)


roeierez:
@ok300 I think the field you added on the LnPaymentDetails is OK but we need also to handle these two requirements:
1. persist the lnurl success action
2. fetch it from the db as part of the LnPaymentDetails.

IMO for 1 we should create another table for lnurl related data (data that is related to lnurl actions). The schema of this table should contain payment_hash field and the related data field in json format (in our case the success action).
When we have the success action we should just save it to that table.
For 2 we need to change the list_payment (https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/persist/transactions.rs#L52) in the persister to join with the lnurl related data table baed on the payment_hash and use populate the returned model accordingly.

We also need to change the lnurl_pay so the returned Payment will contain the success action. We can do that by populating the returned Payment from send_payment with the success action we have before returning it to the user.

this way we decouple the greenlight LN payment conversion from the Lnurl related data.
Do you think it makes sense?

ok300:
@roeierez thanks, makes sense! Working on it.

libs/sdk-core/src/persist/db.rs
```diff
@@ -190,6 +190,14 @@ impl SqliteStorage {
              
              DROP TABLE old_swaps;            
             "),
+            M::up(
+                "
+             CREATE TABLE IF NOT EXISTS lnurl_success_action (
```
roeierez, line 195: I think it is better to cal this table in a more generic way such as `payments_external_info` and create a foreign key from the Payments.id column to a column here called payment_id (for lightning payments it is the hash).
Also I think we should call the value column `lnurl_success_action`.

This way we gain the following advantages:
1. It is clear that it is associated with the payments table and the db validates that using the foreign key.
2. We for sure will need to map more data in the future (such as lnurl metadata) so we can do that by simply adding more columns to this table.

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -1,16 +1,34 @@
...
+    /// Inserts payments into the payments table.
+    ///
+    /// If a payment has details of type [LnPaymentDetails] which contain a [SuccessActionProcessed],
+    /// then it is inserted in the separate SuccessActions table.
+    ///
+    /// The [SuccessActionProcessed] can be inserted separately via [SqliteStorage::insert_lnurl_success_action],
+    /// if it's not part of the payment when this method is called.
     pub fn insert_payments(&self, transactions: &[Payment]) -> Result<()> {
+        // Insert available SuccessActions in separate table
+        for p in transactions.iter() {
```
roeierez, line 20: I don't think we need this. The success action is inserted after succesfull payment in a different insert.

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -107,16 +151,26 @@ impl SqliteStorage {
...
             fee_msat: row.get(4)?,
             pending: row.get(5)?,
             description: row.get(6)?,
             details: row.get(7)?,
-        })
+        };
+
+        match payment.details {
+            PaymentDetails::Ln { ref mut data } => {
+                data.lnurl_success_action = row.get(8)?;
```
roeierez, line 168: Nice!

roeierez (changes requested):
Looks great! dropped some feedback.

libs/sdk-core/src/persist/transactions.rs
ok300: Done

libs/sdk-core/src/persist/db.rs
ok300: Done

libs/sdk-core/src/persist/db.rs
```diff
@@ -190,6 +190,15 @@ impl SqliteStorage {
              
              DROP TABLE old_swaps;            
             "),
+            M::up(
+                "
+             CREATE TABLE IF NOT EXISTS payments_external_info (
+              payment_hash TEXT NOT NULL PRIMARY KEY,
```
roeierez, line 196: NIT:  In my last feedback I mentioned to change the payment_hash to payment_id so we can associate data later on non lightning payments (closed channel for example).

roeierez (approved):

roeierez (approved):

libs/sdk-core/src/persist/db.rs
ok300: Sorry I must have missed that. Done.

PR #90, opened 2023-02-06
Add PaymentSucceed & PaymentFailed events

roeierez:


ok300 (approved):

erdemyerebasmaz:
Shouldn't we at least add an error message to PaymentFailed event?

roeierez:
@erdemyerebasmaz good feedback, see here: https://github.com/breez/breez-sdk/pull/92

PR #94, opened 2023-02-08
Ignore reserved outputs for on-chain balance

roeierez:
Here we use this fix for greenlight (https://github.com/Blockstream/greenlight/issues/74) in order to ignore the reserved outputs.
A reserved output is one that is used in a current transaction that is still unconfirmed. This allow us to ignore it when calculating the on-chain balance allowing clients to get better user experience when sweeping from the wallet.
fixes https://github.com/breez/c-breez/issues/330


ok300 (approved):
LGTM.

ok300:
FYI: If you rebase this on master, the CI warnings should go away. I fixed them in b2935cc986bcbe63b6027ae310f1b3b9b722efd9

Issue #45, opened 2023-01-08
Generate React Native module


roeierez:
- [ ] Implement the native kotlin code
- [ ] Implement the swift native code
- [ ] Implement the javascript wrapper
- [ ] Implement code generation for the kotlin/swift code
- [ ] implement code generation for the javascript wrapper

I believe the end result should be a 100% generated code for RN Native module however we should first develop the module manually and then add the automatic code generation.
Here is an intro on how to create such native modules: https://reactnative.dev/docs/native-modules-intro
Such module has two parts:
1. The native (swift, kotlin) code - which is basically enough for using it from javascript but without type safety for objects passing as arguments.
2. The javascript wrapper - That makes it easier to use and provide type safe interface for the module.

### Creating the native code

I think this could be a wrapper on top of the uniffi interface, for example in kotlin it is another kotlin class that delegates the calls to the uniffi generated functions.
The reason why we need a wrapper is because the wrapper provides the interface for the javascript code and it has some constraints (for example doesn't support objects but use ReadableMaps instead https://reactnative.dev/docs/native-modules-android#argument-types) and we can't use the uniffi generated interface for that. So the RN native code needs to provide the interface to the javacript as similar as possible to the uniffi interface and then handle argument mapping when delegating the calls.

After that is in place we can develop the javascript code that consists mostly of delegation to the native code and convert maps to type safe interfaces.

PR #96, opened 2023-02-11
[WIP] React Native framework

dangeross:
This PR adds `sdk-react-native` framework to the libs available in breez-sdk. Following the initial steps of #45 this is a manually implemented framework, which builds the bindings available in `sdk-bindings` and copies the resulting libraries into the android and ios projects.  Currently only `mnemonicToSeed` is implemented from the SDK as proof-of-concept and included in the framework is an example project that can be built to test this.


libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -0,0 +1,27 @@
...
+class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {  
+    override fun getName(): String {
+        return "BreezSDK";
+    }
+
+    @ReactMethod
+    fun mnemonicToSeed(mnemonic: String, promise: Promise) {
+        try {
+            var seed = mnemonicToSeed(mnemonic);
+            var response = writableMapOf("type" to "seed", "data" to seed);
```
roeierez, line 19: What is the advantage of adding the "type" key here? What is the advantage of this structure vs returning ReadableArray as a single value?

libs/sdk-react-native/android/src/main/java/com/breezsdk/MapResponse.kt
```diff
@@ -0,0 +1,60 @@
+package com.breezsdk
+
+import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.WritableArray
+import com.facebook.react.bridge.WritableMap
+
+fun writableMapOf(vararg values: Pair<String, *>): WritableMap {
```
roeierez, line 7: Isn't it better to use the readonly version for the return types (ReadableMap)?

roeierez (commented):
@dangeross this looks really good! I left some comments to understand better your decisions.
Thank you very much!

libs/sdk-react-native/android/src/main/java/com/breezsdk/MapResponse.kt
dangeross: You're right, these return types should be ReadableArray or ReadableMap.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
dangeross: I intend to change this to just map the SDK responses as much as possible.

roeierez (approved):
@dangeross LGTM!

roeierez:
@dangeross added you as a collaborator, feel free to merge.

PR #101, opened 2023-02-13
React Native wrappers

dangeross:
Continued implementation of kotlin & swift native wrappers around breez-sdk service calls and map response data back over the react native bridge. Implementation of the javascript module interface with typing.

- [x] registerNode
- [x] recoverNode
- [x] initServices
- [x] setLogStream
- [x] parseInvoice
- [x] parseInput
- [x] mnemonicToSeed
- [x] start
- [x] stop
- [x] sendPayment
- [x] sendSpontaneousPayment
- [x] receivePayment
- [x] withdrawLnurl
- [x] nodeInfo
- [x] listPayments
- [x] sweep
- [x] fetchFiatRates
- [x] listFiatCurrencies
- [x] listLsps
- [x] connectLsp
- [x] fetchLspInfo
- [x] lspId
- [x] closeLspChannels
- [x] receiveOnchain
- [x] inProgressSwap
- [x] listRefundables
- [x] refund
- [x] executeDevCommand
- [x] sync
- [x] recommendedFees


roeierez:
@dangeross I went over this PR and so far it looks great.
Seeing all these subtasks I gm guessing your plan is to implement the rest of the methods in this single PR?  I am fine with any decision of yours.

dangeross:
Yeah, my plan was to implement everything in a single PR. I can probably manage a couple of methods each day, so if you prefer to break it up into smaller chunks and have others work on it too, I'm fine with that.

roeierez:
@dangeross We are still busy with the c-breez alpha release so I am fine with both options for now. Thank you!

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -211,6 +222,20 @@ class BreezSDKMapper {
...
     }
     
+    static func asLnUrlWithdrawRequestData(reqData: [String: Any]) -> LnUrlWithdrawRequestData? {
+        if let callback = reqData["callback"] as? String,
+            let k1 = reqData["k1"] as? String,
+            let defaultDescription = reqData["defaultDescription"] as? String,
+            let minWithdrawableStr = reqData["minWithdrawable"] as? String,
+            let maxWithdrawableStr = reqData["maxWithdrawable"] as? String {
+            if let minWithdrawable = UInt64(minWithdrawableStr), let maxWithdrawable = UInt64(maxWithdrawableStr) {
+                return LnUrlWithdrawRequestData(callback: callback, k1: k1, defaultDescription: defaultDescription, minWithdrawable: minWithdrawable, maxWithdrawable: maxWithdrawable)
```
dangeross, line 232: Is there a more elegant way to do this?

roeierez: Are you referring to the type conversion from js to swift? shouldn't we read it directly as double?

libs/sdk-react-native/src/index.tsx
```diff
@@ -1,25 +1,329 @@
...
+export async function receivePayment(amountSats: Long, description: string): Promise<LnInvoice> {
+    const response = await BreezSDK.receivePayment(amountSats.toString(), description)
+    return response as LnInvoice
+}
+
+export async function withdrawLnurl(reqData: LnUrlWithdrawRequestData, amountSats: Long, description?: string): Promise<LnUrlWithdrawCallbackStatus> {
+    const response = await BreezSDK.withdrawLnurl(
+        {
+            ...reqData,
+            minWithdrawable: reqData.minWithdrawable.toString(),
```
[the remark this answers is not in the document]
dangeross: I guess it's bias from using LND with react native, with the js protobuf it converts the int64 values to long. JavaScript represents numbers using 64 bit double-precision, so for integer values it should be ok

libs/sdk-react-native/ios/BreezSDKMapper.swift
dangeross: The int64 values are converted to string over the bridge, but i'll refactor that.

libs/sdk-react-native/src/index.tsx
```diff
@@ -1,25 +1,329 @@
...
+            return response.data as LnInvoice
+        case InputType.LNURL_AUTH:
+            return response.data as LnUrlAuthData
+        case InputType.LNURL_ERROR:
+            return response.data as LnUrlErrorData
+        case InputType.LNURL_PAY:
+            return response.data as LnUrlPayRequestData
+        case InputType.LNURL_WITHDRAW:
+            return response.data as LnUrlWithdrawRequestData
+        case InputType.NODE_ID:
```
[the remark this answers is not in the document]
dangeross: Good point

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -1,27 +1,236 @@
...
+                e.printStackTrace()
+                promise.reject(TAG, "Error calling stop", e)
+            }
+        } ?: run {
+            promise.reject(TAG, "BreezServices not initialized")
+        }
+    }
+
+    @ReactMethod
+    fun sendPayment(bolt11: String, amountSats: String?, promise: Promise) {
```
dangeross, line 159: @roeierez I'm testing changing the type to optional Double:
```
    fun sendPayment(bolt11: String, amountSats: Double?, promise: Promise) {
```
but when mirroring it in javascript:
```
export async function sendPayment(bolt11: string, amountSats?: number): Promise<Payment> {
```
it causes an error if no amountSats is passed:
```
Attempt to invoke virtual method double java.lang.Double.doubleValue() on a null object reference
```
Is it possible to send a 0 amountSats in the case no param is provided? Does it affect the sdk? Other options could be to wrap all params over the bridge in an object, or passing as a string.

roeierez: Which component is trying to call `java.lang.Double.doubleValue() `? According to the React Native docs for this case we should use number? in JS (as you just did) and Double in kotlin (not Double?), see here: https://reactnative.dev/docs/native-modules-android#argument-types

Also I admit I don't understand why this might help, worth trying as the docs clearly shows it should be supported.

roeierez: Also how do you pass it from the typescript to the native code? As is without conversion? Maybe if you can paste here the JS code I would be able to help.

dangeross: I tried also with using `number?` in JS and `Double` in kotlin. Same error. Below is the typescript code, calling without any conversion:
```
export async function sendPayment(bolt11: string, amountSats?: number): Promise<Payment> {
    const response = await BreezSDK.sendPayment(bolt11, amountSats)
    return response as Payment
}
```
if the function is called with only a bolt11 param the error occurs
```

await sendPayment("lnbc1p37jj09pp5rt6...")
```

dangeross: Here the stack trace from react native:
Screenshot 2023-02-16 at 10 32 01

roeierez: @dangeross I have looked at the bridge code and indeed there is not way to pass nullable primitives.
In that case I think your suggestion to pass 0 from the typescript code to the native code and then again null from the native code to the sdk is the best option.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -1,27 +1,494 @@
...
+                var payment = breezServices.sendSpontaneousPayment(nodeId, amountSats.toULong())
+
+                promise.resolve(readableMapOf(payment))
+            } catch (e: SdkException) {
+                e.printStackTrace()
+                promise.reject(TAG, "Error calling sendSpontaneousPayment", e)
+            }
+        } ?: run {
+            promise.reject(TAG, "BreezServices not initialized")
+        }
```
roeierez, line 188:
```suggestion
            try {
                var payment = getBreezServices().sendSpontaneousPayment(nodeId, amountSats.toULong())

                promise.resolve(readableMapOf(payment))
            } catch (e: SdkException) {
                e.printStackTrace()
                promise.reject(TAG, "Error calling sendSpontaneousPayment", e)
            }
        } 
```

Perhaps we can simplify a bit this patter that return on every method? It would need to create a method getBreezServices() that returns the initialized services or throws.

dangeross: I'll do that, for swift also

PR #103, opened 2023-02-14
Add raspiberry pi target

ademar111190:
Breez sdk running on Raspberry PI :tada: 

Screenshot_20230213_183743

### DIFF
--- a/libs/sdk-core/makefile
+++ b/libs/sdk-core/makefile
@@ -18,7 +18,7 @@ init:
 	rustup target add aarch64-apple-darwin x86_64-apple-darwin
 	rustup target add aarch64-apple-ios-sim
 	rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
-	rustup target add aarch64-unknown-linux-gnu
+	rustup target add aarch64-unknown-linux-gnu armv7-unknown-linux-gnueabi i686-unknown-linux-gnu x86_64-unknown-linux-gnu
 	@if [ $$(uname) == "Darwin" ] ; then cargo install cargo-lipo ; fi
 	cargo install cbindgen
 	cargo install cargo-ndk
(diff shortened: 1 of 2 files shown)


libs/sdk-core/makefile
```diff
@@ -17,20 +17,22 @@ init:
...
 ## all: Compile iOS, Android
-all: ios macos android
+all: ios macos android linux
 
 ## ios: Compile the iOS universal library
 ios: ios-universal
 
 ios-universal: $(SOURCES) ndk-home
-	cargo lipo --release
+	@if [ $$(uname) == "Darwin" ] ; then \
```
roeierez, line 33: Do we need this? Isn't it better to just fail if it can't build?

ademar111190: I believe it is not allowed to build ios binaries outside of MacOS, but I'm not sure, I did add it because on Linux it fails when I tried to build it.

roeierez: Yes exactly, isn't it better that it will fail vs report DONE but won't build?

ademar111190: gotcha, you are right.

ademar111190:
> If you add linux, please add also arm7, i686 and x86_64 architectures

I don't have a Linux running on these specs to test, Isn't it better to wait for someone that can test and let this person add it? @yaslama

yaslama:
> > If you add linux, please add also arm7, i686 and x86_64 architectures
> 
> I don't have a Linux running on these specs to test, Isn't it better to wait for someone that can test and let this person add it? @yaslama

Please add them and I will check them.

libs/sdk-core/makefile
```diff
@@ -66,21 +66,36 @@ aarch64-apple-darwin: $(SOURCES)
 	fi
 	@echo "[DONE] $@"
 
-## android: Compile the android targets (arm64, armv7 and i686)
```
ademar111190, line 69: A minor update in the comment and reorder to follow a pattern, i.e. 💄

ademar111190:
> > > If you add linux, please add also arm7, i686 and x86_64 architectures
> > 
> > 
> > I don't have a Linux running on these specs to test, Isn't it better to wait for someone that can test and let this person add it? @yaslama
> 
> Please add them and I will check them.

Added, please let me know if something breaks, at the worst scenario I can create some VM and try to fix it.

ok300:
@ademar111190 some CI errors were fixed on `master`, if you rebase this PR maybe it also fixes the CI problems here.

roeierez (approved):

yaslama (approved):

PR #108, opened 2023-02-15
Add service method for starting reverse swap

ok300:


libs/sdk-core/src/breez_services.rs
```diff
@@ -425,8 +426,25 @@ impl BreezServices {
     }
 
     pub async fn reverse_swap_info(&self) -> Result<ReverseSwapInfo> {
-        let rsi = crate::boltzswap::reverse_swap_info().await?;
-        return Ok(rsi);
+        crate::boltzswap::reverse_swap_info().await
```
roeierez, line 429: Shouldn't we add the `reverse_swap_info` to the `ReverseSwapperAPI` trait implemented by BoltzServer?
It is good to have traits for APIs not only because we will support different swap implementations but also for the code to be testable.

libs/sdk-core/src/reverseswap.rs
```diff
@@ -2,15 +2,66 @@ use std::sync::Arc;
...
+    timeout_block_height: u64,
+    lockup_address: String
+}
 
 #[tonic::async_trait]
 impl ReverseSwapperAPI for BreezServer {
-    async fn create_reverse_swap(&self) -> Result<ReverseSwap> {
-        let m = String::from("");
-        return Ok(ReverseSwap { error_message: m });
+    async fn create_reverse_swap(&self, amount_sat: u64, onchain_claim_address: String) -> Result<ReverseSwap> {
```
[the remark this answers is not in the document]
ok300: How about as a method in

https://github.com/breez/breez-sdk/blob/970f144d3e3a1c781eaa4da6393d4d6ed6bff2fd/libs/sdk-core/src/reverseswap.rs#L17-L25

It already encapsulates reverseswap-related structs, and it's a field of `BreezServices`.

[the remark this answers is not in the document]
ok300: The methods for normal swap are also directly in `BreezServer`:

* Normal swap: `impl SwapperAPI for BreezServer`
* Reverse swap: `impl ReverseSwapperAPI for BreezServer`

Should both be delegated to other structs? Like the existing

```rust
pub struct BreezServices {
    btc_receive_swapper: Arc<BTCReceiveSwap>,
    btc_send_swapper: Arc<BTCSendSwap>,
    // ...
}
```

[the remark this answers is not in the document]
roeierez: The BTCSendSwap seems right to me as well but I think the boltz API methods should be defined in the trait ReverseSwapperAPI and BreezServices should interact only with BTCSendSwap (not directly with botlz API).
This way we will:
1. Separate the API from the swap client logic
2. Be able to write tests.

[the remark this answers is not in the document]
ok300: Done.

libs/sdk-core/src/breez_services.rs
ok300: Done.

ok300:
I moved the reverse swap methods in `BoltzApi`, which is a component of `BTCSendSwap`, which is a field of `BreezServices`.

This should address the two comments above, but @roeierez please let me know if I missed something.

Next step is to register the tx to be monitored.

libs/sdk-core/src/boltzswap.rs
```diff
@@ -77,18 +85,59 @@ struct Pairs {
...
+            .body(get_boltz_reverse_swap_args(
+                amount_sat,
+                preimage_hash_hex,
+                pair_hash,
+                onchain_claim_address,
+                routing_node,
+            ))
+            .send()
+            .await?
+            .text()
```
roeierez, line 120: reqwest crate has a built in json feature where you can use the json() method to deserialize the response. Might be more elegant, see here: https://docs.rs/reqwest/latest/reqwest/struct.Response.html

libs/sdk-core/src/breez_services.rs
```diff
@@ -425,8 +427,39 @@ impl BreezServices {
     }
 
     pub async fn reverse_swap_info(&self) -> Result<ReverseSwapInfo> {
-        let rsi = crate::boltzswap::reverse_swap_info().await?;
-        return Ok(rsi);
+        self.btc_send_swapper
+            .reverse_swapper_api
```
roeierez, line 431: I think it is better to make the reverse_swapper_api private field of the btc_send_swapper and make an appropriate method that uses it internally.

roeierez (changes requested):
Some small comments, otherwise looks good.

libs/sdk-core/src/boltzswap.rs
ok300: Good point, done.

libs/sdk-core/src/breez_services.rs
ok300: Done.

PR #114, opened 2023-02-21
Fast receive

roeierez:
The goal of this PR was to sync the payments list right after receive with the invoice that is sent from greenlight as part of stream_incoming endpoint. The expectation is that receive would be much faster and without additional sync delays.
As part of the PR this was implemented as well:
1. Use TryFrom trait for converting greenlight structure to Payments
2. Fix MockNodeAPI to use non static cloud_payments list as the static was shared between tests causing them to fail from time to time.


ok300:
Code LGTM. Will try the apk.

roeierez:
@ok300 LMK if you have any feedback. Would be great to merge it.

ok300 (approved):
All good, works!

PR #123, opened 2023-02-23
Remove InputParser and it's dependencies

erdemyerebasmaz:
InputParser that uses an external library is being removed as all capabilities can now be handled using SDK.

- Removed InputParser and it's dependencies.
- Updated dependencies to latest


roeierez (approved):
LGTM but let's wait with the merge until we integrate properly the sdk input parser.

PR #126, opened 2023-02-27
Store the LN Address of an LNURL-pay, if available

### LINKED ISSUE #91: Add lightning address to payment model
We need to add the lightning address to the payment model https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/models.rs#L259-L268 in order to replicate the breez mobile logic here https://github.com/breez/breezmobile/blob/232f1178d86b1b7152a03036ab33aafcf0c933a9/lib/bloc/account/account_model.dart#L885

ok300:
Fixes #91

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -140,6 +140,7 @@ dictionary LnPaymentDetails {
     boolean keysend;
     string bolt11;
     SuccessActionProcessed? lnurl_success_action;
+    string? ln_address;
 };
 
 dictionary ClosedChannelPaymentDetails {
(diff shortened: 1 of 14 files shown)


roeierez (changes requested):
Looks good! React Native changes are missing.

roeierez:
> There is one test failing, I'm still not sure why:

@ok300  It seems that the ln_address change needs to be done also to `get_payment_by_hash` in persiste/transactions.rs.
I think it fails now which causes the test to fail.

ok300:
Correct!

roeierez (approved):

dangeross:
I think react native mappers still need updating for this too:
kotlin: https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
swift: https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/ios/BreezSDKMapper.swift

ok300:
Thanks @dangeross I added a separate PR for those.

Is there a way to add a CI check for these bindings?

PR #128, opened 2023-03-01
Notifications draft

roeierez:


ubbabeck (approved):
Concise and easy to understand  🙌

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Chain notification service
+Breez server will implement the chain notifications for sdk clients. Connected to bitcoind breez server download every block, check for webhook registration and trigger notifications. API key is needed to subscribe for such notification.
+The following endpoints are needed:
+- `registration_id add_registration(current_registration_id, url)` - get the current registration_id if exists or create one if the former was deleted. Also associate the url with the new registration if is not already associated. At the first time this endpoint is called the `current_registration_id` can be null. This mechanism allows the service to delete expired registrations according to its own policy requiring the client to refresh the registration from time to time signaling it is truly active.
+- `subscribe_address_transactions(registration_id, url, address)` API for subscribing to address notifications.
+- `subscribe_tx_confirmation(registration_id, url, txid)` API for subscribing to specific transaction notifications.
```
ok300, line 39: Very clear and concise overall, thanks.

One question about the chain notification service for the SDK: since the `url` is an argument for subscribing to these notifications, I assume it's not possible to subscribe, for example, in the CLI app, correct?

roeierez: Good question. 
I started this doc by writing that the goal of the SDK is to provide the trigger to the delivery service (push, mail, etc...) So in fact it is the responsibility of the app to connect to the delivery service (e.g subscribe to push notifications) and the app provider to supply the delivery service (e.g a backend service that actually send the push notifications, email, etc..). 
In case we want the cli to receive notifications this kind of design would work as well, for example: We can implement the delivery service as a gRPC service that the cli connects to on startup with some random token and registers for webook that contains that random token.
The hook points to the gRPC service and once triggered, the service identify the right subscriber associated with that token and send a gRPC message over the stream.

roeierez: Saying all the main use case for notifications is to be notified when not online so I don't see a need for that in the cli which by its nature is always online.

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+The `registration_id` is a random unique identifier that identify the caller and allow the sdk client to both update the hook url and subscribe to notification services.
+
+### Notification Delivery Service
+This service is the destination of the web hooks. Its job is to identify the protocol, type and data of the notification, craft the formatted notification and send it over the right channel (mobile push for example)
+Such service should be opertated by an app provider that would like to send push notifications to users.
```
JssDWt, line 44: nit: operated

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+### SDK
+- `register_notifications(url, platform, token)` API method for the user to register all kind of notifications.
+- Build the webhook URL (defined later in the document) and use it in the following cases:
+ - When creating an on-chain receive swap address - Use breez server API to register for notification of any tx associated with this address
+ - Use LSP API to register webhook to be notified when there is an incoming paypment(htlc) for the user node.
+
+### LSP
+- `subscribe_notifications(url, signature)` API for registering a webhook by sdk clients. The signature is the url signed by the node id. The LSP should maintain a mapping between node ids to web hooks URLs. These webhooks (from the client perspective) tend to change frequently as for example in push notifications a new token is generated from time to time so the client is responsible to update the url on any change.
+
+When an htlc is intercepted in the LSP and the destination peer is not online the LSP will notify the web hook with the right type and data, signling incoming payment, in attempt to wake up the destination node, wait for a some time (how much?) and once the peer is online will forward the htlcs.
```
JssDWt, line 32: nit: signaling

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+- The web hook notifies the delivery services that construct the specific notification format and payload and send it to the device. In that case the push message can be silent, waking up the app to run the signer and complete the swap payment.
+
+## Suggested implementation
+There are several parts for achieving the above:
+
+### SDK
+- `register_notifications(url, platform, token)` API method for the user to register all kind of notifications.
+- Build the webhook URL (defined later in the document) and use it in the following cases:
+ - When creating an on-chain receive swap address - Use breez server API to register for notification of any tx associated with this address
+ - Use LSP API to register webhook to be notified when there is an incoming paypment(htlc) for the user node.
```
JssDWt, line 27: nit: payment

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+- `subscribe_address_transactions(registration_id, url, address)` API for subscribing to address notifications.
+- `subscribe_tx_confirmation(registration_id, url, txid)` API for subscribing to specific transaction notifications.
+The `registration_id` is a random unique identifier that identify the caller and allow the sdk client to both update the hook url and subscribe to notification services.
+
+### Notification Delivery Service
+This service is the destination of the web hooks. Its job is to identify the protocol, type and data of the notification, craft the formatted notification and send it over the right channel (mobile push for example)
+Such service should be opertated by an app provider that would like to send push notifications to users.
+
+## Webhook structure
+A webhook is just a url but for mobile we require `platform` and `token` parameters for example: <base_url>?platform=<platform>&token=<token>
```
JssDWt, line 47: This seems specific to the app use case. I don't think it's relevant for a webhook document. The client could register any (https) url. The url will be client specific, depending on the implementation of the Notification Delivery Service.

It does make sense to mention that these webhook urls should be authenticated (to avoid DoS attacks on Notification Delivery Service endpoints)

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+Applications have their own way of delivering notifications to end users such as: mobile push, web push, emails, text messages etc...
+This is usually implemented using some backend service that has a way to communicate with the end user. The SDK goal is to provide the trigger including the type and notification data to that service. A simple and standard option to implement that trigger is using webhooks.
+
+## The use cases:
+
+### Offline payment
+User is offline and would like to be notified to accept a received payment.
+- The client associate its node id with the webhook by calling a dedicated API endpoint of the LSP.
+- The sender pays, htlcs land at the LSP which detects the last hop is not connected as a peer and send a POST request to the associated webhook with the payload that includes the notification type and data.
+- The web hook notifies the delivery service that builds the specific notification format and payload and send it to the device. In that case the push message can be silent, waking up the app to run the signer and process the payment.
```
JssDWt, line 11: Brainstorm: There's a period here where the LSP will hold the htlc and will wait for the client to 'respond' to the webhook by waking up and connecting. Timeout would be the simplest option here.

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+Where:
+- `platform` is one of: apn, fcm
+- `token` is the token given by the client OS.
+
+Triggering the webhook is done by initiating a POST request with the following json payload:
+
+{
+ "type": <hook_type>,
+ "data": <hook_data>
+}
```
JssDWt, line 57: a timestamp is usually good practice to add in webhook contents too.
And I'm thinking a signature could be beneficial, because the webhook endpoint will be a rather open endpoint with credentials in the url. (you want to somehow restrict the services that can call the webhook endpoint in order to wake up your mobile phone)

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+- The client initiate a swap and send funds to the on-chain address.
+- The client registers to a push notification for the address using web hook.
+- Once the transaction is confirmed the notification server send a POST request to the associated webhook with the payload that includes the notification type and data.
+- The web hook notifies the delivery services that construct the specific notification format and payload and send it to the device. In that case the push message can be silent, waking up the app to run the signer and complete the swap payment.
+
+## Suggested implementation
+There are several parts for achieving the above:
+
+### SDK
+- `register_notifications(url, platform, token)` API method for the user to register all kind of notifications.
```
JssDWt, line 24: Any webhook subscription should have an expiry time, to avoid having infinite dangling subscriptions. (maybe we could eeven notify when the expiry time is coming up)

JssDWt (commented):
I left some thoughts and comments. Overall looks really good.

doc/SDK_NOTIFICATIONS.md
```diff
@@ -0,0 +1,77 @@
+# Breez SDK Notifications
+Applications have their own way of delivering notifications to end users such as: mobile push, web push, emails, text messages etc...
+This is usually implemented using some backend service that has a way to communicate with the end user. The SDK goal is to provide the trigger including the type and notification data to that service. A simple and standard option to implement that trigger is using webhooks.
+
+## The use cases:
+
+### Offline payment
+User is offline and would like to be notified to accept a received payment.
+- The client associate its node id with the webhook by calling a dedicated API endpoint of the LSP.
+- The sender pays, htlcs land at the LSP which detects the last hop is not connected as a peer and send a POST request to the associated webhook with the payload that includes the notification type and data.
```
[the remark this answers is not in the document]
roeierez: The greenlight node is offline, LSP triggers the webook that notify the mobile client which in response schedule the node and start the signer. The node must be scheduled to peer with the LSP.



> Probably both the LSP and the Greenlight node could dispatch webhooks.

Yes, this is the plan indeed although for different cases. The signer would probably use the webhook to settle pending htlcs and the LSP in this case would use webhook to schedule the node.

doc/SDK_NOTIFICATIONS.md
roeierez: I was thinking the expiry time is subject to the service policy. For example the service may decide that if the client didn't refresh the registration for 3 days the registration would expire. 



> maybe we could even notify when the expiry time is coming up

Nice. Notification (in mobile could be even silent) to refresh the registration seems a great option to achieve this goal UX wise.

doc/SDK_NOTIFICATIONS.md
roeierez: > This seems specific to the app use case. I don't think it's relevant for a webhook document. The client could register any (https) url. The url will be client specific, depending on the implementation of the Notification Delivery Service

I agree.

roeierez: @JssDWt I do want to add the webhook structure we use so it will be clear to everyone involved the structure, params etc...
Two options:
1. I can open another doc for that 
2. I can leave that in this doc but under "Breez Specific" section

What do you think?

JssDWt: Good question. If the SDK client registers for events at the LSP itself and the Notification Delivery Service (NDS, breez server in our case) is simply a passthrough, it could make sense to define that here.
But more likely the client will have to register for notifications at the NDS itself? And the client and the NDS would have some kind of contract on how the notifications will be delivered and it what form.

So some options on how that could work.
1) 
- client requests notification url from NDS
- client registers notification url with LSP
- NDS maps events from LSP to client

2)
- client asks NDS to get notifications
- NDS registers notification URL with LSP
- NDS maps events from LSP to client

In the end I think it makes sense there will be a contract between LSP and client, it's just the delivery/transport to the client that may differ. It's just a little bit hard to reason about where this contract needs to go. Ideally we'd define as much as possible in the SDK itself, because the SDK will respond to these events coming from the LSP.

So; long story short. I think it makes sense to define the webhook contract here for now. Make the SDK expect those exact objects as defined in this contract, only the transport between NDS/client should be explicitly left out for the implementor to decide. Another document might make sense, because the webhooks are on the LSP side of the spec.

roeierez: In my mind the NDS doesn't communicate with the client directly but is only a "proxy" to transfer web hook triggers to a specific platform notification and deliver it to the client. 
So the flow goes like this:
1. client (SDK) registers a webhook URL with the LSP
2. LSP, upon notification, send POST request to the web hook URL (which is pointed to the NDS)
3. NDS receives and parses the request, convert it into a specific platform notification (such as push) and deliver it to the client.

For this to happen correctly the client should put some information on the webhook URL for the NDS to identify it (such as push token and platform). This is why I added the info in the document but maybe I should have scoped it with "implemention detail" or as an example for mobile push.

JssDWt: OK, so that's option 1. I agree the NDS doesn't communicate with the client directly. But it 'knows' how to reach the client to deliver the message (so that part can't be spec'ed). In order to know how to reach the client it needs some kind of data, which needs to go in the URL. So the URL is made by the NDS (or the client will know how to construct a valid URL to the NDS). I think it makes sense to add an endpoint to the NDS so the client can request a URL there.

Anyway the URL construction is NDS specific, we could add a separate document with Breez-specific examples.

PR #129, opened 2023-03-05
Update model to distinguish DB-stored struct vs Boltz-retrieved info struct

ok300:
Separate the two distinct structs:
- `ReverseSwapInfo` = the Breez view of a reverse swap (as stored in DB)
- `ReverseSwapPairInfo` = the Boltz reverse swap info for the BTC/BTC pair, containing point-in-time exchange rate and fees

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -107,14 +107,54 @@ pub trait SwapperAPI: Send + Sync {
     async fn complete_swap(&self, bolt11: String) -> Result<()>;
 }
 
+/// Details about the BTC/BTC reverse swap pair, at this point in time
+///
+/// Maps the result of https://docs.boltz.exchange/en/latest/api/#getting-pairs for the BTC/BTC pair
 #[derive(Clone, PartialEq, Debug)]
-pub struct ReverseSwapInfo {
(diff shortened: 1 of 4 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -107,16 +107,56 @@ pub trait SwapperAPI: Send + Sync {
...
+    pub claim_address: String,
+    pub lockup_address: String,
+
+    // dynamic data
+    pub status: ReverseSwapStatus,
+}
+
+/// The status of a reverse swap
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum ReverseSwapStatus {
```
roeierez, line 142: @ok300  I believe that as part of monitoring the swap on-chain we will need to update the data such as invoice paid hash, lock_tx_id, claim_tx_id, etc...
So it is early to know if we need these statuses or can derive the state from the data in the db. Ok for now but keep in mind we might (or not) remove these in the future.

roeierez (approved):

libs/sdk-core/src/models.rs
ok300: Good point. Even better if the statuses can be derived.

PR #130, opened 2023-03-05
Move util methods and structs from swap to chain

ok300:
To monitor whether utxos for an address are confirmed or not, some utils in `swap.rs` are used.

Since the same logic will be used in the reverse swap, I moved these structs and utils to a common place (`chain.rs`).

This covers
- the `Utxo` and `AddressUtxos` structs
- the `get_utxos` util method

While doing this, I realised there was potentially a bug in the swap code:

```rust
if !confirmed_txs.is_empty()
        && current_tip - confirmed_block >= swap_info.lock_height as u32
    {
        swap_status = SwapStatus::Expired
    }
```

does not cover the case when no transactions are confirmed, but the tip already moved beyond `lock_height`.

I removed the first check and that block became:
```rust
if current_tip - confirmed_block >= swap_info.lock_height as u32
{
     swap_status = SwapStatus::Expired
}
```


roeierez:
> does not cover the case when no transactions are confirmed, but the tip already moved beyond `lock_height`.

@ok300 but the lock_height is relative to the confirmed block. So if there is not confirmed tx the swap is not expired.

ok300:
Wasn't aware of that, I thought it's absolute. Will update.

ok300:
Adding the check makes `test_spent_swap` fail because

```
'swap::tests::test_spent_swap' panicked at 'assertion failed: `(left == right)`
  left: `Initial`,
 right: `Expired`', sdk-core/src/swap.rs:865:9
```

Looking closer at that test, it checks that

a) there are no confirmed transactions

```rust
assert_eq!(swap.confirmed_tx_ids, Vec::<String>::new());
```

and b) the swap status is expired

```rust
assert_eq!(swap.status, SwapStatus::Expired);
```

However, the status is not updated to `Expired` precisely because I (re-)added the check for non-empty confirmed txs in

```rust
if !confirmed_txs.is_empty()
    && current_tip - confirmed_block >= swap_info.lock_height as u32
{
    swap_status = SwapStatus::Expired
}
```

Could it be that extra check is not actually needed in the snippet above, since it restricts the function to only update the status to `Expired` if there are confirmed txes?

Or maybe the test has a mistake? It expects both: no confirmed txes and status `Expired`.

---

<details>
  <summary>More detailed analysis</summary>

This became an issue because, as part of the consolidation, in `refresh_swap_on_chain_status` the lookup

```rust
let confirmed_txs: Vec<OnchainTx> = txs
    .clone()
    .into_iter()
    .filter(|t| t.status.block_height.is_some())
    .collect();
```

was changed to

```rust
let utxos = get_utxos(bitcoin_address.clone(), txs)?;
let confirmed_txs = &utxos.confirmed;
```

which for this test is empty, because the test covers the case when, after the swap, the funds are also spent. The `get_utxo` method excludes utxos that were spent. Therefore, `confirmed_txs` is empty. Therefore, the swap status is not updated to `Expired` and the test fails.

I did this because `get_utxo` was already there just 1-2 lines above, and I wanted to better understand which of these two ways to get the confirmed txes I should use for the reverse swap. They both seemed to do the same, so I consolidated into the simplest.

I can revert this to the way it was before, but IMO it's worth it to clarify the confirmed_txes issue above which fixes the test.
</details>

ok300:
Fixed the failing test. This is ready for review.

Essentially it moves 
- the `Utxo` and `AddressUtxos` structs
- the `get_utxos` util method

from `swap.rs` to `chain.rs` because they will be used by reverse swap functions.

roeierez (approved):
LGTM

PR #134, opened 2023-03-07
Store extra payment info if no SuccessActionData found

ok300:
This ensures payments to LN Addresses that don't return a `SucessActionData` are still stored in the DB.


roeierez (approved):
LGTM

erdemyerebasmaz (approved):

PR #135, opened 2023-03-07
Store LNURL metadata in extra payment info table

ok300:
Stores metadata, which can contain text, icons or other fields from the LNURL endpoint.


roeierez (approved):
LGTM

erdemyerebasmaz (approved):
LGTM

PR #138, opened 2023-03-08
Add pay_lnurl to bindings

ok300:


roeierez (approved):
LGTM. We need to add the react native part but this can be done in a separate PR.

PR #141, opened 2023-03-10
Add react native bindings for LNURL-pay

ok300:


roeierez (approved):

dangeross (commented):
It looks like the kotlin and swift changes still need adding.

roeierez:
> It looks like the kotlin and swift changes still need adding.

Right, I missed the fact the native logic is not implemented @ok300

dangeross:
Hi @ok300, if you don't mind, I'm going to branch your PR and add the native updates for kotlin and swift.

ok300:
Hi, I was about to do that right now, but please go ahead. You'll probably do it better and faster than me.

Was struggling to see how to map the enum in the right way in the new syntaxes, without having a local compiler to give feedback if its correct or not.

ok300:
Superseded by #149

PR #145, opened 2023-03-12
Add more data to PaymentFailed event

### LINKED ISSUE #143: Missing payment reference with failed payment event
Currently the payment failure event only emits an error string without reference to the original payment. To resolve which payment has failed a client will also need a reference to the payment, be it the original payment object or payment hash.

roeierez:
This PR adds the following data to the payment failed event:
1. Node ID
2. Optional parsed bolt11 invoice.

fixed #143

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -214,7 +214,7 @@ dictionary InvoicePaidDetails {
 dictionary PaymentFailedData {
     string error;
     string node_id;
-    LNInvoice? bolt11;
+    LNInvoice? invoice;
 };
 
 [Enum]
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -211,13 +211,19 @@ dictionary InvoicePaidDetails {
     string bolt11;
 };
 
+dictionary PaymentFailedData {
+    string error;
+    string node_id;
+    LNInvoice? bolt11;
+};
+
 [Enum]
 interface BreezEvent {
     NewBlock(u32 block);
     InvoicePaid(InvoicePaidDetails details);
     Synced();
     PaymentSucceed(Payment details);
-    PaymentFailed(string error);
+    PaymentFailed(PaymentFailedData details);
 };
 
 callback interface LogStream {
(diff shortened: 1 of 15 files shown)


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -211,13 +211,19 @@ dictionary InvoicePaidDetails {
     string bolt11;
 };
 
+dictionary PaymentFailedData {
+    string error;
+    string node_id;
+    LNInvoice? bolt11;
```
dangeross, line 217: bolt11 is usually a string with the encoded payment request. Would in be clearer to rename the lnInvoice param to invoice?

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -172,6 +172,14 @@ class BreezSDKMapper {
             "bolt11": invoicePaidDetails.bolt11
         ]
     }
+
+    static func dictionaryOf(paymentFailedData: PaymentFailedData) -> [String: Any?] {
+        return [
+            "error": paymentFailedData.error,
+            "bolt11": dictionaryOf(lnInvoice: paymentFailedData.bolt11),
```
dangeross, line 179: If bolt11 is optional then `dictionaryOf(lnInvoice)` needs to handle the nil case.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
```diff
@@ -190,6 +190,14 @@ fun readableMapOf(invoicePaidDetails: InvoicePaidDetails): ReadableMap {
     )
 }
 
+fun readableMapOf(paymentFailedData: PaymentFailedData): ReadableMap {
+    return readableMapOf(
+            "error" to paymentFailedData.error,
+            "bolt11" to readableMapOf(paymentFailedData.bolt11)
```
dangeross, line 196: If bolt11 is optional then `readableMapOf(lnInvoice: LnInvoice)` needs to handle the null case.

ok300 (approved):
The Rust side LGTM.

libs/sdk-bindings/src/breez_sdk.udl
roeierez: Agree. Done

libs/sdk-react-native/ios/BreezSDKMapper.swift
roeierez: Right, missed that. Fixed.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
roeierez: Right, missed that. Fixed.

dangeross (approved):
LGTM!

PR #148, opened 2023-03-13
Update dependencies to latest & generate bindings

erdemyerebasmaz:
- Update flutter_rust_bridge & flutter_rust_bridge_codegen to 1.69.0
- Update gl-client to 0.1.6 [rs: Move the scheduler proto into a submod](https://github.com/Blockstream/greenlight/commit/928c2002d1a73eca222de7138d69e46a907d8028)
- Update lightning-invoice to 0.22.0 [Add _delta suffix to min_final_cltv_expiry](https://github.com/lightningdevkit/rust-lightning/commit/fb10fc9590dc2173cafbf5dcef3673b8dabcd386)
- Target mockito 0.31.1
- Correct a typo in sdk-cli README


roeierez:
@erdemyerebasmaz this looks good to me! 
One question is were you able to test the upgrade flow? Asking because I see there is a gl-client version that inolves upgrading the node at the backend (the c-lightning node).
I think it worth checking that we are on the safe side here before upgrading gl-client dependency especially because this kind of upgrade is not reversible.

erdemyerebasmaz:
> @erdemyerebasmaz this looks good to me!
> One question is were you able to test the upgrade flow? Asking because I see there is a gl-client version that inolves upgrading the node at the backend (the c-lightning node).
> I think it worth checking that we are on the safe side here before upgrading gl-client dependency especially because this kind of upgrade is not reversible.

Haven't tested with cli or app yet but all test cases were passing. I will test the affected flows on this c-breez branch:
- https://github.com/breez/c-breez/pull/534

roeierez (approved):
Tested & LGTM!

PR #150, opened 2023-03-16
Expose lnurlAuth & generate Dart bindings

erdemyerebasmaz:
- Make LnUrlAuthCallbackStatus compatible with Dart binding
- Expose perform_lnurl_auth via lnurl_auth in bindings
- Add documentation for lnurl auth
  - Change or add additional information as you like
- Detailed lnurlWithdraw's documentation

### DIFF
--- a/libs/sdk-core/src/lnurl/auth.rs
+++ b/libs/sdk-core/src/lnurl/auth.rs
@@ -1,5 +1,4 @@
-use crate::lnurl::auth::model::*;
-use crate::{LnUrlAuthRequestData, NodeAPI};
+use crate::{LnUrlAuthRequestData, LnUrlCallbackStatus, NodeAPI};
 use anyhow::{anyhow, Result};
 use bitcoin::secp256k1::{Message, Secp256k1};
 use bitcoin::util::bip32::ChildNumber;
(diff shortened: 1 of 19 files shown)


libs/sdk-core/src/lnurl/auth.rs
```diff
@@ -9,9 +10,35 @@ use reqwest::Url;
 use std::str::FromStr;
 use std::sync::Arc;
 
-pub type LnUrlAuthCallbackStatus = LnUrlWithdrawCallbackStatus;
+pub(crate) mod model {
```
roeierez, line 13: Does it mean that now we have duplication of this type? Also in the models.rs?

roeierez: If there isn't any difference between them can't we use one type for both? LnUrlCallbackStatus maybe?
@ok300 what do you think?

erdemyerebasmaz: > If there isn't any difference between them can't we use one type for both? LnUrlCallbackStatus maybe?
> @ok300 what do you think?

Done https://github.com/breez/breez-sdk/pull/150/commits/a6a804e63aecc203e2ad4159c673672efdbcf123

roeierez:
@erdemyerebasmaz I think we need also to implement the react native changes

libs/sdk-core/src/lnurl/auth.rs
roeierez: @dangeross this change introduce a change in the lnurl operations return type. It unified the result into one structure: `LnUrlCallbackStatus` that has the same structure for all lnurl operations.
Would you like to review that before we are moving forward?

erdemyerebasmaz:
@dangeross Latest changes can't be seen on this PR right now due to a git issue. Meanwhile you can review react native changes here https://github.com/breez/breez-sdk/commit/2c8db6f9fd6e3e5e75cadf921f2d6a867c83a951

libs/sdk-core/src/lnurl/auth.rs
dangeross: @roeierez Sure thing, I'll take a look now

dangeross (approved):

roeierez (approved):

PR #152, opened 2023-03-17
Fix react native UShort mapping

dangeross:
Fix UShort mapping for LnUrlPayRequestData param commentAllowed. 

Should this not be a boolean @roeierez?


roeierez:
> Fix UShort mapping for LnUrlPayRequestData param commentAllowed.
> 
> Should this not be a boolean @roeierez?

Although I find the name to be a bit misleading as well this is implemented according to the lnurl spec where the `commentAllowed` is the number of  characters in the comment:
https://github.com/lnurl/luds/blob/luds/12.md

roeierez (approved):

roeierez (approved):

PR #153, opened 2023-03-25
sdk-flutter: add desktop target

icota:
This is primarily to make development (on the desktop) easier. 

The 4 Linux targets in `sdk-core` feel like a bit of an overkill but to be consistent I added `x86_64-unknown-linux-gnu` even though having a simple `linux` target would be my preference.

### DIFF
--- a/libs/sdk-flutter/lib/native_toolkit.dart
+++ b/libs/sdk-flutter/lib/native_toolkit.dart
@@ -4,12 +4,24 @@ import 'bridge_generated.dart';
 
 BreezSdkCore? _breezSDK;
 
+const _libName = "libbreez_sdk_core.so";
+
+class UnsupportedPlatform implements Exception {
+  UnsupportedPlatform(String s);
+}
+
 BreezSdkCore getNativeToolkit() {
   if (_breezSDK == null) {
-    final DynamicLibrary lib = Platform.isAndroid
-      ? DynamicLibrary.open("libbreez_sdk_core.so")   // Load the dynamic library on Android
-      : DynamicLibrary.process();
-    _breezSDK = BreezSdkCoreImpl(lib);
+    if (Platform.isAndroid || Platform.isLinux) {
+      // On Linux the lib needs to be in LD_LIBRARY_PATH or working directory
+      _breezSDK = BreezSdkCoreImpl(DynamicLibrary.open(_libName));
+    } else if (Platform.isIOS || Platform.isMacOS) {
+      // iOS and macOS are statically linked
+      _breezSDK = BreezSdkCoreImpl(DynamicLibrary.process());
+    } else {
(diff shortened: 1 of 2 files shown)


roeierez:
@icota what do you mean by simple linux target? We need to be able to build for all architectures.

icota:
@roeierez I think there's no need to bloat the makefile with 32-bit archs. But if you have some reason to support those devices that's cool.

libs/sdk-flutter/lib/native_toolkit.dart
```diff
@@ -4,12 +4,15 @@ import 'bridge_generated.dart';
...
 BreezSdkCore? _breezSDK;
 
+const _libName = "libbreez_sdk_core.so";
+
 BreezSdkCore getNativeToolkit() {
   if (_breezSDK == null) {
-    final DynamicLibrary lib = Platform.isAndroid
-      ? DynamicLibrary.open("libbreez_sdk_core.so")   // Load the dynamic library on Android
-      : DynamicLibrary.process();
+    final DynamicLibrary lib = Platform.isIOS
```
roeierez, line 11: Perhaps it is better to be explicit here? e.g `if Platform.isLinux`
This way we can add Macos and Windows later.

libs/sdk-flutter/lib/native_toolkit.dart
```diff
@@ -4,12 +4,15 @@ import 'bridge_generated.dart';
...
+        ? DynamicLibrary.process() // iOS is statically linked
+        : DynamicLibrary.open(
+            Platform.isAndroid ? _libName : "breez-sdk/libs/sdk-flutter/linux/lib/$_libName");
```
roeierez, line 14: Why is this path used and not just `_libName` ?

roeierez (commented):
Thank you @icota that will sure help us running the tests on desktop and open the path to running a desktop app.
Left a few comments/questions

libs/sdk-flutter/lib/native_toolkit.dart
icota: Sure thing, makes sense.

libs/sdk-flutter/lib/native_toolkit.dart
icota: I was being lazy and hard-coded where it ended up in my project. The correct way is to set `LD_LIBRARY_PATH` env variable. I've added a comment to reflect that.

icota:
I've changed the makefile by adding one `desktop` target that defaults to `x86_64` Linux but the user can specify any target she wants with the `TARGET` variable.

I'll probably be integrating `sdk-flutter` into my project with a cargo/build.rs combo but I think this is a flexible approach for the makefile crowd.

roeierez (approved):
LGTM. Thanks Igor!

PR #155, opened 2023-03-31
Generate and publish rustdoc on main branch commits

ok300:
This PR adds a GH workflow that generates the rustdoc and publishes it to Github Pages, for every commit to the `main` branch.


roeierez (approved):
Great!

PR #156, opened 2023-04-01
Implement copyWith Config extension

roeierez:
Add copyWith extension to flutter SDK Config for easier usage of the default config. Instead of initialising all config field one now can write:

`sdk.defaultConfig().copyWith(workingDir:"/path")`


erdemyerebasmaz (approved):
LGTM

PR #157, opened 2023-04-02
Use clap as basis for cli implementation

roeierez:
Use clap crate as the basis for cli implementation which includes the following advantages over the current implementation:
1. Parsing commands
2. Generate help description from comments
3. Allow defining options and parameters using struct enums.
4. Support positional and not positional arguments

And more https://github.com/clap-rs/clap

### DIFF
--- a/tools/sdk-cli/README.md
+++ b/tools/sdk-cli/README.md
@@ -3,8 +3,9 @@
 A simple cli tool that sends commands to the sdk.
 
 ## Run
-```cargo run```
+```cargo run -- --data_dir <data directory>```
 
+The data directory is optional (defaults to the current directory) and it is used to save the node data.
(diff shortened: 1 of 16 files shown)


tools/sdk-cli/README.md
```diff
@@ -3,8 +3,9 @@
 A simple cli tool that sends commands to the sdk.
 
 ## Run
-```cargo run```
+```cargo run -- --data_dir <data directory>```
```
ok300, line 6: When trying a custom `data_dir`, the CLI still uses the default `history.txt`.

For example:

```
mkdir test-dir
cargo run -- --data_dir test-dir
```

Pressing up/down arrows after it starts cycles through the `history.txt` from the default location.

PR #168, opened 2023-04-04
Reverse swap: add dynamic statuses

ok300:
Dynamically determine the reverse swap status based on known fields and current block height.


ok300:
Rebased on the latest `reverse-swap-feature` branch, which was rebased on latest `master`

libs/sdk-core/src/models.rs
```diff
@@ -132,41 +131,84 @@ pub struct ReverseSwapPairInfo {
...
 
+    /// Boltz API failed to create the reverse swap
+    InitialError = 1,
+
     /// The HODL invoice has been paid.
-    InvoicePaid = 1,
+    InvoicePaid = 2,
+
+    /// The reverse swap has been created. Paying the HODL invoice was attempted, but failed.
+    FailedToPayInvoice = 3,
```
roeierez, line 198: @ok300  Is there a difference in behavior between FailedToPayInvoice and InitialError?
If yes then two fields are required, if no perhaps it is better to have only one field for error?

ok300: Good point, got rid of `InitialError`.

ok300:
As per our latest sync:
- I simplified the status handling
- reduced locally stored data (instead, lookup what's needed from Boltz)
- implemented the "success case" scenario

libs/sdk-core/src/breez_services.rs
```diff
@@ -473,14 +500,52 @@ impl BreezServices {
...
+                // which cannot happen now (at the beginning of the reverse swap)
+
+                // TODO It doesn't fail when trying to pay more sats than max_payable?
+                match pay_thread_res {
+                    Ok(Ok(_)) => Err(anyhow!("Payment of HODL invoice unexpectedly returned")),
+                    Ok(Err(e)) => Err(anyhow!("Failed to pay HODL invoice: {e}")),
+                    Err(e) => Err(anyhow!("Failed to wait for payment of HODL invoice: {e}"))
+                }
+            },
+            paid_invoice_res = self.poll_initial_boltz_status_transition(&created_reverse_swap_info.id) => {
```
roeierez, line 542: Perhaps it is easier to see the timeout here as another select branch rather inside the `poll_initial_boltz_status_transition` function?

roeierez (commented):
@ok300 that starts to shape up really good!
It is a work in progress but I dropped some comments and I think you can implement the sqlite storage now.
Doing so we need to split into two tables same as we did in the regular swap:
1. one table for the immutable info (so sync to other devices)
2. the other table with the cached (reproducible info).

libs/sdk-core/src/breez_services.rs
ok300: Good point, addressed in 5371ca5

libs/sdk-core/src/breez_services.rs
```diff
@@ -453,8 +455,33 @@ impl BreezServices {
...
     pub async fn reverse_swap_info(&self) -> Result<ReverseSwapPairInfo> {
-        self.btc_send_swapper.reverse_swap_info().await
+        self.btc_send_swapper.reverse_swap_pair_info().await
+    }
+
+    async fn poll_initial_boltz_status_transition(&self, id: &str) -> Result<()> {
+        // Check the status every 5 seconds, for up to 1 minute
+        // Wait slightly longer than the max duration configured, to make sure we cover the full duration
+        let wait_duration_sec = 5;
+        let wait_duration_max_sec = Config::production().payment_timeout_sec; // TODO Lookup config from SDK
```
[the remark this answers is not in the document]
ok300: Added a `node_api.config()` getter in 63ba039.

libs/sdk-core/src/boltzswap.rs
```diff
@@ -85,58 +84,163 @@ struct Pairs {
...
-                pair_hash,
-                onchain_claim_address,
+                pair_hash.clone(),
+                claim_pubkey.clone(),
                 routing_node,
             ))
             .send()
             .await?
-            .json()
+            .text()
```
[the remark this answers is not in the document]
ok300: Because `text()` lets us access the raw response body directly for debugging.

If I use `json()`, the automatic conversion to `BoltzApiReverseSwapStatus` kicks in and hides the raw json text.

ok300:
@roeierez @yaslama the PR is ready for review.

libs/sdk-core/src/persist/db.rs
```diff
@@ -205,6 +205,20 @@ impl SqliteStorage {
...
              ALTER TABLE payments_external_info ADD COLUMN lnurl_metadata TEXT;
             "),
+            M::up("
+             CREATE TABLE IF NOT EXISTS reverse_swaps (
+               id TEXT PRIMARY KEY NOT NULL,
+               created_at INTEGER NOT NULL DEFAULT CURRENT_TIMESTAMP,
+               local_preimage BLOB NOT NULL UNIQUE,
+               local_private_key BLOB NOT NULL UNIQUE,
+               destination_address TEXT NOT NULL,
+               boltz_api_status TEXT NOT NULL,
```
roeierez, line 215: We need to split (different tables) the reverse swap immutable data from the data that can be retrieved using the API and chain information.
This way we can sync the immutable data only and every app can build the dynamic data itself.
Similar to the btc_swap feature

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
```diff
@@ -473,14 +493,56 @@ impl BreezServices {
...
+                    Ok(_) => Ok(created_reverse_swap_info),
+                    Err(e) => Err(anyhow!(e))
+                }
+            }
+        }
+    }
+
+    /// Returns an optional in-progress [ReverseSwapInfo]
+    pub async fn in_progress_reverse_swap(&self) -> Result<Option<ReverseSwapInfo>> {
+        let monitored = self.btc_send_swapper.list_monitored()?;
```
roeierez, line 540: Not sure all monitored are also defined as "in progress".
"in progress" is a swap that blocks other swaps from starting and I remember we were discussing the fact that the last confirmation step should be still monitored but will not block a new swap.

libs/sdk-core/src/breez_services.rs
```diff
@@ -473,14 +493,56 @@ impl BreezServices {
...
+        info!("Created reverse swap: {created_reverse_swap_info:?}");
+
+        // Wait until one of the following happens:
+        // - trying to pay the HODL invoice explicitly fails from Greenlight
+        // - the regular poll of the Breez API detects the status of this reverse swap advanced to LockTxMempool
+        //   (meaning Boltz detected that we paid the HODL invoice)
+        // - the max allowed duration of a payment is reached
+        tokio::select! {
+            pay_thread_res = tokio::time::timeout(
+                Duration::from_secs(self.node_api.config().payment_timeout_sec as u64),
```
roeierez, line 514: It seems better to take the config from the current component rather than adding another method on the NodeAPI trait. We can pass the config downstream to the btc_send_swapper.

libs/sdk-core/src/models.rs
```diff
@@ -132,63 +135,113 @@ pub struct ReverseSwapPairInfo {
...
+    pub hodl_bolt11: String,
+    pub onchain_amount_sat: u64,
+    pub redeem_script: String,
+}
+
+impl ReverseSwapInfo {
+    /// Dynamically determine the status, based on the currently known Boltz API status for this reverse swap
+    pub fn status(&self) -> ReverseSwapStatus {
+        match &self.boltz_api_status {
+            BoltzApiReverseSwapStatus::SwapCreated => ReverseSwapStatus::Initial,
```
roeierez, line 166: Boltz uses very detailed status but I am not sure we need/want to expose that to the user. IMO we can map these statuses to :
Initial - created in the db and lock tx is not yet in mempool
InProgress - lock tx in mempool
Completed - successfully completed
Expired/Invalidated -   swap is not usable anymore and shouldn't be monitored.

Could be other options as well as long as they are simplified and not exposing the internals @yaslama @ok300 what do you think?

libs/sdk-core/src/reverseswap.rs
```diff
@@ -22,68 +33,238 @@ pub struct CreateReverseSwapResponse {
...
+            None => Err(anyhow!("Could not determine lock address type")),
+        }
+    }
+
+    pub(crate) async fn execute_pending_reverse_swaps(&self) -> Result<()> {
+        let monitored = self.refresh_monitored_reverse_swaps().await?;
+        info!("Found {} monitored reverse swaps", monitored.len());
+
+        // Depending on the new state, decide next steps and transition to the new state
+        for rs in monitored {
```
roeierez, line 228: I am missing the case where we started a reverse swap, paid for the old invoice but from some reason the remote swap provider has canceled the payment.
It seems that these swaps will be monitored until expiration and won't marked as canceled.

libs/sdk-core/src/models.rs
roeierez: @ok300 on second thought I now see that the ReverseSwapStatus is not exposed to the user so why do we need it at all. 
Can't we use internally the boltz status?

libs/sdk-core/src/breez_services.rs
roeierez, line 514: On a second thought why do we need to monitor and poll for the status here?
It seems that we only care if the lock tx is confirmed.

libs/sdk-core/src/breez_services.rs
```diff
@@ -473,14 +493,56 @@ impl BreezServices {
         pair_hash: String,
     ) -> Result<ReverseSwapInfo> {
         let routing_hop = self.lsp_info().await?;
-        self.btc_send_swapper
+        let created_reverse_swap_info = self
```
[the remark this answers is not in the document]
ok300: Good point, that is now addressed in e95c67d0dbd9d332199de58b270fe4a2f16fe7a4

libs/sdk-core/src/breez_services.rs
```diff
@@ -453,34 +453,42 @@ impl BreezServices {
...
+        }
+    }
+
+    /// Returns an optional in-progress [ReverseSwapInfo]
+    pub async fn in_progress_reverse_swap(&self) -> Result<Option<ReverseSwapInfo>> {
+        let monitored = self.btc_send_swapper.list_monitored().await?;
+        match monitored.len() {
+            x if x == 0 => Ok(None),
+            x if x == 1 => Ok(monitored.first().cloned()),
+            _ => Err(anyhow!("Found more than one ongoing reverse swaps")),
```
roeierez, line 490: We don't need to return error here. In theory there could be more than one in progress swap for example when running the node in multiple apps. IMO we can either return the first one or return a list of swaps.

libs/sdk-core/src/models.rs
```diff
@@ -132,63 +137,137 @@ pub struct ReverseSwapPairInfo {
...
+        let status = match &self.boltz_api_status {
+            BoltzApiReverseSwapStatus::SwapCreated => {
+                debug!("Looking up payment via NodeAPI");
+                let invoice = parse_invoice(&self.hodl_bolt11)?;
+                let payment_hash = hex::decode(invoice.payment_hash)?;
+                match node_api.lookup_payment(payment_hash).await? {
+                    None => ReverseSwapStatus::FailedToPayInvoice, // Was aborted before pay() could throw error
+                    Some(p) => match p.pending {
+                        true => ReverseSwapStatus::Initial, // Payment is in progress
+                        false => ReverseSwapStatus::FailedToPayInvoice, // The node reports it as failed
```
roeierez, line 188: Why failed? Not pending might also mean payment succeed.

libs/sdk-core/src/greenlight.rs
```diff
@@ -306,6 +306,23 @@ impl NodeAPI for Greenlight {
         })
     }
 
+    async fn lookup_payment(&self, payment_hash: Vec<u8>) -> Result<Option<crate::Payment>> {
```
roeierez, line 309: Perhpas it is better to exten the sync process to also synchronize pending payments to the db and then just use the db existing lookpu for payment by hash?
The advantage is that we won't need to add another method to the NodeAPI trait and won't need to pass it as a dependency to the reverse swapper. Also we need the pending payments in the db as need to return them in the list_payments API.

roeierez (changes requested):
Providing some early feedback.

libs/sdk-core/src/breez_services.rs
ok300: Good point, addressed it in 5c0376deebe8e3d3aef9c1c178612aa45a624041

This also answers an open question I added earlier today about exactly this
> Clarify meaning of "in progress rev swap": does it covered all monitored stauses? if not (e.g. ClaimTxMempool = in progress), then the API call to `list_in_progress_rev_swaps` can theoretically return more than one?

libs/sdk-core/src/models.rs
ok300: Normally yes, but for the HODL invoice not the case because the preimage is not shared.

This invoice is settled (not pending anymore) when we publish the ClaimTx, which reveals the preimage to Boltz, which they then use to settle the invoice.

libs/sdk-core/src/greenlight.rs
ok300: @roeierez  That's true. But if we do it that way, then we have to do a `sync` before checking a rev swap payment status, in case it's in progress, to get the latest payment state from GL.

For my test wallet, `sync` takes 2.8-3 seconds, `lookup_payment` about 0.7 seconds. Is that an acceptable delay?

Should I then keep using `lookup_payment` to determine the rev swap status, and inject / update this pending payment in our DB?

Or should I just add pending payments to `sync` and switch to using that, despite the extra delay?

roeierez: The extra delay is because the sync Api and the connection pooling are not optimal. This will change in the future so I think it is cleaner to use sync and query theocal db despite the extra temporary delay.
Let's make sure though that we call sync once for all addreses that we refresh and not once per address

libs/sdk-core/src/breez_services.rs
ok300: Done 0e5371710b514f5ca3c96e0ac8e7ba31b498509b

libs/sdk-core/src/greenlight.rs
ok300: Done: d887e9deae88bf0810995c0653a7ee7442b84c78

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -46,6 +50,12 @@ impl SqliteStorage {
         Ok(())
     }
 
+    fn delete_pending_payments(&self) -> Result<usize> {
+        self.get_connection()?
+            .execute("DELETE FROM payments WHERE pending = true", [])
```
roeierez, line 55: Need to make sure we only delete lightning payments. Pending can be true also for closed channel.

libs/sdk-core/src/models.rs
```diff
@@ -132,63 +134,144 @@ pub struct ReverseSwapPairInfo {
...
+}
 
-    // dynamic data
-    pub status: ReverseSwapStatus,
+impl ReverseSwapInfo {
+    /// Dynamically determine the status, based on the cached Boltz API status.
+    ///
+    /// In case of [BoltzApiReverseSwapStatus::SwapCreated], the payment status is looked up on the LN node,
+    /// to distinguish between a reverse swap that is still ongoing and one where the HODL payment failed.
+    pub async fn status(&self, persister: Arc<SqliteStorage>) -> Result<ReverseSwapStatus> {
```
roeierez, line 175: It seems that we don't need the ReverseSwapStatus as a struct that maps from boltz status as it is only used internally. How about using directly the boltz status and get rid of this struct?

libs/sdk-core/src/persist/transactions.rs
ok300: Done, tested: e7f9d6ab472baf3f6246d342ee0bfa7a426c881d

libs/sdk-core/src/breez_services.rs
ok300: Done 50e9b212dc471c50a93480de16b24f2acbdefd3a

libs/sdk-core/src/models.rs
ok300: We still need other information (in addition to the Boltz status) to determine the status of a rev swap.

For example:

- Boltz status `swap.created` + paying the HODL invoice failed = we should track it as failed
- Boltz status `swap.created` + paying the HODL invoice succeded = we should track it as in progress

I added a simplified `ReverseStatusSwap` in 0c344187ca0e7f004e2d2428caac30585522a969 :

https://github.com/breez/breez-sdk/blob/0c344187ca0e7f004e2d2428caac30585522a969/libs/sdk-core/src/models.rs#L309-L326

libs/sdk-core/src/models.rs
ok300: This is addressed on two levels:

First, the possible values for `ReverseSwapStatus` were simplified.

Second, the `BreezServices` doesn't return the full `ReverseSwapInfo` to the user, containing internals like the HODL invoice or the redeem script. Instead, a stripped-down version of that is returned in 7f3d5a25fbc0853722f407b4b100f11e03dbe30d

https://github.com/breez/breez-sdk/blob/7f3d5a25fbc0853722f407b4b100f11e03dbe30d/libs/sdk-core/src/models.rs#L311-L317

libs/sdk-core/src/greenlight.rs
```diff
@@ -537,17 +537,20 @@ async fn pull_transactions(
...
         .into_inner();
     debug!("list payments: {:?}", payments);
-    // construct the payment transactions
-    let sent_transactions: Result<Vec<crate::models::Payment>> = payments
+    // construct the payment transactions (pending and complete)
+    let outbound_transactions: Result<Vec<crate::models::Payment>> = payments
         .payments
         .into_iter()
-        .filter(|p| p.created_at as i64 > since_timestamp && p.status() == PayStatus::Complete)
+        .filter(|p| {
```
roeierez, line 544: We now add to the db also pending payments and this means that any code that was using the `get_payment_by_hash` function is now have different meaning because now also pending payments may return.
Specificaly the regular swap use that function so I think we should check for the right state there.

libs/sdk-core/src/models.rs
```diff
@@ -132,63 +142,276 @@ pub struct ReverseSwapPairInfo {
...
+    /// In case of [BoltzApiReverseSwapStatus::SwapCreated], the payment status is looked up on the LN node,
+    /// to distinguish between a reverse swap that is still ongoing and one where the HODL payment failed.
+    pub(crate) async fn get_dynamic_breez_status(
+        &self,
+        persister: Arc<SqliteStorage>,
+    ) -> Result<ReverseSwapStatus> {
+        let status = match &self.cache.boltz_api_status {
+            BoltzApiReverseSwapStatus::SwapCreated => {
+                let invoice = parse_invoice(&self.hodl_bolt11)?;
+                match persister.get_payment_by_hash(&invoice.payment_hash)? {
```
roeierez, line 195: Does this behave right on restart as well? if we paid a hodl invoice, synced and the payment is pending in the db, then restarted. In this case after the initial sync the payment is pending immediately? Just want to make sure that greenlight doesn't have some time to detect this pending payment otherwise we might mark this payment as canceled while there is still ongoing payment.

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -270,6 +270,28 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
         FROM old_swaps;
 
        DROP TABLE old_swaps;            
+       ",
+
+        // reverse_swaps      holds the immutable data
+        // reverse_swaps_info holds the cached data, which can be reconstructed by any client
+       "
+       CREATE TABLE IF NOT EXISTS reverse_swaps (
```
roeierez, line 278: This should be created on the sync db

roeierez (changes requested):
Almost there, a few comments.

libs/sdk-core/src/breez_services.rs
```diff
@@ -453,34 +453,49 @@ impl BreezServices {
...
+                        amount_sat,
+                        onchain_recipient_address,
+                        pair_hash,
+                        self.lsp_info().await?.pubkey,
+                    )
+                    .await;
+
+                if create_res.is_err() {
+                    warn!("Creating reverse swap failed, syncing to refresh and persist status");
+                    self.sync().await?;
```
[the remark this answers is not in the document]
ok300: As an indirect way to persist the new rev swap status. The event listener reacts to `sync` by calling `refresh_monitored_reverse_swaps()`.

May be better to instead just persist the status of this one failed rev swap. Will adjust.

libs/sdk-core/src/models.rs
ok300: Should be the case, but I only tested and confirmed the failure case, because the initial HODL payment fails with the current mempool fees.

Will try again and post here if this works in the successful case too.

libs/sdk-core/src/boltzswap.rs
```diff
@@ -85,58 +84,169 @@ struct Pairs {
...
+    /// Call and parse response as per https://docs.boltz.exchange/en/latest/api/#getting-status-of-a-swap
+    ///
+    /// #### Errors
+    ///
+    /// This method returns an error for  HTTP or connection errors (404 not found, 400 bad request,
+    /// 502 server error, etc).
+    ///
+    /// Boltz API errors (e.g. providing an invalid ID arg) are returned as a successful response of
+    /// type [BoltzApiCreateReverseSwapResponse::BoltzApiError]
+    async fn get_dynamic_boltz_status(&self, id: String) -> Result<BoltzApiReverseSwapStatus> {
```
yaslama, line 215: Why not get_boltz_status?

libs/sdk-core/src/reverseswap.rs
```diff
@@ -22,68 +37,342 @@ pub struct CreateReverseSwapResponse {
...
+    }
+
+    /// Updates the state of monitored reverse swaps in the cache table. This includes the blocking
+    /// reverse swaps as well, since the blocking statuses are a subset of the monitored statuses.
+    async fn refresh_monitored_reverse_swaps(&self) -> Result<()> {
+        let to_check = self.list_monitored().await?;
+        for rs in to_check {
+            let id = rs.id.clone();
+            let api = self.reverse_swapper_api.clone();
+            let fresh_boltz_status = api.get_dynamic_boltz_status(id.clone()).await?;
```
yaslama, line 340: Why do we need to get the status from Boltz? In theory, we can deduce the status from the LN payments and the onchain information. Perhaps we can keep the status only for logging.

libs/sdk-core/src/greenlight.rs
ok300: Is this an acceptable solution? 87c0f1624ecfd8142c4d13ebee6bb84c2b59d290

roeierez: Yes.

libs/sdk-core/src/persist/migrations.rs
ok300: Done 371a123f61403d0b86770ddd39a44210e1db9da5

libs/sdk-core/src/reverseswap.rs
ok300: Done: 6303df600cb561dc8b0948cd2dae96c606271992

libs/sdk-core/src/boltzswap.rs
ok300: Done in 7dacf9abcd3a63a49ae961fae1a2d02647c2d491

libs/sdk-core/src/reverseswap.rs
```diff
@@ -22,68 +36,438 @@ pub struct CreateReverseSwapResponse {
...
+    }
+
+    /// Determine the reverse swap status, based on the the states of the lockup and claim tx
+    pub(crate) async fn get_breez_status(
+        &self,
+        rsi: &ReverseSwapInfo,
+    ) -> Result<ReverseSwapStatus> {
+        let payment_hash = rsi.get_preimage_hash().to_hex();
+        let is_invoice_pending = self.persister.get_payment_by_hash(&payment_hash)?.is_some();
+        let status = match is_invoice_pending {
```
roeierez, line 410: The `is_invoice_pending` sounds a bit misleading because the `get_payment_by_hash` API returns also completed payments...
I can see cases where we restart the node and the payment is completed but the status is still `Initial` at this point.
In that case we will end in the Cancelled state which is wrong. Also I don't see how we advance the state to `CompletedSeen` or `CompletedConfirmed` since the branch expects false result.

In general once we have a payment (either pending or completed) we should never move to canceled state.

**Suggestion:**

We enter into this method only with the following states (Initial, InProgress, CompletedSeen) and our job is to advance the state if needed.

So for every existing payment (pending or completed) we can do the following based on the status:

--Initial
  |-- if `is_lockup_tx_in_mempool_or_confirmed` transition to InProgress otherwise do nothing.

--InProgress or CompletedSeen
  |----fetch claim tx.
  |------if in mempool -> transition to CompletedSeen
  |------if confirmed -> transition to CompletedConfirmed
  
  If we don't get a result for get_payment_by_hash then it is the only state where we transition to Cancelled.
  
  What do you think @ok300

roeierez (changes requested):

libs/sdk-core/src/reverseswap.rs
ok300: This code (which checks the dynamic / fresh status) is only called in `refresh_reverse_swap()`, which is only called to update the status of monitored swaps in `refresh_monitored_reverse_swaps`.

The monitored swaps are already only of the states you mentioned (Initial, InProgress, CompletedSeen). So this is only called for those states.

Assuming status Initial, a node restart would indeed cancel a rev swap, because it would interrupt GL in the middle of paying it. This is the critical initial phase where the payment either succeeds or fails, anything before that lands "in limbo", e.g. cancelled.

But assuming status InProgress or subsequent, then the pending payment is still found, and the next lookups cover the claim tx.

So maybe I'm missing something, but IMO your suggestion is already implemented. I should restructure the methods for better readability.

Let's have a call tomorrow to clarify.

ok300: Addressed in 65d07dd1be48609174d2f794fb2c37528b573090

roeierez (approved):
LGTM! Let's wait with the merge until we agree on the public API naming.
Excited to see it merged soon.

PR #171, opened 2023-04-07
Add build step for Swift package

### LINKED ISSUE #137: Add build step to create iOS framework
This should make iOS devs life a lot easier to work with.

cnixbtc:
Resolves #137.

This adds a build step to the Makefile in `libs/sdk-bindings` which will create a fully configured Swift package in `libs/sdk-bindings/bindings-swift`.

This package can then be added to any Xcode project by simply dropping the folder `bindigns-swift` folder into the project. It can also be used by other Swift packages as a local dependency. The approach is inspired by the way [bdk](https://github.com/bitcoindevkit/bdk-ffi/tree/master/bdk-swift) ships its Swift bindings.

Another cool thing that bdk does is automatically push a fully configured Swift package to a [separate repository](https://github.com/bitcoindevkit/bdk-swift) using GitHub Actions. It would be cool if we could do that to. That way the Breez SDK could be used without anyone needing to build it manually. Instead, people could just add the repo URL to their project and Xcode/SPM would do the rest.

I'd say we do this step by step, though. This PR only adds the build step to create a local Swift package. Imo, setting up and configuring the hosted package in a separate repo can be a follow up PR if we agree that it makes sense.

---

👉 I removed the old `swift-ios` and `swift-darwin` build steps. Let me know if I should add them back.

### DIFF
--- a/libs/sdk-bindings/README.md
+++ b/libs/sdk-bindings/README.md
@@ -20,17 +20,11 @@ make init
 ### Swift
 
 ```
-make swift-ios
+make bindings-swift
 ```
 
-This will generate all the artifacts needed to for an iOS app to start writing code that uses breez sdk in swift.
-All files are generated in the bindings/swift-ios folder.
-We also provides the same binding for mac os by running the following command:
-
-```
-make swift-darwin
-```
-The above will generate the artifacts in the ffi/swift-darwin folder.
+This will produce a fully configured Swift Package in `bindings-swift/`.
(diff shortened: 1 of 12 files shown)


roeierez:
@cnixbtc This is great and I will go over it in more details later.
We should add the removed build steps as they are used in the sdk-react-native build.
Also it would be great if you can push to a branch in breez repo so the CI will run.

cnixbtc:
Closing in favor of #172.

PR #172, opened 2023-04-07
Add build step for Swift package

### LINKED ISSUE #137: Add build step to create iOS framework

cnixbtc:
Resolves #137.

This adds a build step to the Makefile in `libs/sdk-bindings` which will create a fully configured Swift package in `libs/sdk-bindings/bindings-swift`.

This package can then be added to any Xcode project by simply dropping the folder `bindigns-swift` folder into the project. It can also be used by other Swift packages as a local dependency. The approach is inspired by the way [bdk](https://github.com/bitcoindevkit/bdk-ffi/tree/master/bdk-swift) ships its Swift bindings.

Another cool thing that bdk does is automatically push a fully configured Swift package to a [separate repository](https://github.com/bitcoindevkit/bdk-swift) using GitHub Actions. It would be cool if we could do that to. That way the Breez SDK could be used without anyone needing to build it manually. Instead, people could just add the repo URL to their project and Xcode/SPM would do the rest.

I'd say we do this step by step, though. This PR only adds the build step to create a local Swift package. Imo, setting up and configuring the hosted package in a separate repo can be a follow up PR if we agree that it makes sense.

---

👉 I removed the old `swift-ios` and `swift-darwin` build steps. Let me know if I should add them back.

### DIFF
--- a/libs/sdk-bindings/makefile
+++ b/libs/sdk-bindings/makefile
@@ -9,11 +9,12 @@ init:
 	rustup target add aarch64-apple-ios-sim
 	#rustup target add armv7-apple-ios armv7s-apple-ios i386-apple-ios ## deprecated
 	rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
+	@if [ $$(uname) == "Darwin" ] ; then cargo install cargo-lipo ; fi
 	cargo install cbindgen
 	cargo install cargo-ndk
 
 ## all: Compile iOS, Android
-all: bindings-swift kotlin
+all: swift-ios swift-darwin bindings-swift kotlin
 
 ios-universal: $(SOURCES)		
 	mkdir -p ../target/ios-universal/release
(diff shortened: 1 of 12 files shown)


libs/sdk-bindings/makefile
```diff
@@ -14,21 +14,21 @@ init:
...
 	cargo build --release --target x86_64-apple-ios ;\
-	lipo -create -output ../target/ios-universal/release/libbreez_sdk_bindings.a ../target/aarch64-apple-ios/release/libbreez_sdk_bindings.a ../target/x86_64-apple-ios/release/libbreez_sdk_bindings.a	
-
+	cargo build --release --target aarch64-apple-ios-sim ;\
+	lipo -create -output ../target/ios-universal/release/libbreez_sdk_bindings.a ../target/aarch64-apple-ios-sim/release/libbreez_sdk_bindings.a ../target/x86_64-apple-ios/release/libbreez_sdk_bindings.a
 
 darwin-universal: $(SOURCES)
 	mkdir -p ../target/darwin-universal/release
-	cargo lipo --release --targets aarch64-apple-darwin		
-	cargo lipo --release --targets x86_64-apple-darwin
```
cnixbtc, line 29: Could we switch this to using `cargo build` directly, like we do for the iOS targets? 🤔

roeierez: I think you are probably right about it.

roeierez:
> I'd say we do this step by step, though. This PR only adds the build step to create a local Swift package. Imo, setting up and configuring the hosted package in a separate repo can be a follow up PR if we agree that it makes sense.

Sure thing! here is the repo for the following PR: https://github.com/breez/breez-sdk-swift

roeierez (approved):
LGTM!

roeierez:
@cnixbtc @dangeross once the swift package is uploaded to a public repo it would be great also to change the react native plugin to use that url. This way we can publish the plugin without uploading the binaries

dangeross:
> @cnixbtc @dangeross once the swift package is uploaded to a public repo it would be great also to change the react native plugin to use that url. This way we can publish the plugin without uploading the binaries

Sounds good. Any plans for a similar kotlin library so the react native module can be completely without binaries? Would be nice to automate it also based off of breez-sdk-swift releases but all the bridging and mapping is currently manually written.

roeierez:
@dangeross Yes we have also plan to do the same for android. @cnixbtc is leading this and going to handle it after we finish with swift.

PR #175, opened 2023-04-13
fix log stream

### LINKED ISSUE #174: Setting log stream does not stream log entries
Binaries compiled for kotlin and swift do not seem to be producing log entries from the breez-sdk.

```
breez_sdk.setLogStream(logStream: BreezSDKLogStream(emitter: self))
```

```
class BreezSDKLogStream: NSObject, LogStream {
    var emitter: RCTEventEmitter
    
    static let emitterName: String = "breezSdkLog"
    
    init(emitter: RCTEventEmitter) {
        self.emitter = emitter
    }
    
    func log(l: LogEntry) {
        self.emitter.sendEvent(withName: BreezSDKListener.emitterName,
                               body: [
                                "line":  l.line,
                                "level": l.level
                               ])
    }
}
```

roeierez:
This fixed #174 

Two fixes here:
1. Setting the filter level to trace (before it was Info and prevented from lower severity logs to be sent).
2. Prevent the internal uniffi logs from being added to the stream ( to prevent infinite loop with the callback).

### DIFF
--- a/libs/sdk-bindings/src/uniffi_binding.rs
+++ b/libs/sdk-bindings/src/uniffi_binding.rs
@@ -14,39 +14,42 @@ use breez_sdk_core::{
     RecommendedFees, RouteHint, RouteHintHop, SuccessActionProcessed, SwapInfo, SwapStatus, Symbol,
     UnspentTransactionOutput, UrlSuccessActionData,
 };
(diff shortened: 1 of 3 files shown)


dangeross (approved):
LGTM!

PR #179, opened 2023-04-17
Update docs (SDK bindings README, rustdoc) for C#

ok300:
Add steps and instructions for the C# bindings.

### DIFF
--- a/libs/sdk-bindings/README.md
+++ b/libs/sdk-bindings/README.md
@@ -1,12 +1,16 @@
 # breez-sdk-bindings
 
 This project provides bindings for breez-sdk to various languages.
-Currently supported languges are kotlin & swift.
+Currently supported languges are kotlin, swift and C#.
 For kotlin & swift, we are using [uniffi](https://github.com/mozilla/uniffi-rs).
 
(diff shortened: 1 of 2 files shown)


libs/sdk-bindings/README.md
```diff
@@ -1,12 +1,16 @@
...
 For kotlin & swift, we are using [uniffi](https://github.com/mozilla/uniffi-rs).
 
-## prerequisite
-* set the ANDROID_NDK_HOME env variable to your sdk home folder
-* install cargo-ndk to provide simplified android build: ```cargo install cargo-ndk``` 
+## Prerequisites
+* When building for Android:
+  * Set the ANDROID_NDK_HOME env variable to your sdk home folder
+  * Install `cargo-ndk` to provide simplified android build: ```cargo install cargo-ndk``` 
+* When building C# bindings:
```
roeierez, line 11: I don't think we need this specific section for C# because there is a target in the make file that does it and you added documentation for it bellow

ok300: You're right. Done.

PR #180, opened 2023-04-17
Separate swaps table to swaps & swaps_info

roeierez:
This PR is a preparation for the multi app sync feature which separates the swaps table into two tables:
1. swaps - contains only the immutable data that has all the information (hash, keys) needed to continue the swap process. This table should be synchronized across devices
2. swaps_info - contains cached data (chain info, paid amount, etc..)  - This table shouldn't be synced.

The swaps persistency layer still provides the same API but manages the data according to the new schema.


ok300 (approved):

ok300:
`SwapInfo` contains data from both tables (persistent data in `swaps` and cached data in `swaps_info`).

Would it make sense to reflect that in the struct as well? Like for example:

* group the dynamic data from `SwapInfo` into a field `SwapInfo.cache` of type `SwapInfoCache`
* rename `swaps` table to `swaps_data` (for persistent fields in `SwapInfo)
* rename `swaps_info` table to `swaps_cache` (for `SwapInfoCache`)

roeierez:
> `SwapInfo` contains data from both tables (persistent data in `swaps` and cached data in `swaps_info`).
> 
> Would it make sense to reflect that in the struct as well? Like for example:
> 
> * group the dynamic data from `SwapInfo` into a field `SwapInfo.cache` of type `SwapInfoCache`
> * rename `swaps_info` table to `swaps_cache` (for `SwapInfoCache`)

You might be right but honestly I am not sure as the split is internal and the user will always be interested in all fields.
Specifically this PR is dealing with schema changes and I prefer to leave the API and exported types as is for this refactoring to minimize the moving parts. I prefer to discuss this further after merging.

PR #181, opened 2023-04-19
Proper handle absolute and relative payments fee

roeierez:
Here we fix the fee handling according to cln behavior.
It turns out that we can't pass both maxfeepercent and max_fee to the rpc call, only one of them (https://lightning.readthedocs.io/lightning-pay.7.html)

The strategy is in the case the user has configured max_fee we calculate the fees according to the percentage and if it exceeds the max_fee we will use the max_fee.
Otherwise we will use the percentage.

This is related to https://github.com/breez/c-breez/pull/544

### DIFF
--- a/libs/sdk-core/src/greenlight.rs
+++ b/libs/sdk-core/src/greenlight.rs
@@ -336,6 +336,32 @@ impl NodeAPI for Greenlight {
     ) -> Result<crate::models::Payment> {
         let mut client = self.get_client().await?;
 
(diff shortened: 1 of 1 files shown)


ok300 (approved):

ademar111190 (approved):
LGTM just one question, is it possible to pass `None` instead of setting `maxfeepercent = 0.0`? Both ways are ok for me.

PR #182, opened 2023-04-22
Pass config to RN initServices

dangeross:
This PR adds a wrapper for defaultConfig and changes the initServices bridge call to pass an application defined config

### DIFF
--- a/libs/sdk-react-native/src/index.tsx
+++ b/libs/sdk-react-native/src/index.tsx
@@ -19,6 +19,11 @@ const BreezSDK = NativeModules.BreezSDK
 
 const BreezSDKEmitter = new NativeEventEmitter(BreezSDK)
 
+export enum EnvironmentType {
+    PRODUCTION = "production",
+    STAGING = "staging"
+}
+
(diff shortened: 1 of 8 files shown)


roeierez (approved):

libs/sdk-react-native/src/index.tsx
```diff
@@ -92,6 +97,18 @@ export type BitcoinAddressData = {
...
+export type Config = {
+    breezserver: string
+    mempoolspaceUrl: string
+    workingDir: string
+    network: Network
+    paymentTimeoutSec: number
+    defaultLspId?: string
+    apiKey?: string
+    maxfeeSat?: number
+    maxfeepercent: number
```
ok300, line 109: nit: shouldn't this be in camel case `maxfeePercent`?

dangeross: This is a direct copy of the naming from the SDK

ok300: Good point, then all good

PR #188, opened 2023-04-27
Remove absolute maxfee config option.

roeierez:
This parameter doesn't make sense as part of global config as in lightning fees are proportional to the payment amount which is different in each payment.
After this omit we are left with `maxfee_percent` (relative fee) and we plan to add `exemptfee_sats` for small payments (https://lightning.readthedocs.io/lightning-pay.7.html)

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -37,8 +37,7 @@ dictionary Config {
     Network network;
     u32 payment_timeout_sec;
     string? default_lsp_id;
-    string? api_key;
-    u64? maxfee_sat;
+    string? api_key;    
     f64 maxfee_percent;
 };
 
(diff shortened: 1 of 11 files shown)


ok300:
Why doesn't `maxfee_sat` make sense?

Even if the LN fee is proportional, some could still want to set an absolute limit, like "max fee 5%, but no more than 1000 sat".

Or am I missing something? IMO that's a useful scenario to support.

roeierez:
> Why doesn't `maxfee_sat` make sense?
> 
> Even if the LN fee is proportional, some could still want to set an absolute limit, like "max fee 5%, but no more than 1000 sat".
> 
> Or am I missing something? IMO that's a useful scenario to support.

I meant it doesn't make sense as a global config. It makes sense as  per payment API.

ok300 (approved):

ok300 (approved):

PR #189, opened 2023-04-27
Buy Bitcoin with Moonpay

ademar111190:
Related: https://github.com/breez/c-breez/pull/556

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1,9 +1,27 @@
+use std::cmp::max;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use anyhow::{anyhow, Result};
+use bip39::*;
+use bitcoin_hashes::{sha256, Hash};
+use tokio::runtime::Runtime;
+use tokio::sync::{mpsc, Mutex};
+use tokio::time::{sleep, Duration};
+use tonic::codegen::InterceptedService;
+use tonic::metadata::{Ascii, MetadataValue};
+use tonic::service::Interceptor;
+use tonic::transport::{Channel, Uri};
+use tonic::{Request, Status};
+
 use crate::chain::{ChainService, MempoolSpace, RecommendedFees};
 use crate::fiat::{FiatCurrency, Rate};
 use crate::greenlight::Greenlight;
 use crate::grpc::channel_opener_client::ChannelOpenerClient;
 use crate::grpc::fund_manager_client::FundManagerClient;
 use crate::grpc::information_client::InformationClient;
+use crate::grpc::signer_client::SignerClient;
 use crate::grpc::PaymentInformation;
 use crate::input_parser::LnUrlPayRequestData;
 use crate::invoice::{add_routing_hints, parse_invoice, LNInvoice, RouteHint, RouteHintHop};
@@ -20,24 +38,11 @@ use crate::models::{
     FiatAPI, GreenlightCredentials, LnUrlCallbackStatus, LspAPI, Network, NodeAPI, NodeState,
     Payment, PaymentDetails, PaymentType, PaymentTypeFilter, SwapInfo, SwapperAPI,
 };
+use crate::moonpay::MoonPayApi;
 use crate::persist::db::SqliteStorage;
 use crate::swap::BTCReceiveSwap;
-use crate::{LnUrlAuthRequestData, LnUrlWithdrawRequestData, PaymentResponse};
-use anyhow::{anyhow, Result};
-use bip39::*;
-use bitcoin::hashes::{sha256, Hash};
-use std::cmp::max;
-use std::str::FromStr;
-use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
-use tokio::runtime::Runtime;
-use tokio::sync::{mpsc, Mutex};
-use tokio::time::{sleep, Duration};
-use tonic::codegen::InterceptedService;
-use tonic::metadata::{Ascii, MetadataValue};
-use tonic::service::Interceptor;
-use tonic::transport::{Channel, Uri};
-use tonic::{Request, Status};
+use crate::BuyBitcoinProvider::MoonPay;
+use crate::{BuyBitcoinProvider, LnUrlAuthRequestData, LnUrlWithdrawRequestData, PaymentResponse};
 
 /// Trait that can be used to react to various [BreezEvent]s emitted by the SDK.
 pub trait EventListener: Send + Sync {
@@ -133,6 +138,7 @@ pub struct BreezServices {
     node_api: Arc<dyn NodeAPI>,
     lsp_api: Arc<dyn LspAPI>,
     fiat_api: Arc<dyn FiatAPI>,
+    moonpay_api: Arc<dyn MoonPayApi>,
     chain_service: Arc<dyn ChainService>,
     persister: Arc<SqliteStorage>,
     payment_receiver: Arc<PaymentReceiver>,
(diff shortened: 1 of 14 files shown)


roeierez (changes requested):
Looks good!

libs/sdk-core/src/moon_pay/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub(crate) mod moon_pay_api;
```
roeierez, line 1: We should use moonpay instead of moon_pay (also in the folder name).

libs/sdk-core/src/moon_pay/moon_pay_config.rs
```diff
@@ -0,0 +1,25 @@
+#[derive(Clone)]
+pub struct MoonPayConfig {
```
roeierez, line 2: I would rather not expose the configuration (not even out of the moonpay module) and not to the user.
All the values are hard coded eventually and I am not sure we want the user to change them.
The only exception that we need to think about is how to deal with the API key.

libs/sdk-core/src/breez_services.rs
```diff
@@ -834,6 +854,10 @@ impl BreezServicesBuilder {
                 .fiat_api
                 .clone()
                 .unwrap_or_else(|| breez_server.clone()),
+            moon_pay_api: Arc::new(Mutex::new(MoonPayApi::new(
+                self.config.moon_pay_config.clone(),
+                Box::new(breez_server.get_signer_client().await?),
```
roeierez, line 859: Other API traits are implemented in BreezServer (LspAPI, FiatAPI, SwapperAPI). For consistency perhaps it is better to add implementatino of the signer to BreezServer the same way and pass it to the MoonPayApi ?

libs/sdk-core/src/moon_pay/moon_pay_api.rs
```diff
@@ -0,0 +1,23 @@
...
+    config: MoonPayConfig,
+    signer: Box<dyn MoonPayUrlSigner>,
+}
+
+impl MoonPayApi {
+    pub fn new(config: MoonPayConfig, signer: Box<dyn MoonPayUrlSigner>) -> Self {
+        Self { config, signer }
+    }
+
+    pub async fn sign_moon_pay_url(&mut self, swap_info: &SwapInfo) -> Result<String> {
```
roeierez, line 17: Why do we need mut here? This forced the code to use mutex in BreezServices and I don't see why we need mut reference. If we omit the mut then we wont need the mutex.

libs/sdk-core/src/models.rs
```diff
@@ -159,6 +173,18 @@ impl Config {
             default_lsp_id: Some(String::from("ea51d025-042d-456c-8325-63e430797481")),
             api_key: None,
             maxfee_percent: 0.5,
+            moon_pay_config: MoonPayConfig {
```
roeierez, line 176: I prefer to not expose the config, we actually don't need a config now as we use all the values hard coded.

roeierez (changes requested):

libs/sdk-core/src/moon_pay/moon_pay_config.rs
ademar111190: I'm thinking of using the same way as breez mobile, what do you think?

libs/sdk-core/src/moonpay/moonpay_api.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::SwapInfo;
+
+#[tonic::async_trait]
+pub trait MoonPayApi: Send + Sync {
+    async fn sign_moon_pay_url(&self, url_data: &dyn MoonPayUrlData) -> Result<String>;
+}
+
+#[tonic::async_trait]
+impl MoonPayApi for BreezServer {
+    async fn sign_moon_pay_url(&self, url_data: &dyn MoonPayUrlData) -> Result<String> {
```
roeierez, line 14: Not sure we need the MoonPayUrlData trait. Looks simpler just to pass the two values exlicitly.

libs/sdk-core/src/moonpay/moonpay_url_signer.rs
```diff
@@ -0,0 +1,152 @@
...
+    async fn sign_moon_pay_url(
+        &self,
+        moonpay_api_key: &str,
+        wallet_address: &str,
+        max_quote_currency_amount: &str,
+    ) -> Result<String>;
+}
+
+#[tonic::async_trait]
+impl MoonPayUrlSigner for SignerClient<tonic::transport::Channel> {
```
roeierez, line 19: Seems not related. Why do we need this implementation for the signer?  We are using breez server to sign don't we?

roeierez (changes requested):
Code looks good. Droped some comments/questions

roeierez (changes requested):
Code looks good. Droped some comments/questions

libs/sdk-core/src/moonpay/moonpay_url_signer.rs
ademar111190: Yes, we are using the breez server. Under the hood, its SignerClient is the one signing, so I thought having this trait would be a good thing.

ademar111190: I started removing the `MoonPayUrlSigner`, but I got a problem; without this layer, I cannot unit test unless changing how the breez server is made by adding a trait to `get_signer_client` and implementing my version of it on the mocks version. What do you prefer? The `MoonPayUrlSigner` for the `SignerClient` or a `SignerClientProvider` for the Breez server? Or another approach I'm not aware of.

ademar111190: Not because the `get_signer_client(&self)`, I'll implement it so you can check what I'm talking about

ademar111190: No success; I couldn't find a way to unit test without adding this layer.

ademar111190: The test I'm talking about is this one: https://github.com/breez/breez-sdk/pull/189/files#diff-a20deab6b2aa49948975cb34e64974e4fef825884eb03b024717519c82d410ceR72-R131

I'm creating a mock web server to simulate the URL signature, so I need a layer or mechanism to provide a mock SignerClient.

I've succeeded in running the test without this layer, but the test was executing a real request to the breez server, which is a bad idea for a unit test. Ideally, we mock the server or mock the client.

Do you see another way of doing it?

libs/sdk-core/src/moonpay.rs
```diff
@@ -0,0 +1,117 @@
...
+        currency_code: String::from("btc"),
+        color_code: String::from("#055DEB"),
+        redirect_url: String::from("https://buy.moonpay.io/transaction_receipt?addFunds=true"),
+        enabled_payment_methods: String::from(
+            "credit_debit_card,sepa_bank_transfer,gbp_bank_transfer",
+        ),
+    }
+}
+
+async fn sign_moonpay_url(
```
roeierez, line 31: Perhaps rename to `create_moonpay_url` ? This function doesn't really sign.

libs/sdk-core/src/moonpay.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[tonic::async_trait]
+pub(crate) trait MoonPayApi: Send + Sync {
+    async fn sign_moonpay_url(&self, swap_info: &SwapInfo) -> Result<String>;
+}
+
+#[tonic::async_trait]
+impl MoonPayApi for BreezServer {
+    async fn sign_moonpay_url(&self, swap_info: &SwapInfo) -> Result<String> {
```
roeierez, line 59: Also here the signing is internal, perhaps `buy_bitcoin_url` is preferred?

libs/sdk-core/src/moonpay.rs
ademar111190: Oh sure, I did move the function to see if the idea works and completely forgot to revisit and refine it.

roeierez (approved):
LGTM!

PR #202, opened 2023-05-05
Fix iOS flutter makefile bug

VictorAppiah123:
The command: `make ios` have this response: `make: Nothing to be done for ios.`
The problem was inside the _makefile_ where the alias command was missing.


roeierez (approved):
Thanks you!

PR #207, opened 2023-05-11
Improve RN error responses

dangeross:
Propagate Breez SDK errors back over the react native bridge


roeierez (approved):
LGTM. Thanks again @dangeross

PR #209, opened 2023-05-16
Implement golang binding

dangeross:
This PR adds a golang binding using the uniffi-rs extension https://github.com/NordSecurity/uniffi-bindgen-go


roeierez (approved):
LGTM. This is realy great!

PR #211, opened 2023-05-17
fix React Native example app

roeierez:
This PR fixes the example app to reference the react native plugin from the node_modules directory and not from the parent directory.
Updated the docs also to include a step that builds the parent directory node modules.


dangeross (approved):

ademar111190 (approved):
Thanks @roeierez 
Maybe we can create later a showcase repo where the examples use the SDK available without the need to build locally; for this case, I think what you did is the correct approach.

PR #213, opened 2023-05-18
Add RN build that uses embedded frameworks

roeierez:
Turns out the only way to support all architectures in React Native is to use Fat framework.
Curently React Native doesn't suppot swift package so we have to embed the framework inside.
I changed the react native build target for ios to copile the swift package and then copy the resulted XCFramework. Also the Ract Native xcode project now add the framework to the linker input.
I had to rename the BreezSDK to RNBreezSDK as it conflicts with the swift package source.


dangeross (changes requested):
I'm making suggestions here, as I can't do it in the files.

In libs/sdk-react-native/ios/RNBreezSDK.m
```suggestion
-   @interface RCT_EXTERN_MODULE(BreezSDK, RCTEventEmitter)
+   @interface RCT_EXTERN_MODULE(RNBreezSDK, RCTEventEmitter)
```

In libs/sdk-react-native/src/index.tsx
```suggestion
-   const BreezSDK = NativeModules.BreezSDK
-       ? NativeModules.BreezSDK
+   const BreezSDK = NativeModules.RNBreezSDK
+       ? NativeModules.RNBreezSDK
```

roeierez:
@dangeross this was exactly what I was looking for. Thanks a lot! Just pushed.

dangeross (approved):
Looks good! I'll retest later today in Satimoto

PR #226, opened 2023-05-26
flutter: add credentials to registerNode

JssDWt:
Greenlight requires a single-use invite code to register a node now. Add the invite code as a parameter to the `registerNode` call, or Greenlight credentials.


JssDWt:
Note that vscode decided to format a bunch of things on save. Are there specific formatting settings you guys use for flutter?

roeierez:
> 

`dartfmt` with default settings

ok300:
A while ago @erdemyerebasmaz suggested
```
dart format --fix -l 110 libs
```
from the SDK root folder.

PR #228, opened 2023-05-29
Update `vout.value` datatype from `u32` to `u64`

ok300:
The `Vout.value` field, which shows the amount of sats in an output, was of type `u32`.

This has a range of 0..2^32-1 = 4294967296 (4.294.967.296 sats), so any utxo of more than 4.29 BTC would overflow.

I increased this to `u64` and propagated the change.


roeierez (approved):
LGTM

PR #232, opened 2023-05-31
Add new argument to `send_onchain` for claim tx feerate

ok300:


libs/sdk-core/src/persist/migrations.rs
```diff
@@ -383,5 +383,8 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
 
          DROP TABLE old_payments_external_info;
         ",
+        "
+        ALTER TABLE sync.reverse_swaps ADD COLUMN sat_per_vbyte INTEGER NOT NULL DEFAULT 10;
```
roeierez, line 387: @ok300  Why default to 10? Why default at all if it is mandatory also by the user?

ok300: Because this migration upgrades an existing table, and by adding a new mandatory column it needs a default value.

ok300: I rebased on #240 , so since we're changing the migration order, I used the opportunity to replace the
```
ALTER TABLE sync.reverse_swaps ADD COLUMN sat_per_vbyte INTEGER NOT NULL DEFAULT 10;
```
here with just a new column in the rev swap table:

https://github.com/breez/breez-sdk/commit/307a042a0b9436e919a6ba065e6204ee1b7b2872#diff-1c18a7114601088370c028a024e166f533e16bfe5127d76afe45923815cf9278R377

This solves the question of detault value. Since the new column now is created with the table, it needs no default.

PR #234, opened 2023-05-31
Add bindings for `send_onchain`, `in_progress_reverse_swaps`

ok300:


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -288,6 +296,13 @@ dictionary ReverseSwapPairInfo {
     u64 fees_claim;
 };
 
+dictionary SimpleReverseSwapInfo {
```
roeierez, line 299: SimpleReverseSwapInfo is not a good name for the end user.
Better to use ReverseSwapInfo or ReverseSwap

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
ok300: Done, renamed
- `ReverseSwapInfo` => `FullReverseSwapInfo`
- `SimpleReverseSwapInfo` => `ReverseSwapInfo`

roeierez (approved):

PR #237, opened 2023-06-02
Change unit of RecommendedFees to u64

ok300:


roeierez (changes requested):
@ok300 we need also the sweep API to use the same type (u32 or u64)

JssDWt:
@ok300 @roeierez 
Is this a good time to change chainfee related things to a more generally preferred unit, like sat/kw? I think using weight units instead of vbyte is a preference thing (I'd personally pick weight units, because that represents the actual fees better). The precision is a different thing. Sat/vbyte is not very fine grained, going from 1 sat/vbyte to 2 satt/vbyte is a big step. I think generally you want to be able to express 1.1 sat/vbyte for example. sat/kw adresses that, because 1 sat/kw == 0.004 sat/vbyte.

roeierez:
Yes I agree.

roeierez:
@JssDWt @ok300 but we need to be consistent also with the recommended_fees API that right now returns sat per byte. Let's think about it and also consider the changes in c-breez regarding this.

roeierez:
> @roeierez yes `sweep` already uses `u64`.
> 
> The only remaining difference is `sats_per_byte` vs `sats_per_vbyte`.

Right, sorry I missed that.

roeierez:
@ok300 I see now greenlight  has the option to pass `sat_per_kw`.
So let's also change the sweep API to receive `sat_per_vbyte` instead of sat_per_byte and pass it to greenlight with the right fee option (we need to duplicate with 250 to convert to `sat_per_kw`).
IMO once this is done we are consistent with both naming and fee rate in our on-chain API and then we can speak about @JssDWt  suggestion.

I think it is better for us to have a discussion on that next week before implementing as this change has more implications.

ok300:
I opened a separate PR #242 for changing `sweep` to use `sat_per_vbyte`.

This PR is only about the `u32` -> `u64` change in `RecommendedFees`.

PR #242, opened 2023-06-05
Update sweep feerate unit to sats per vbyte

ok300:


roeierez (approved):
LGTM.
Note that we need a change in c-breez because of the parameter name change.

dangeross:
Please also update is libs/sdk-react-native/ios/ `RNBreezSDK.m` and `RNBreezSDK.swift`

PR #243, opened 2023-06-05
Add react native bindings for reverse swap methods

ok300:


libs/sdk-react-native/src/index.tsx
```diff
@@ -568,6 +592,21 @@ export const refund = async (swapAddress: string, toAddress: string, satPerVbyte
     return response
 }
 
+export const fetchReverseSwapFees = async (): Promise<ReverseSwapPairInfo> => {
```
roeierez, line 595: @ok300  These methods should be implemented in BreezSDK (android & iOS modules).

ok300: Done, although by hand + not sure how to validate the syntax.

Are there any CLI commands I could run to compile or check the syntax of the react native modules we have here?

dangeross (changes requested):

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -497,6 +497,33 @@ class BreezSDKMapper {
             "lastRedeemError": swapInfo.lastRedeemError
         ]
     }
+
+    static func dictionaryOf(reverseSwapPairInfo: ReverseSwapPairInfo) -> [String: Any?] {
```
dangeross, line 501: Can be `[String: Any]` as there are no optional values

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -497,6 +497,33 @@ class BreezSDKMapper {
...
+            "min": reverseSwapPairInfo.min,
+            "max": reverseSwapPairInfo.max,
+            "feesHash": reverseSwapPairInfo.feesHash,
+            "feesPercentage": reverseSwapPairInfo.feesPercentage,
+            "feesLockup": reverseSwapPairInfo.feesLockup,
+            "feesClaim": reverseSwapPairInfo.feesClaim
+        ]
+    }
+
+    static func dictionaryOf(reverseSwapInfo: ReverseSwapInfo) -> [String: Any?] {
```
dangeross, line 512: Can be `[String: Any]` as there are no optional values

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -497,6 +497,33 @@ class BreezSDKMapper {
...
+        return [
+            "min": reverseSwapPairInfo.min,
+            "max": reverseSwapPairInfo.max,
+            "feesHash": reverseSwapPairInfo.feesHash,
+            "feesPercentage": reverseSwapPairInfo.feesPercentage,
+            "feesLockup": reverseSwapPairInfo.feesLockup,
+            "feesClaim": reverseSwapPairInfo.feesClaim
+
```
dangeross, line 520: Think this is a copy/paste error

libs/sdk-react-native/src/index.tsx
```diff
@@ -83,6 +83,14 @@ export enum SwapStatus {
     EXPIRED = "expired"
 }
 
+export enum ReverseSwapStatus {
+    INITIAL = "initial",
+    INPROGRESS = "inprogress",
+    CANCELLED = "cancelled",
+    COMPLETEDSEEN = "completedseen",
+    COMPLETEDCONFIRMED = "completedconfirmed"
```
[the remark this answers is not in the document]
roeierez: Yes that's the convention in typescript.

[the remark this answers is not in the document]
dangeross: So try to keep the value case as the url file (needs changes in the native mapping and typescript) and the key case as typescript e.g. `COMPLETED_CONFIRMED = "CompletedConfirmed"`

libs/sdk-react-native/ios/BreezSDKMapper.swift
ok300: Good catch, removed.

libs/sdk-react-native/ios/BreezSDKMapper.swift
ok300: Done

libs/sdk-react-native/ios/BreezSDKMapper.swift
ok300: Done

libs/sdk-react-native/src/index.tsx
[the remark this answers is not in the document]
ok300: I tried the same case as done in https://github.com/breez/breez-sdk/pull/254/files#diff-2d928afd2f658c921620fc5a1f8017f1618f00463b9deae70e779f90ce308801

namely:

```js
export enum ReverseSwapStatus {
    INITIAL = "initial",
    IN_PROGRESS = "in_progress",
    CANCELLED = "cancelled",
    COMPLETED_SEEN = "completed_seen",
    COMPLETED_CONFIRMED = "completed_confirmed"
}
```

which worked.

Should I keep it this way? Or remove the underscores from both the SDK enums and the tsx enum?

libs/sdk-react-native/ios/RNBreezSDK.m
```diff
@@ -195,6 +195,24 @@ @interface RCT_EXTERN_MODULE(RNBreezSDK, RCTEventEmitter)
...
+)
+
+RCT_EXTERN_METHOD(
+    sendOnchain: (NSInteger*)amountSat
+    onchainRecipientAddress: (NSString*)onchainRecipientAddress
+    pairHash: (NSString*)pairHash
+    satPerVbyte: (NSInteger*)satPerVbyte
+    rejecter: (RCTPromiseRejectBlock)reject
+)
+
```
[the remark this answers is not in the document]
ok300: Sorry I couldn't get my iOS simulator to work. Xcode keeps giving me this error:

sdk-ios-error

I tried different images (iOS 13, 14, 15), to re-install Xcode, clear the cache, and several other tips found online, but getting the same error.

I'm getting it even if I want to start the simulator independent of the example app, so its likely due to my environment (macOS 11.7.6 / Xcode 13.2.1).

Could one of you please check and update the swift files with what is missing? Thanks in advance.

(Or if you have other ideas I could try, let me know, I will give it a shot.)

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -548,4 +568,14 @@ class BreezSDKMapper {
...
         }
     }
+
+    static func valueOf(reverseSwapStatus: ReverseSwapStatus) -> String {
+        switch(reverseSwapStatus) {
+        case .initial: return "initial"
+        case .inprogress: return "inprogress"
+        case .cancelled: return "cancelled"
+        case .completedseen: return "completedseen"
+        case .completedconfirmed: return "completedconfirmed"
```
dangeross, line 578: These should match the equivalent kotlin enum values which are the lowercase of this enum:
```
enum class ReverseSwapStatus {
    INITIAL,IN_PROGRESS,CANCELLED,COMPLETED_SEEN,COMPLETED_CONFIRMED;
}
```

libs/sdk-react-native/ios/RNBreezSDK.m
```diff
@@ -195,6 +195,24 @@ @interface RCT_EXTERN_MODULE(RNBreezSDK, RCTEventEmitter)
...
+RCT_EXTERN_METHOD(
+    inProgressReverseSwaps: (RCTPromiseResolveBlock)resolve
+    rejecter: (RCTPromiseRejectBlock)reject
+)
+
+RCT_EXTERN_METHOD(
+    sendOnchain: (NSInteger*)amountSat
+    onchainRecipientAddress: (NSString*)onchainRecipientAddress
+    pairHash: (NSString*)pairHash
+    satPerVbyte: (NSInteger*)satPerVbyte
```
dangeross, line 212: These `NSInteger` types should be `NSUInteger` here in `ios/RNBreezSDK.m`. In the swift implementation you can use the `UInt64` type for the same params.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -433,6 +433,39 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
...
+            var inProgressReverseSwaps = getBreezServices().inProgressReverseSwaps()
+            promise.resolve(readableArrayOf(inProgressReverseSwaps))
+        } catch (e: SdkException) {
+            e.printStackTrace()
+            promise.reject(TAG, e.message ?: "Error calling inProgressReverseSwaps", e)
+        }
+    }
+
+    @ReactMethod
+    fun sendOnchain(amountSat: ULong, onchainRecipientAddress: String, pairHash: String, satPerVbyte: ULong, promise: Promise) {
```
dangeross, line 459: The RN bridge doesn't like `ULong`, instead use `Double` then convert in the method e.g. `amountSat.toULong()`

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -433,6 +433,39 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
...
+        try {
+            var response = getBreezServices().sendOnchain(amountSat, onchainRecipientAddress, pairHash, satPerVbyte)
```
dangeross, line 461: nit: consistent naming: some places it's `amountSats` other places `amountSat`

libs/sdk-react-native/ios/RNBreezSDK.m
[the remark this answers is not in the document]
ok300: Done.

libs/sdk-react-native/ios/BreezSDKMapper.swift
ok300: Done.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
ok300: Done.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKListener.kt
```diff
@@ -16,6 +16,9 @@ class BreezSDKListener(private val emitter: RCTDeviceEventEmitter): EventListene
             is BreezEvent.PaymentFailed -> emitter.emit(emitterName, readableMapOf("type" to "paymentFailed", "data" to readableMapOf(e.details)))
             is BreezEvent.PaymentSucceed -> emitter.emit(emitterName, readableMapOf("type" to "paymentSucceed", "data" to readableMapOf(e.details)))
             is BreezEvent.Synced -> emitter.emit(emitterName, readableMapOf("type" to "synced"))
+            is BreezEvent.BackupStarted -> emitter.emit(emitterName, readableMapOf("type" to "backupStarted"))
+            is BreezEvent.BackupSucceeded -> emitter.emit(emitterName, readableMapOf("type" to "backupSucceded"))
```
dangeross, line 20: backupSucceeded

PR #244, opened 2023-06-05
Add windows targets

dangeross:
Add targets for windows (x86_64), golang and csharp

Any ideas how to test this @roeierez?


roeierez:
@dangeross I think we can configure the current CI of github actions to run in a matrix that contains windows or add a CI that runs the tests in windows

roeierez (approved):

PR #254, opened 2023-06-15
fix RN example and moonpay enum value casing

roeierez:
Better not use camel casing in enum values as it behaves differently in iOS/Android mappers and bindings.
Also the example now is fixes and works on both platforms.


libs/sdk-react-native/example/App.js
```diff
@@ -86,10 +87,11 @@ const App = () => {
             let deviceKey = await getSecureItem(GREENLIGHT_DEVICE_KEY_STORE)
             let deviceCert = await getSecureItem(GREENLIGHT_DEVICE_CERT_STORE)
 
-            if (!deviceKey) {
+            if (deviceKey == null) {
                 const greenlightCredentials = await recoverNode(Network.BITCOIN, seed)
 
                 addLine("recoverNode", null)
+                setSecureItem(MNEMONIC_STORE, mnemonic)
```
dangeross, line 94: This is already done above if the mnemonic is not defined

dangeross (commented):
Looks good, just one comment

libs/sdk-react-native/example/App.js
roeierez: Thanks, leftover from my experiments to recover. Addressed.

dangeross (approved):

PR #260, opened 2023-06-20
Add backup functionality to react native

roeierez:
Draft until Greelight is fixed and we will be able to test it and pass the CI.

### DIFF
--- a/libs/sdk-react-native/ios/BreezSDKMapper.swift
+++ b/libs/sdk-react-native/ios/BreezSDKMapper.swift
@@ -152,7 +152,7 @@ class BreezSDKMapper {
         }
     }
     
-    static func dictionaryOf(backupStatus: BackupStatus) -> [String: Any] {
+    static func dictionaryOf(backupStatus: BackupStatus) -> [String: Any?] {
         return [
             "backedUp": backupStatus.backedUp,
             "lastBackupTime": backupStatus.lastBackupTime,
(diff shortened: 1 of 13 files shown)


libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -152,6 +152,13 @@ class BreezSDKMapper {
         }
     }
     
+    static func dictionaryOf(backupStatus: BackupStatus) -> [String: Any] {
```
dangeross, line 155: I think this should be `[String: Any?]` as lastBackupTime is optional/nullable

libs/sdk-react-native/src/index.tsx
```diff
@@ -388,6 +395,11 @@ export type UnspentTransactionOutput = {
     reservedToBlock: number
 }
 
+export type BackupStatus = {
+    backedUp: boolean
+    lastBackupTime: number
```
dangeross, line 400: `lastBackupTime?: number`

libs/sdk-react-native/ios/BreezSDKMapper.swift
roeierez: You probably right although I didn't get any complain from the compiler and it worked.

roeierez: fixed

libs/sdk-react-native/src/index.tsx
roeierez: fixed

dangeross (approved):

PR #263, opened 2023-06-20
Use BreezSDK pod as iOS dependency of RN plugin

cnixbtc:
[WIP] This PR needs to be resolved first: https://github.com/breez/breez-sdk-swift/pull/9

To test this, I published a version of updated RN plugin under my personal npm namespace and integrated it in a fresh RN project. I then checked if I could successfully build and use the `mnemonicToSeed` interface. That worked. 🙌


roeierez:
@cnixbtc Thanks a lot!
One question: will we be able to build locally the plugin without publishing a swift package and use in the example app?
The reason is because during development we want to test the example app with a built version without publishing.

cnixbtc:
> @cnixbtc Thanks a lot! One question: will we be able to build locally the plugin without publishing a swift package and use in the example app? The reason is because during development we want to test the example app with a built version without publishing.

Good question. The ask is to be able to develop the RN plugin / the RN example app with a local build of the BreezSDK without it needing to be published as a CocoaPod first right? I think that is a fair concern. I don't think this can be done comletely seemlessly, though. I guess that this would probably involve manually changing a line in the RN plugins podspec to change the source from where the BreezSDK is loaded (as a local, vendored framework as we had it until now or pulled remotely from the CocoaPods trunk). 

I will look into this tonight.

roeierez:
@cnixbtc changing a line in the podspec or something similar sounds reasonable. Just wanted you to take that into consideration

cnixbtc:
I think I've found a way to make it work but I still need to test it. Will report back here.

cnixbtc:
Alright I think everything should work now. I've tested:

- A fresh RN project as well as the example project when using a local build of the SDK via `vendored_framework` in the plugin's podspec
- The "new" way of pulling in the `BreezSDK` pod instead of the locally built SDK.

I'd like to test this once more once we release a new version of the pod that has the latest change (lowering the deployment target https://github.com/breez/breez-sdk-swift/pull/9) in just to be sure. I'd also like to add a sentence or two to the plugins docs explaining the new setup. I'd be great to get some feedback on the approach already, though. I don't expect those things to change much.

I'll add CI changes to publish the plugin to npm in a separate PR!

libs/sdk-react-native/ios/RNBreezSDK.swift
```diff
@@ -1,4 +1,42 @@
 import Foundation
+#if canImport(BreezSDK)
+import BreezSDK
+#else
+enum BreezSDK {
```
cnixbtc, line 5: This is a bit odd. It works though and doesn't affect the production release. Only local dev setups. If anyone has a better idea to solve this let me know! 🙏 😄

roeierez: @cnixbtc I think if we import a module in swift we can use the functions without the module prefix so we don't need to maintain this proxy enum. Did you try that?

dangeross: I actually think this is a pretty cool approach 👏

dangeross (approved):
Looks great!

libs/sdk-react-native/ios/RNBreezSDK.swift
dangeross: > @cnixbtc I think if we import a module in swift we can use the functions without the module prefix so we don't need to maintain this proxy enum. Did you try that?

Would the module function names not conflict with the class function names?

cnixbtc: > @cnixbtc I think if we import a module in swift we can use the functions without the module prefix so we don't need to maintain this proxy enum. Did you try that?

The module is already explicitly imported. In the locally-build Breez SDK scenario both the SDK and the RN plugin are the same module: `breez_sdk`.

Screenshot 2023-06-22 at 21 35 02

---

The problem is pretty much what @dangeross said:

> Would the module function names not conflict with the class function names?

That's correct. We need to disambiguate e.g., `mnemonicFromSeed` in `BreezSDK.swift` from `mnemonicFromSeed` in `RNBreezSDK.swift`:

Screenshot 2023-06-22 at 21 36 49

---

That's why we have to explicitly prefix the module name to make this clear. The simplest way to do this would be to rename the RN plugin module to `BreezSDK` from `breez_sdk` but we cannot do this since then the module name would clash with the `BreezSDK` CocoaPod in the production scenario.

I hope this makes sense, let me know if not. Took me quite a while to figure it out myself. 😅

roeierez: Understood. Another alternative comes to my mind is to refer to a localy built pod package (which have the same namespace as production) is it possible?

cnixbtc: Oh, I like that idea! Will try that out tomorrow.

cnixbtc: An alternative podspec for local development would work as well. The only inconvenient thing I noticed while testing was that in development mode (when the locally built Breez SDK is part of the RN plugin), we'll get this warning in the plugin's Xcode project:

Screenshot 2023-06-25 at 11 10 51

Let me know which approach you guys would prefer. I'm slightly erring towards the alternative podspec since that would drop the need to manually update the namespace proxy in case something changes (which probably wouldn't happen often anyway—but still...).

roeierez: @cnixbtc I am tending towards using local podspec and agree with the reason you mentioned (not having to maintain the proxy).
What are the local changes needed in order to run in dev mode?

cnixbtc: Nice. I've made the change and also added some developer docs. Basically you do this:

You start out with the prod podspec being active:

- `breez_sdk.podspec`

and the dev podspec being disabled:

- `BreezSDK.podspec.dev`

To switch to dev mode you need to rename the files:

- `breez_sdk.podspec.prod`
- `BreezSDK.podspec`

The dev/prod file endings can be whatever. Whatever ends in `podspec` will be taken by React Native. Imo it would be cool if you could control what podspec react native uses that in a better way (e.g., via a config) but a quick search turned up no results. I think the file-renaming is an acceptable way of doing it for now, though.

roeierez: I agree.

roeierez (approved):

libs/sdk-react-native/ios/RNBreezSDK.swift
dangeross: Looks good

PR #264, opened 2023-06-20
RN android threaded service

dangeross:
This PR reworks the android implementation of the native module to run the Breez SDK on a separate thread as a service. 

I'll leave it as Draft while doing on device testing.

### DIFF
--- a/libs/sdk-core/src/lib.rs
+++ b/libs/sdk-core/src/lib.rs
@@ -20,8 +20,9 @@
 //! ```ignore
 //! let mnemonic = Mnemonic::new(Words12, English);
 //! let seed = Seed::new(&mnemonic, "");
+//! let invite_code = Some("...")
 //!
-//! let creds = BreezServices::register_node(Network::Bitcoin, seed.as_bytes().to_vec()).await?;
+//! let creds = BreezServices::register_node(Network::Bitcoin, seed.as_bytes().to_vec(), None, invite_code).await?;
 //! let sdk = BreezServices::init_services(
 //!         BreezServices::default_config(EnvironmentType::Production),
 //!         seed.to_vec(),


roeierez:
@dangeross thanks a lot for this!
One question is why did you decide to implement that using a Service rather than multithreading approach?
As far as I understand both approaches will fulfill the use case of offloading work to the background freeing the UI thread and the Service added value is in different use case when the app needs some more background processing time in the background even if the Activity is not shown.

dangeross:
I wanted to try the service because I thought (1) it might be beneficial to have a service when making payments between apps on the same device, the OS is less likely to kill a service when the app be moved to the background. (2) Slightly selfishly on my part, I wanted to test if it helps with apps (like Satimoto) running payments in the background, for example being triggered by a background task and running a headless (no UI) payment. In theory the service needs less time to initialise, sync and perform the payment.

roeierez:
The Service runs in the main thread of the app so you still need to create threads for handling the sdk operations. 
In that sense it is another layer that is only used to keep the app running when the user leaves it. I also don't think it takes less time to initialize (unless it is already initialized and running but the OS will terminate it after several minutes in the background eventually).

I think the user case of completing the payment even if the use leaves the app (or receiving) is a good one but I am not sure the service belongs to the sdk... and if it does then maybe it should be part of the android aar package?
At the sdk level I am more in favor of a simple thread pool that executes the sdk functions which solves exactly what we want (offloading from the main thread).
BTW for paying in the background and keeping the app alive even if the user leaves the app there is a better option than a service that you can implement easily at the app level: WorkManager (https://developer.android.com/topic/libraries/architecture/workmanager)
For React native maybe you want to take a look at this? https://github.com/infuse-us/react-native-headless-work-manager
So eventually from the app level you will run a task that just call the` breez_sdk.send_payment ` This task will be kept alive even in headless mode for a reasonable time and allow the payment to complete.
We use the work manager in Breez mobile and also in c-breez now implementing the receive case when the app is closed.
What do you think?

dangeross:
> The Service runs in the main thread of the app so you still need to create threads for handling the sdk operations. In that sense it is another layer that is only used to keep the app running when the user leaves it. 

The service is setup to run in it's own process, so doesn't run on the main UI thread. It's defined in [AndroidManifest.xml](https://github.com/breez/breez-sdk/pull/264/files#diff-56c83828a30177db6a3dfc109f65410575bf62f9f61d3709c8987178fba42887). I already see a huge difference running it on a device.

I'll look into the work-manager, I've been using [react-native-background-fetch](https://github.com/transistorsoft/react-native-background-fetch) and [react-native-queue](https://github.com/billmalarky/react-native-queue) to manage queuing background tasks for both android and iOS, trying to give a consistent user experience for both. It's not so easy to wake the app up to make frequent payments if you're not streaming audio or something in the background. I think best case from the background tasks if the OS is heavily hibernating apps, the tasks run at 15 minute intervals for 30 seconds. Push notifications might help, though the OS might not give network access (had that issue before) or suppress notifications. I read Pheonix wallet from Acinq implements custom notification handling to avoid the suppression. I'll dig into the c-breez app to see how you use the work manager.

> I think the user case of completing the payment even if the use leaves the app (or receiving) is a good one but I am not sure the service belongs to the sdk... and if it does then maybe it should be part of the android aar package?
> At the sdk level I am more in favor of a simple thread pool that executes the sdk functions which solves exactly what we want (offloading from the main thread).

Happy to keep this experimental for now and rework it into a thread pool model if you prefer.

PR #272, opened 2023-06-21
RN android thread pool

dangeross:
This PR changes the android implementation of the native module to use an executor thread pool to offload long-running operations.


libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -147,12 +233,23 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
...
+                data.putString("data", serialize(readableMapOf("status" to "ok")))
+            } catch (e: SdkException) {
+                e.printStackTrace()
+                message.what = MSG_ERROR
+                data.putString("error", e.message ?: "Error calling start")
+            }
+
+            message.data = data
+            responseHandler.sendMessage(message)
         }
```
dangeross, line 253: Am I on the right track @roeierez? The executor would run this block to start the Breez SDK in a thread from the thread pool.

dangeross:
I'm currently testing this in a Satimoto build, zero UI lockups so far

PR #277, opened 2023-06-25
Associate lightning payments with original payment amount

roeierez:
fixed https://github.com/breez/c-breez/issues/568
This PR fixes the `fee_msat` field in payments that required to open a channel.
In that case the fee is not a regular lightning payment but is calculated by:
`payer_amount - incomoing_amount`
For the sdk to be able to calculate this we need to associate the payer_amount with the payment hash.

`Sync changes`
The above association should be synced between devices/apps so a new table was added open_channel_payment_info.
The sync db migration are were split from the main db migration starting from this point and are running on sync as well.

### DIFF
--- a/libs/sdk-core/src/persist/sync.rs
+++ b/libs/sdk-core/src/persist/sync.rs
@@ -91,6 +91,12 @@ impl SqliteStorage {
 
     pub(crate) fn import_remote_changes(&self, remote_storage: &SqliteStorage) -> Result<()> {
         let sync_data_file = remote_storage.sync_db_path();
+        match SqliteStorage::migrate_sync_db(sync_data_file.clone()) {
+            Ok(_) => {}
+            Err(e) => {
+                log::error!("Failed to migrate sync db, probably local db is older than remote, skipping migration: {}", e);
+            }
+        }
 
         let mut con = self.get_connection()?;
         let tx = con.transaction()?;
@@ -163,6 +169,18 @@ impl SqliteStorage {
             [],
         )?;
 
+        // sync remote swap_refunds table
+        tx.execute(
+            "
+        INSERT INTO sync.open_channel_payment_info
+         SELECT
+          payment_hash,
+          payer_amount_msat
+         FROM remote_sync.open_channel_payment_info
+         WHERE payment_hash NOT IN (SELECT payment_hash FROM sync.open_channel_payment_info);",
+            [],
+        )?;
+
         tx.commit()?;
         con.execute("DETACH DATABASE remote_sync", [])?;
 
@@ -212,6 +230,11 @@ fn test_sync() {
     let remote_storage = SqliteStorage::new(test_utils::create_test_sql_dir());
     remote_storage.init().unwrap();
     remote_storage.insert_swap(remote_swap_info).unwrap();
+
+    remote_storage
+        .insert_open_cannel_payment_info("123", 100000)
+        .unwrap();
+
     remote_storage
         .import_remote_changes(&local_storage)
         .unwrap();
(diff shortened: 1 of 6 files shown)


libs/sdk-core/src/persist/sync.rs
```diff
@@ -163,6 +169,18 @@ impl SqliteStorage {
...
 
+        // sync remote swap_refunds table
+        tx.execute(
+            "
+        INSERT INTO sync.open_channel_payment_info
+         SELECT
+          payment_hash,
+          payer_amount_msat
+         FROM remote_sync.open_channel_payment_info
+         WHERE payment_hash NOT IN (SELECT payment_hash FROM sync.open_channel_payment_info);",
```
JssDWt, line 180: These records should probably not be correlated by the payment hash, but by some kind of internally created checking_id. Amount + payment hash is already better. [source](https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-June/003983.html#:~:text=The%20mitigation%20is%20quite%20simple.%20Backends%20should%20either%20use%20self%2Dgenerated%20unique%20%22checking%20id%27s%22%20for%20looking%20up%20internal%20payments%20or%20use%20additional%20checks%20to%20make%20sure%20that%20the%20invoice%20details%20have%20not%20been%20messed%20around%20with%20(e.g.%2C%20asserting%20amount(A)%20%3D%3D%20amount(B)).)

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -90,6 +90,28 @@ impl SqliteStorage {
         Ok(())
     }
 
+    /// Inserts payer amount for invoices that requries openning a channel.
+    pub fn insert_open_cannel_payment_info(
```
JssDWt, line 94: nit

```suggestion
    /// Inserts payer amount for invoices that require opening a channel.
    pub fn insert_open_channel_payment_info(
```

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -204,6 +235,12 @@ impl SqliteStorage {
             data.ln_address = row.get(10)?;
         }
 
+        // In cae we have a record of the open channel fee, let's use it.
```
JssDWt, line 238: nit 
```suggestion
        // In case we have a record of the open channel fee, let's use it.
```

JssDWt (commented):
Looking good.

libs/sdk-core/src/persist/sync.rs
roeierez: This is internal associated by the user and not involving any third party or  assumptions like were mentioned in this case.
The issue there, as I understand it, that they associated a payment hash with an internal payment (god knows what that is but certainly not LN payment) which they use to credit the account.
In our case it is only to associate an already existing invoice (with a unique payment hash) to its already deduced payer amount for fee calculations. Do you see any concrete problematic use case here?

JssDWt: Yeah that's a specific case. But I think in general the payment hash doesn't identify a payment, and we should be careful not to identify the payment with a payment hash in our software. 

I don't see any problematic use case here now, but I can see how it could develop into a problematic case if software is written on top of this.

roeierez: I get your point @JssDWt but in our case it is not used to identify a payment. It is used to associated data to an invoice.
Invoices are indexed by a unique id in the sdk storage (which seems OK isn't it?) therefore I see it as a natural step to associate this data using the unique id in this specific case.

ok300 (approved):

libs/sdk-core/src/persist/sync.rs
JssDWt: And the unique id is the payment hash you mean? 
I don't see a problem now, but we have to be careful to never do stuff like "ah, I have an invoice here with payment hash x, so that must be the one from the unique index in my database".

yaslama (approved):

JssDWt (approved):

PR #278, opened 2023-06-25
Use Breez SDK from Maven repository

cnixbtc:
I would've assumed that this is all that needs to be done to make the Breez SDK available to the plugin without having a locally-built version on my machine. However, when I run the example app, I get the following error:

```
Could not determine the dependencies of task ':app:processDebugResources'.
> Could not resolve all task dependencies for configuration ':app:debugRuntimeClasspath'.
   > Could not find breez_sdk:bindings-android:0.1.0.
     Searched in the following locations:
       - file:/Users/daniel/Developer/bitcoin/breez-sdk/libs/sdk-react-native/example/node_modules/react-native/android/breez_sdk/bindings-android/0.1.0/bindings-android-0.1.0.pom
       - file:/Users/daniel/Developer/bitcoin/breez-sdk/libs/sdk-react-native/example/node_modules/jsc-android/dist/breez_sdk/bindings-android/0.1.0/bindings-android-0.1.0.pom
       - https://repo.maven.apache.org/maven2/breez_sdk/bindings-android/0.1.0/bindings-android-0.1.0.pom
       - https://dl.google.com/dl/android/maven2/breez_sdk/bindings-android/0.1.0/bindings-android-0.1.0.pom
       - https://www.jitpack.io/breez_sdk/bindings-android/0.1.0/bindings-android-0.1.0.pom
     Required by:
         project :app > project :@breeztech_react-native-breez-sdk
         project :app > project :breeztech_react-native-breez-sdk
```

👉 Somehow gradle is not seeing our repository and therefore doesn't find the Breez SDK.

What's interesting is that when I add our maven dependency to the example's project-level gradle file (`example/android/build.gradle`), everything works perfectly. But of course it would be nicer if the consumer or our RN plugin wouldn't have to do any change to their project.

I'm not sure why the repository declaration in the plugins gradle file is not seen. @dangeross do you have any ideas of why that could be? Is there some special configuration in the example project that prevents this?

### DIFF
--- a/libs/sdk-bindings/bindings-android/jitpack.yml
+++ b/libs/sdk-bindings/bindings-android/jitpack.yml
@@ -1,0 +1,4 @@
+jdk:
+  - openjdk11
+before_install:
+  - cd libs/sdk-bindings/bindings-android
--- a/libs/sdk-react-native/android/build.gradle
+++ b/libs/sdk-react-native/android/build.gradle
@@ -56,14 +56,17 @@ repositories {
 
 allprojects {
     repositories {
+        maven {
+            url "https://mvn.breez.technology/releases"
+        }
         google()
     }
 }
 
 dependencies {
     //noinspection GradleDynamicVersion
+    implementation "breez_sdk:bindings-android:0.1.0"
     implementation "com.facebook.react:react-native:+"  // From node_modules
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.21"
     implementation("net.java.dev.jna:jna:5.6.0@aar")
-
 }


dangeross:
I'll checkout the PR locally and have a test

dangeross:
Same happens for me too, it seems like gradle ignores the Breez module repositories and it only works if I add the maven repository to the example project

cnixbtc:
Thanks for checking @dangeross! I'm not the biggest gradle expert so I haven't yet found out if this is a peculiarity with the way our example project is setup or if this is a general thing in all React Native projects.

To find out, I wanted to test integrating the plugin into a fresh React Native project but ran into [this weird issue](https://github.com/facebook/react-native/issues/38056) that make it impossible to run a fresh React Native project on my machine. So I guess I have to get that fixed first before I can test further.

In case I can't get that fixed soon, I'll reach out to you again @dangeross if that's ok with you. Maybe you can then try it out in a fresh project to see if it would work there.

dangeross:
No problem. I'll test a fresh project and will also test it in my own app

cnixbtc:
> No problem. I'll test a fresh project and will also test it in my own app

Thanks! Let me know how it goes. I'll try to fix my React Native setup in the meantime.

dangeross:
BTW the fresh project setup worked for me, I initialised it the same as in your issue.

RN: 0.72.0
node: 16.14.2
ruby: 2.7.4

dangeross:
Testing in my existing app fails with the same error above. 

I added the maven repository to the project's `build.gradle`, which seems to resolve but has compilation issues because the RN bridge in the repo is newer than the maven dependency (I know that would not happen when using npm + maven)

dangeross:
Test with the fresh app failed also with the same error. 

With this one I installed the SDK via npm, then changed the `build.gradle` to reflect the PR and removed the binaries. After adding the maven repo to the project's repositories, it resolved and compiled

dangeross:
Is there anything stopping us publishing the package to another maven repository?

cnixbtc:
Thanks so much for testing! 🙏🏼

Bummer... I think the problem might be that RN automatically generates an `allprojects` repo declaration in the project-level gradle file which then overwrites the ones from the subproject gradle files.

But we can't be the only ones with that problem so there must be some way of doing it (I hope). I'll try to have another look at this later this week.

cnixbtc:
> Is there anything stopping us publishing the package to another maven repository?

You mean one that is already declared by default? I'm not a maven expert but that would obviously be a solution as well!

roeierez:
@dangeross @cnixbtc I have nothing against publishing to another repository. Do you want to give it a try?

cnixbtc:
I've setup jetpack (see #285) and the resolving works but now I see compilation errors. That's strange since when we use our maven repo we don't see any. Hope I can get to the bottom of this soon. 🙄 🤞 😅

dangeross:
> I see compilation errors

@cnixbtc What are the compilation errors you're getting?

cnixbtc:
> @cnixbtc What are the compilation errors you're getting?

It seems like whatever is downloaded from jitpack doesn't include all the files we're uploading...

dangeross:
I don't know much about jitpack, does it download from the libs/sdk-bindings/bindings-android directory or from the repository release tag?

cnixbtc:
> I don't know much about jitpack, does it download from the libs/sdk-bindings/bindings-android directory or from the repository release tag?


Me neither, but the way it works is that you give it a script that is supposed to build your project to a local maven repo. Then, wehenver you tag your repo (or alternatively on every commit you want) you can trigger jitpack to run this script. They then take this local maven repo and publish it.

When I run the script locally on my machine it correctly builds the project. The source files included in the local build are:

```
.
├── META-INF
│   └── MANIFEST.MF
├── breez_sdk
│   └── breez_sdk.kt
└── technology
    └── breez
        └── BuildConfig.java
```

This works in the example app when I integrate it using a local maven repo declaration. 

But for some reason, when I inspect what jitpack is actually publishing, it's missing the `breez_sdk.kt` file:

```
.
├── META-INF
│   └── MANIFEST.MF
└── technology
    └── breez
        └── BuildConfig.java
```


 I suspect that this is might be related to the fact that we don't use the standard reverse domain syntax (`technology.breez.breez_sdk.kt`). Maybe there's some magic involved on jitpack's side that I don't understand. It's a bummer that you cannot just upload whatever you build locally the way we do it with our private maven repo. 

That's the status quo but I haven't had time to dig deeper. I guess if we don't find a solution for making it work nicely with our private maven repo or jitpack, then the final straw would be to try to get into maven central. But I first would like to do some research how other people handle this. We can't be the only ones with this issue.

cnixbtc:
I do have a suspicion of what it could be so I'll spend some more time with jitpack. Will report back.

cnixbtc:
Ok I've fixed the issue described above (see https://github.com/breez/breez-sdk/pull/285). Last remaining problem is the following:

```
 Duplicate class com.sun.jna.Version found in modules jetified-jna-5.8.0 (net.java.dev.jna:jna:5.8.0) and jetified-jna-5.8.0-runtime (net.java.dev.jna:jna:5.8.0)
```

Somehow the JNA dependency seems to get installed twice. Haven't looked into this but this seems fixable.

dangeross:
Cool. This seems like a very similar [problem](https://stackoverflow.com/questions/73068303/duplicate-class-a-a-found-in-modules-jetified-android) with jetified. If that doesn't fix it, maybe try to exclude the duplicate dependency by group/name: https://stackoverflow.com/a/21765656

cnixbtc:
Thanks for the pointers @dangeross!

cnixbtc:
What's weird to me is that when I replicate the Jitpack build on my machine (by [running this script](https://github.com/breez/breez-sdk/pull/285/files#diff-8cb949924730d38620cc3886598c0f7c18c2aa7bfa03d7bbc10f9c0f0a032545) locally which configures the BreezSDK bindings in my local maven repositoy), and then use it by checking out 0.1.0 and applying this patch:

``` diff
diff --git a/libs/sdk-react-native/android/build.gradle b/libs/sdk-react-native/android/build.gradle
index c3ee0f5..7964deb 100644
--- a/libs/sdk-react-native/android/build.gradle
+++ b/libs/sdk-react-native/android/build.gradle
@@ -62,6 +62,8 @@ allprojects {
 
 dependencies {
     //noinspection GradleDynamicVersion
+    //implementation "com.github.breez:breez-sdk:d46b9d7c44" // from jitpack
+    implementation "breez_sdk:bindings-android:0.1.0" // from local maven repo
     implementation "com.facebook.react:react-native:+"  // From node_modules
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.21"
     implementation("net.java.dev.jna:jna:5.6.0@aar")
diff --git a/libs/sdk-react-native/example/android/build.gradle b/libs/sdk-react-native/example/android/build.gradle
index 54abfc4..bcbf132 100644
--- a/libs/sdk-react-native/example/android/build.gradle
+++ b/libs/sdk-react-native/example/android/build.gradle
@@ -58,5 +58,6 @@ allprojects {
         }
         google()
         maven { url 'https://www.jitpack.io' }
+        mavenLocal()
     }
 }
```

and then running the example app everything works.

But when I switch to using the version from Jitpack (which as far as I can tell is the same thing), all of the sudden I get those duplicate class errors. Again I'm not a Gradle expert so I'm probably missing something obvious here. I haven't had any luck in trying to exclude the `com.sun.jna` group yet as in the second link you posted @dangeross. But maybe I just haven't found the right spot to do so. I tried doing it in the project level build gradle of the RN plugin. I couldn't try out the first link yet since testing changes in the bindings are pretty complicated and would require a fresh release I believe.

cnixbtc:
Will reopen this once we have a fix for Jitpack.

PR #279, opened 2023-06-25
Support python binding

roeierez:


dangeross (approved):
Looks good! 🎉

Can also add Python as a supported language at the top of the /libs/sdk-bindings/README and in the Build & Test section of the main /README

ok300:
> Can also add Python as a supported language at the top of the /libs/sdk-bindings/README and in the Build & Test section of the main /README

Done in 9046a960ca0134977a8f3d0a6fc2c9320ca3dd31

PR #281, opened 2023-06-26
Allow user request/response for backup

roeierez:
The initial version of the backup feature included the `start_backup` API method which started the backup process without waiting for completion and return the end result.
The user had to monitor the backup events (started, succeeded failed) in order to track the in progress operation.
The PR changes the following in order for the UX to be simpler:

1. change the `start_backup` to `backup` which now:
- Enque a backup request
- Wait for it to complete and return the result.
2. The `BackupWatcher` now monitors a request channel in addition to the sql hooks channel, so the user requests goes to that channel directly and not via the `sync_requests` table as before.  This allows the removal of `add_sync_request` method completely. 
3. I have noticed the use of temp directory fails in Android so the code was changed to use a temporary directory inside the sdk working dir.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -425,7 +425,7 @@ interface BlockingBreezServices {
    BackupStatus backup_status();
 
    [Throws=SDKError]
-   void start_backup();
+   void backup();
 
    [Throws=SDKError]
    sequence<Payment> list_payments(PaymentTypeFilter filter, i64? from_timestamp, i64? to_timestamp);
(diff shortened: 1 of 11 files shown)


ok300 (approved):

dangeross:
Is the intent to remove the backup events and return the status as the result? 

Side question, in the `backup_status` response, does `backed_up` field switch to false once the SDK deems it out of date?

roeierez:
> Is the intent to remove the backup events and return the status as the result?

No, events are useful as well so this doesn't change.



> Side question, in the `backup_status` response, does `backed_up` field switch to false once the SDK deems it out of date?

SDK trigger backup whenever needed on a best effort. In cases when the automatic backup fails (network error for example) an event will be fired (BackupFailed) and this flag will be switched to `false`. The intention is that the application will be aware of that and may choose to invoke the `backup` API explicitly.

dangeross:
> No, events are useful as well so this doesn't change.

Ah, I see. First reading I was expecting a `backup_status` response from the `backup` request. But a non-error response should be considered a successful backup.

dangeross (approved):

PR #282, opened 2023-06-26
Receive onchain: validate and select fee params

ok300:
- Improve conversion between `models::OpeningFeeParams` and `grpc::OpeningFeeParams`
  - with custom serde parsing into `DateTime` for easier time operations
- Select and validate fee params in `receive_onchain`
- Address `TODO` in `get_longest_valid_opening_fee_params`
  - filter fees that are valid for at least 48h, choose the cheapest

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1152,50 +1152,33 @@ impl Receiver for PaymentReceiver {
         let mut short_channel_id = parse_short_channel_id("1x0x0")?;
         let mut destination_invoice_amount_sats = amount_sats;
 
-        let cheapest_opening_fee_params =
-            lsp_info.choose_channel_opening_fees(req_data.opening_fee_params, true)?;
+        // Ensure opening fees params is set (either from the user, or from the LSP)
+        let cheapest_opening_fee_params = lsp_info
+            .choose_channel_opening_fees(req_data.opening_fee_params, true)?
+            .context("The provider is currently not opening channels")?;
+
         // check if we need to open channel
         let open_channel_needed = node_state.inbound_liquidity_msats < amount_msats;
         if open_channel_needed {
             info!("We need to open a channel");
 
             // we need to open channel so we are calculating the fees for the LSP
-            if let Some(ofp) = &cheapest_opening_fee_params {
-                let channel_fees_msat_calculated =
-                    amount_msats * ofp.proportional as u64 / 1_000_000 / 1_000_000;
-                let channel_fees_msat = max(channel_fees_msat_calculated, ofp.min_msat);
+            let ofp = &cheapest_opening_fee_params;
+            let channel_fees_msat_calculated =
+                amount_msats * ofp.proportional as u64 / 1_000_000 / 1_000_000;
+            let channel_fees_msat = max(channel_fees_msat_calculated, ofp.min_msat);
 
-                info!("zero-conf fee calculation option: lsp fee rate (proportional): {}:  (minimum {}), total fees for channel: {}",
+            info!("zero-conf fee calculation option: lsp fee rate (proportional): {}:  (minimum {}), total fees for channel: {}",
                 ofp.proportional, ofp.min_msat, channel_fees_msat);
 
-                if amount_msats < channel_fees_msat + 1000 {
-                    return Err(anyhow!(
-                        "requestPayment: Amount should be more than the minimum fees {} sats",
-                        ofp.min_msat / 1000
-                    ));
-                }
-                // remove the fees from the amount to get the small amount on the current node invoice.
-                destination_invoice_amount_sats = amount_sats - channel_fees_msat / 1000;
-            } else {
-                // TODO: This branch only exists because there may be existing swaps
-                // that use the old fee rate mechanism. Remove this after those
-                // swaps have expired.
-                let channel_fees_msat_calculated =
-                    amount_msats * lsp_info.channel_fee_permyriad as u64 / 10_000 / 1_000_000;
-                let channel_fees_msat = max(
-                    channel_fees_msat_calculated,
-                    lsp_info.channel_minimum_fee_msat as u64,
-                );
-
-                if amount_msats < channel_fees_msat + 1000 {
-                    return Err(anyhow!(
-                        "requestPayment: Amount should be more than the minimum fees {} sats",
-                        lsp_info.channel_minimum_fee_msat / 1000
-                    ));
-                }
-                // remove the fees from the amount to get the small amount on the current node invoice.
-                destination_invoice_amount_sats = amount_sats - channel_fees_msat / 1000;
+            if amount_msats < channel_fees_msat + 1000 {
+                return Err(anyhow!(
+                    "requestPayment: Amount should be more than the minimum fees {} sats",
+                    ofp.min_msat / 1000
+                ));
             }
+            // remove the fees from the amount to get the small amount on the current node invoice.
+            destination_invoice_amount_sats = amount_sats - channel_fees_msat / 1000;
         } else {
             // not opening a channel so we need to get the real channel id into the routing hints
             info!("Finding channel ID for routing hint");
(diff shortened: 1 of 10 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1306,27 +1305,20 @@ impl Receiver for PaymentReceiver {
     }
 }
 
-fn validate_opening_fee_params_menu(
-    opening_fee_params_menu: &Vec<models::OpeningFeeParams>,
-) -> Result<()> {
+fn validate_opening_fee_params_menu(opening_fee_params_menu: &Vec<OpeningFeeParams>) -> Result<()> {
```
roeierez, line 1308: NIT: rename to `validate_opening_fee_params_options` ?

libs/sdk-core/src/lsp.rs
```diff
@@ -74,7 +67,7 @@ impl LspAPI for BreezServer {
         let response = client.lsp_list(request).await?;
         let mut lsp_list: Vec<LspInformation> = Vec::new();
         for (key, value) in response.into_inner().lsps.into_iter() {
-            lsp_list.push(convert_to_lsp_info(key, value));
+            lsp_list.push(convert_to_lsp_info(key, value)?);
```
roeierez, line 70: Now that this may fail with error we should not let this affect other lsps. I believe we should log the error and keep trying to add the rest of the lsps the list.

libs/sdk-core/src/models.rs
```diff
@@ -661,14 +663,29 @@ impl From<OpeningFeeParams> for grpc::OpeningFeeParams {
...
 }
 
+impl TryFrom<grpc::OpeningFeeParams> for OpeningFeeParams {
+    type Error = anyhow::Error;
+
+    fn try_from(gofp: grpc::OpeningFeeParams) -> std::result::Result<Self, Self::Error> {
+        Ok(Self {
+            min_msat: gofp.min_msat,
+            proportional: gofp.proportional,
+            valid_until: serde_json::from_str(&gofp.valid_until).map_err(|e| anyhow!(e))?,
```
roeierez, line 681: This error will propagate all the way up and prevent from this lsp to be returned to the user.
Let's add a meaningful error here such as "lsp <id> `valid_until` field format is violates the specification"

roeierez: I am tending towards moving the fee params validation code to this point. I have noticed we are validating on `get_longest_fee_params` and get `cheapest_fee_params` methods and moving the validation to this point is earlier and holds for all use cases. We can omit LSP that doesn't pass the validation.
@ok300 @JssDWt what do you think?

roeierez (changes requested):
Looks good and simplified!
Dropped some comments.

libs/sdk-core/src/models.rs
JssDWt: Yeah, it's a message spec, so the place to validate whether the message has a valid shape is here.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1348,25 +1340,22 @@ fn get_cheapest_opening_fee_params(
...
+    let now = Utc::now();
+    let duration_48h = chrono::Duration::hours(48);
+    let mut valid_min_48h: Vec<OpeningFeeParams> = opening_fee_params_menu
+        .iter()
+        .filter(|f| f.valid_until - now > duration_48h)
+        .cloned()
+        .collect();
+
+    // Of those, return the cheapest
+    valid_min_48h.sort_by_key(|f| f.min_msat);
```
JssDWt, line 1357: Note that these have to be sorted already according to the spec. So if the menu is valid, you shouldn't have to sort here again. Also, there could be multiple params with the same `min_msat`, but a different `proportional`.

libs/sdk-core/src/breez_services.rs
```diff
@@ -483,23 +483,22 @@ impl BreezServices {
...
-        let longest_valid_opening_fee_params = get_longest_valid_opening_fee_params(
-            &mut self.lsp_info().await?.opening_fee_params_menu,
-        );
+        let channel_opening_fees = opening_fee_params.or(get_longest_valid_opening_fee_params(
+            &self.lsp_info().await?.opening_fee_params_menu,
+        )?);
         let swap_info = self
             .btc_receive_swapper
-            .create_swap_address(longest_valid_opening_fee_params)
+            .create_swap_address(channel_opening_fees)
```
JssDWt, line 501: Does it make sense to give the `create_swap_address` method a request object as well?

libs/sdk-core/src/breez_services.rs
JssDWt: It's called menu in the spec, probably best to stick with the spec naming here.

JssDWt (commented):
Looks good.
I'm not really seeing the full picture yet, will we do a PR of the dynamic_fees branch when we're done?

erdemyerebasmaz:
> Looks good. I'm not really seeing the full picture yet, will we do a PR of the dynamic_fees branch when we're done?

Yes, there will be a general review of the dynamic_fees branch once all subtasks are complete.

libs/sdk-core/src/breez_services.rs
ok300: I moved this to `OpeningFeeParamsMenu::validate()`.

libs/sdk-core/src/lsp.rs
ok300: I had to revert back to non-fallible conversion, so this is not a problem anymore.

libs/sdk-core/src/breez_services.rs
ok300: Since `create_swap_address` has only one arg (`channel_opening_fees`), IMO it makes little sense to wrap it in a separate request object.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1305,57 +1302,71 @@ impl Receiver for PaymentReceiver {
...
+                    false
+                }
+            })
+            .collect();
 
-    // Of those, return the cheapest
-    valid_min_48h.sort_by_key(|f| f.min_msat);
-    Ok(valid_min_48h.first().cloned())
+        // Of those, return the cheapest
+        valid_min_48h.sort_by_key(|f| f.min_msat);
```
erdemyerebasmaz, line 1367: I think this https://github.com/breez/breez-sdk/pull/282#discussion_r1244106104 was left out

> Note that these have to be sorted already according to the spec. So if the menu is valid, you shouldn't have to sort here again. Also, there could be multiple params with the same `min_msat`, but a different `proportional`.

The menu should already be sorted and lower min_msat doesn't necessarily mean cheaper fees as proportional fee may be much higher as explained.

libs/sdk-core/src/breez_services.rs
JssDWt: Yeah, I was just thinking this is a breaking change in the API structure. If we want to add another (optional) parameter, it will be a breaking change again.

libs/sdk-core/src/breez_services.rs
ok300: I addressed that and added a series of tests to cover all the sorting combos I could think of.

LMK if tests looks good, or if I missed anything.

ok300:
I had to change `valid_until` back to `String`, because having it as `DateTime` caused trouble with uniffi.

I did add however a utility method to convert to date when needed: `OpeningFeeParams::valid_until_date()`.

ok300:
Added by the last commit:

* Require a mandatory OpeningFeeParams arg for BTCReceiveSwap::create_swap_address.
* This can come either from the caller (as an arg to receive_onchain or can be looked-up from the LSP.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1323,9 +1323,16 @@ impl OpeningFeeParamsMenu {
...
-            (ofp[0].min_msat, ofp[0].proportional) <= (ofp[1].min_msat, ofp[1].proportional)
+            let larger_min_msat_fee = ofp[0].min_msat < ofp[1].min_msat;
+            let equal_min_msat_fee = ofp[0].min_msat == ofp[1].min_msat;
+
+            let larger_proportional = ofp[0].proportional < ofp[1].proportional;
+            let equal_proportional = ofp[0].proportional == ofp[1].proportional;
+
+            (larger_min_msat_fee && equal_proportional)
+                || (equal_min_msat_fee && larger_proportional)
+                || (larger_min_msat_fee && larger_proportional)
```
erdemyerebasmaz, line 1318: This is semantically the exact condition on spec but doesn't it boil down to each succeeding item must have either a larger min_msat or a larger proportional?

i.e.,
```
let is_ordered = opening_fee_params_menu.windows(2).all(|ofp| {
    ofp[0].min_msat < ofp[1].min_msat || ofp[0].proportional < ofp[1].proportional
});
```

PR #284, opened 2023-06-28
Add payment_by_hash API.

roeierez:


ok300 (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -454,6 +454,13 @@ impl BreezServices {
             .map_err(|err| anyhow!(err))
     }
 
+    /// Fetch a specific payment by its hash.
+    pub async fn payment_by_hash(&self, hash: String) -> Result<Option<Payment>> {
+        self.persister
+            .get_payment_by_hash(&hash)
```
JssDWt, line 460: Does this include LSP data? i.e. the 'bigger' invoice in case of an LSP registered payment? It might be confusing if you made a payment for `x` sats and this function returns info for an invoice of `x - fee` sats.

roeierez: It contains the same information as in the list_payments.
In the open channel case we show the lower amount as the amount paid and the difference in the fee field.
Note that it is a completed/pending payment, not invoice.

JssDWt: Oh yeah thanks I had an invoice in mind, nvm.

PR #286, opened 2023-06-29
Create swap address: add mandatory fee params arg

ok300:
Requre a mandatory `OpeningFeeParams` arg for `BTCReceiveSwap::create_swap_address`.

This can come either from the caller (as an arg to `receive_onchain` or can be looked-up from the LSP.


ok300:
Merging this in the other PR, which means it can be reviewed there.

PR #287, opened 2023-06-29
Enable rpath in profile settings

dangeross:
Set the rpath setting to true in the release profile to allow the runtime library search path to be passed to the linker. This enables the darwin libraries to be distributed in a Golang module and linked to at runtime using cgo.

### DIFF
--- a/libs/Cargo.toml
+++ b/libs/Cargo.toml
@@ -3,6 +3,7 @@ lto = true
 opt-level = "z"
 codegen-units = 1
 panic = "abort"
+rpath = true
 
 [workspace]
 members = [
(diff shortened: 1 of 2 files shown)


roeierez (approved):

Nackoo2000:
This is too technical for me to understand. Does this affect the UX in some way - how and what should I be QA testing?

dangeross:
It effects the binding libraries build for the different languages. The rpath changes how the library is referenced in the file, from an absolute path to relative path and gives the option to replace the path at run time. The change was specifically made for the Go programming language. I have sucessfully tested the swift and kotlin bindings built including this change.

PR #292, opened 2023-07-03
`list_lsps`: Skip LSPs that fail validation

ok300:
- Validate `LspInformation` when initializing
  - Validation consists of checking if the `opening_fee_params_menu` is valid
- `BreezServices::list_lsps()` only returns entries that pass validation

### DIFF
--- a/libs/sdk-core/src/lsp.rs
+++ b/libs/sdk-core/src/lsp.rs
@@ -23,13 +23,7 @@ pub struct LspInformation {
     pub fee_rate: f64,
     pub time_lock_delta: u32,
     pub min_htlc_msat: i64,
-    #[deprecated]
-    pub channel_fee_permyriad: i64,
     pub lsp_pubkey: Vec<u8>,
-    #[deprecated]
-    pub max_inactive_duration: i64,
-    #[deprecated]
-    pub channel_minimum_fee_msat: i64,
     pub opening_fee_params_menu: OpeningFeeParamsMenu,
 }
 
@@ -48,10 +42,7 @@ impl LspInformation {
             fee_rate: lsp_info.fee_rate,
             time_lock_delta: lsp_info.time_lock_delta,
             min_htlc_msat: lsp_info.min_htlc_msat,
-            channel_fee_permyriad: lsp_info.channel_fee_permyriad,
             lsp_pubkey: lsp_info.lsp_pubkey,
-            max_inactive_duration: lsp_info.max_inactive_duration,
-            channel_minimum_fee_msat: lsp_info.channel_minimum_fee_msat,
             opening_fee_params_menu: OpeningFeeParamsMenu::try_from(
                 lsp_info.opening_fee_params_menu,
             )?,
--- a/libs/sdk-core/src/lsp.rs
+++ b/libs/sdk-core/src/lsp.rs
@@ -53,11 +53,7 @@ impl LspInformation {
             max_inactive_duration: lsp_info.max_inactive_duration,
             channel_minimum_fee_msat: lsp_info.channel_minimum_fee_msat,
             opening_fee_params_menu: OpeningFeeParamsMenu::try_from(
-                lsp_info
-                    .opening_fee_params_menu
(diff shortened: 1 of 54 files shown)


erdemyerebasmaz (approved):
LGTM

libs/sdk-core/src/lsp.rs
```diff
@@ -56,7 +57,10 @@ impl LspInformation {
                 .into_iter()
                 .map(|ofp| ofp.into())
                 .collect::<Vec<OpeningFeeParams>>(),
-        }
+        };
+
+        OpeningFeeParamsMenu::try_from(info.opening_fee_params_menu.clone())?;
```
roeierez, line 62: It looks a bit weird that we first convert the opening_fee_params to our own structure and then validate. I is more natural flow to validate the raw input and then convert it.
If this is the only point we need to validate then maybe move the validate to this struct implementation?

JssDWt (approved):

libs/sdk-core/src/lsp.rs
ok300: Addressed in the latest commit:

https://github.com/breez/breez-sdk/pull/292/commits/8e3d706a1eea7a6a728d4514cea8620ca23086ce#diff-209db7ea9780093f9c880649e7f40a90a4615da7cee3111a299bbfa430b3aec2L55-L62

Part of it was also:
- changing `LspInformation.opening_fee_params_menu` type to `OpeningFeeParamsMenu`, so the menu is validated on instantiation
- moved `OpeningFeeParamsMenu` to `models.rs`
- updated uniffi bindings

libs/sdk-core/src/lsp.rs
```diff
@@ -51,33 +52,42 @@ impl LspInformation {
...
             max_inactive_duration: lsp_info.max_inactive_duration,
             channel_minimum_fee_msat: lsp_info.channel_minimum_fee_msat,
-            opening_fee_params_menu: lsp_info
-                .opening_fee_params_menu
-                .into_iter()
-                .map(|ofp| ofp.into())
-                .collect::<Vec<OpeningFeeParams>>(),
-        }
+            opening_fee_params_menu: OpeningFeeParamsMenu::try_from(
+                lsp_info
```
roeierez, line 56: This looks a lot better!
Perhaps also it is better to change the try_from to get the grpc values and move the mapping into our model inside that function?

ok300: Done

roeierez (approved):
LGTM

roeierez:
@ok300 I am guessing the buymoonpay test fails because it uses the mock LSP which returns the old structure. Now that the code validates the LSP list it probably returned empty list.

ok300:
@roeierez turns out it was a combination of
- using the default Staging LSP ID, but using an LSP mock with a test ID of `1`
- getting a `None` dyanmic fee rate in `receive_onchain` because the user fee was `None` and the mocked LSP fee was not at least 48h valid

Fixed in 073499a41303b34edbe8ecc83e4816a22b021a01

erdemyerebasmaz (commented):
Enum changes look good 5c0bb3c90611cce104c1b4ce5d331ed724ae2bc3

libs/sdk-core/src/swap.rs
```diff
@@ -408,7 +395,7 @@ impl BTCReceiveSwap {
                     amount_sats: swap_info.confirmed_sats,
                     description: String::from("Bitcoin Transfer"),
                     preimage: Some(swap_info.preimage),
-                    opening_fee_params: swap_info.channel_opening_fees,
+                    opening_fee_params: None,
```
JssDWt: Shouldn't this be passed in here?

ok300: I checked, this is based on a previous commit. The param is currently passed correctly.

ok300:
As discussed, merging this into `dynamic_fees`, so we can use that as the basis for the full dynamic fees review.

PR #294, opened 2023-07-03
Fee calculation and presentation

ademar111190:
Related: https://github.com/breez/c-breez/pull/621

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -83,7 +83,7 @@ dictionary UnspentTransactionOutput {
     u32 outnum;
     u64 amount_millisatoshi;
     string address;
-    boolean reserved;  
+    boolean reserved;
 };
 
 dictionary NodeState {
@@ -536,6 +536,16 @@ enum BuyBitcoinProvider {
     "Moonpay",
 };
 
+dictionary PrepareSweepRequest {
+    string to_address;
+    u64 sats_per_vbyte;
+};
+
+dictionary PrepareSweepResponse {
+    u64 sweep_tx_weight;
+    u64 sweep_tx_fee_sat;
+};
+
 dictionary SweepRequest {
     string to_address;
     u32 fee_rate_sats_per_vbyte;
@@ -580,7 +590,7 @@ dictionary RefundResponse {
     string refund_tx_id;
 };
 
-interface BlockingBreezServices {   
+interface BlockingBreezServices {
 
    [Throws=SdkError]
    void disconnect();
@@ -686,6 +696,9 @@ interface BlockingBreezServices {
 
    [Throws=SdkError]
    BuyBitcoinResponse buy_bitcoin(BuyBitcoinRequest req);
+
+   [Throws=SdkError]
+   PrepareSweepResponse prepare_sweep(PrepareSweepRequest req);
 };
 
 namespace breez_sdk {  
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -476,6 +476,12 @@ impl BreezServices {
         Ok(SweepResponse { txid })
     }
 
+    pub async fn prepare_sweep(&self, req: PrepareSweepRequest) -> Result<PrepareSweepResponse> {
+        self.start_node().await?;
+        let response = self.node_api.prepare_sweep(req).await?;
+        Ok(response)
+    }
+
     /// Fetch live rates of fiat currencies
     pub async fn fetch_fiat_rates(&self) -> Result<Vec<Rate>> {
         self.fiat_api.fetch_fiat_rates().await
@@ -1430,6 +1436,11 @@ impl BreezServicesBuilder {
             .unwrap_or_else(|| Arc::new(SqliteStorage::new(self.config.working_dir.clone())));
         persister.init()?;
 
+        // mempool space is used to monitor the chain
+        let chain_service = Arc::new(MempoolSpace::from_base_url(
+            self.config.mempoolspace_url.clone(),
+        ));
+
         let mut node_api = self.node_api.clone();
         let mut backup_transport = self.backup_transport.clone();
         if node_api.is_none() {
@@ -1498,11 +1509,6 @@ impl BreezServicesBuilder {
         let breez_server = Arc::new(BreezServer::new(
             self.config.breezserver.clone(),
             self.config.api_key.clone(),
-        ));
-
-        // mempool space is used to monitor the chain
-        let chain_service = Arc::new(MempoolSpace::from_base_url(
-            self.config.mempoolspace_url.clone(),
         ));
 
         let current_lsp_id = persister.get_lsp_id()?;
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -6,18 +6,21 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use anyhow::{anyhow, Result};
 use bitcoin::bech32::{u5, ToBase32};
+use bitcoin::blockdata::constants::WITNESS_SCALE_FACTOR;
+use bitcoin::hashes::Hash;
 use bitcoin::secp256k1::ecdsa::{RecoverableSignature, RecoveryId};
 use bitcoin::secp256k1::PublicKey;
 use bitcoin::secp256k1::Secp256k1;
 use bitcoin::util::bip32::{ChildNumber, ExtendedPrivKey};
+use bitcoin::{Address, OutPoint, Script, Sequence, Transaction, TxIn, TxOut, Txid, Witness};
 use ecies::utils::{aes_decrypt, aes_encrypt};
 use gl_client::node::ClnClient;
 use gl_client::pb::cln::listinvoices_invoices::ListinvoicesInvoicesStatus;
 use gl_client::pb::cln::listpays_pays::ListpaysPaysStatus;
 use gl_client::pb::cln::{
     self, CloseRequest, ListclosedchannelsClosedchannels, ListclosedchannelsRequest,
-    ListinvoicesInvoices, ListpaysPays, ListpeerchannelsRequest, SendcustommsgRequest,
-    StaticbackupRequest,
+    ListfundsRequest, ListfundsResponse, ListinvoicesInvoices, ListpaysPays,
+    ListpeerchannelsRequest, SendcustommsgRequest, StaticbackupRequest,
 };
 use gl_client::pb::cln::{AmountOrAny, InvoiceRequest};
 use gl_client::pb::{OffChainPayment, PayStatus};
@@ -39,7 +42,7 @@ use tonic::Streaming;
 use crate::invoice::parse_invoice;
 use crate::models::*;
 use crate::persist::db::SqliteStorage;
-use crate::{Channel, ChannelState, NodeConfig};
+use crate::{Channel, ChannelState, NodeConfig, PrepareSweepRequest, PrepareSweepResponse};
 use std::iter::Iterator;
 
 const MAX_PAYMENT_AMOUNT_MSAT: u64 = 4294967000;
@@ -60,7 +63,7 @@ impl Greenlight {
     /// If the node is not created, it will register it using the provided partner credentials
     /// or invite code
     /// If the node is already registered and an existing credentials were found, it will try to
-    /// connect to the node using these credentials.    
+    /// connect to the node using these credentials.
     pub async fn connect(
         config: Config,
         seed: Vec<u8>,
@@ -352,6 +355,45 @@ impl Greenlight {
             channels_balance,
         ))
     }
+
+    async fn list_funds(&self) -> Result<ListfundsResponse> {
+        let mut client = self.get_node_client().await?;
+        let funds: ListfundsResponse = client
+            .list_funds(ListfundsRequest::default())
+            .await?
+            .into_inner();
+        Ok(funds)
+    }
+
+    async fn on_chain_balance(&self, funds: ListfundsResponse) -> Result<u64> {
+        let on_chain_balance = funds.outputs.iter().fold(0, |a, b| {
+            if b.reserved {
+                return a;
+            }
+            a + b.amount_msat.clone().unwrap_or_default().msat
+        });
+        Ok(on_chain_balance)
+    }
+
+    // Collect utxos from onchain funds
+    async fn utxos(&self, funds: ListfundsResponse) -> Result<Vec<UnspentTransactionOutput>> {
+        let utxos: Vec<UnspentTransactionOutput> = funds
+            .outputs
+            .iter()
+            .map(|output| UnspentTransactionOutput {
+                txid: output.txid.clone(),
+                outnum: output.output,
+                amount_millisatoshi: output
+                    .amount_msat
+                    .as_ref()
+                    .map(|a| a.msat)
+                    .unwrap_or_default(),
+                address: output.address.clone().unwrap_or_default(),
+                reserved: output.reserved,
+            })
+            .collect();
+        Ok(utxos)
+    }
 }
 
 #[tonic::async_trait]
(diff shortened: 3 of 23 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -523,6 +519,32 @@ impl NodeAPI for Greenlight {
         Ok(client.withdraw(request).await?.into_inner())
     }
 
+    async fn prepare_withdraw(
+        &self,
+        to_address: String,
+        fee_rate_sats_per_vbyte: u32,
+    ) -> Result<TxprepareResponse> {
```
roeierez, line 526: interface looks good here and also in breez_services.rs.
The return value should be an sdk struct defined in models.rs taht contains the raw tx and the relative + absolute fee info. Something as follows:

```
pub struct PrepareWithdrawResponse {
  pub raw_tx_hex: String,
  pub sat_per_vbyte: u32,
  pub fee_sat: u64
}
```

The looks good IMO also as a return result from breez_services.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -446,6 +338,124 @@ impl NodeAPI for Greenlight {
...
+        };
+
+        Ok(client.withdraw(request).await?.into_inner())
+    }
+
+    async fn prepare_withdraw(
+        &self,
+        to_address: String,
+        fee_rate_sats_per_vbyte: u32,
+        withdraw_amount_msats: Option<u64>,
```
roeierez, line 372: Why do we need the amount? Shouldn't we take it from the on-chain balance?

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -446,6 +338,124 @@ impl NodeAPI for Greenlight {
         client.key_send(request).await?.into_inner().try_into()
     }
 
+    async fn start(&self) -> Result<()> {
+        self.get_client().await?;
+        Ok(())
+    }
+
+    async fn sweep(
```
roeierez, line 346: We should be consistent with naming. Currently we have `prepare_withdraw` and `sweep`.
Perhaps change it to:
`prepare_withdraw`, `withdraw`

libs/sdk-core/src/breez_services.rs
```diff
@@ -392,6 +394,21 @@ impl BreezServices {
         Ok(())
     }
 
+    pub async fn prepare_withdraw(
```
roeierez, line 397: In order to minimize backward incompatibility when changing BreezServices interface methods it is better to create a structure that consists of all the input parameters. In this case: `PrepareWithdrawRequest`

roeierez (changes requested):
Starting to shape up! It is hard to review though with all the formatting changes.

libs/sdk-core/src/greenlight/node_api.rs
ademar111190: On Breez Mobile, we can choose an amount to withdraw. So I assumed it was part of the prerequisites, wasn't it?

libs/sdk-core/src/greenlight/node_api.rs
ademar111190: Agree. I'll open a separate PR changing from sweep to withdraw.

libs/sdk-core/src/breez_services.rs
ademar111190: Interesting, I'll do it, thanks Roei 🙏

ademar111190:
> Starting to shape up! It is hard to review though with all the formatting changes.

Thanks, please don't hesitate to give feedback about the implementation.

Regarding the changes, I hope these https://github.com/breez/breez-sdk/pull/307 and https://github.com/breez/breez-sdk/pull/297 helps to reduce it from this PR

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -404,6 +404,24 @@ enum BuyBitcoinProvider {
     "Moonpay",
 };
 
+[Enum]
+interface PrepareWithdrawExtra {
+    UnexpectedFunds();
```
roeierez, line 409: I don't understand this extra enum. We only have one withdraw type. I don't think it is needed.

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -684,6 +694,107 @@ impl NodeAPI for Greenlight {
...
+            let hex_txid = hex::encode(b.funding_txid.clone());
+            if opened_channels.iter().any(|c| c.funding_txid == hex_txid) {
+                return a + b.our_amount_msat;
+            }
+            a
+        });
+        return Ok(off_chain_balance);
+    }
+
+    async fn peers(&self) -> Result<ListPeersResponse> {
```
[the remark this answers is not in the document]
ademar111190: I think it was necessary in the way I was doing, but as I was misunderstanding a few things, it ends up not being necessary. I'll take a look and remove the unnecessary changes.

libs/sdk-bindings/src/breez_sdk.udl
ademar111190: I'll remove it because we are not doing the "spontaneous transaction" with off chain balance. That enum was an abstraction to it, but I believe it will not be necessary after our slack chat.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -494,6 +465,48 @@ impl NodeAPI for Greenlight {
         Ok(client.withdraw(request).await?.into_inner())
     }
 
+    async fn prepare_withdraw(
+        &self,
+        prepare_withdraw_request: PrepareWithdrawRequest,
+    ) -> Result<PrepareWithdrawResponse> {
+        let funds = list_funds(self).await?;
+        let amount = self.on_chain_balance(Some(funds.clone())).await?;
```
roeierez, line 473: @ademar111190  It is better to calculate the amount out of the utxos array. It will minimize an option for a race between the method calls.

ademar111190: @roeierez sorry I haven't noticed this comment before.
I'm avoiding code duplication here. Is there a way to implement your recommendation without duplicating code?

libs/sdk-core/src/greenlight/node_api.rs
ok300: > Starting to shape up! It is hard to review though with all the formatting changes.

@ademar111190 FYI: you can solve this by calling `cargo fmt --all` in `sdk-core` before pushing new commits.

libs/sdk-core/src/breez_services.rs
```diff
@@ -453,6 +453,19 @@ impl BreezServices {
...
+        &self,
+        prepare_withdraw_request: PrepareWithdrawRequest,
```
ok300, line 458:
```suggestion
        req: PrepareWithdrawRequest,
```

Easier to just name the field `req`.

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -493,6 +463,52 @@ impl NodeAPI for Greenlight {
...
+                    txid: it.txid,
+                    outnum: it.outnum,
+                }
+            })
+            .collect();
+
+        let request = cln::UtxopsbtRequest {
+            satoshi: Some(cln::Amount { msat: 1000 }),
+            feerate: Some(cln::Feerate {
+                style: Some(cln::feerate::Style::Perkw(req.fee_rate_sats_per_vbyte)),
```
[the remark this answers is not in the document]
ademar111190: @roeierez I think it is not correct, what I get running this:
```
sdk> prepare_withdraw bc1qgw0nkwgd8sxxrwrfhfa05zgn799vc85pdn5wsa 9
Error: status: Unknown, message: "Error calling method UtxoPsbt: RpcError { code: Some(301), message: \"Could not afford 108500sat using UTXOs totalling 108500sat with weight 271 at feerate 9000\" }", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 25 Aug 2023 13:00:37 GMT", "content-length": "0"} }
```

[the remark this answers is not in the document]
ademar111190: Maybe It is the amount 🤔 
investigating it.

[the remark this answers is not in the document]
ademar111190: Ok, fixed:
```
prepare_withdraw bc1qgw0nkwgd8sxxrwrfhfa05zgn799vc85pdn5wsa 9
PrepareWithdraw ran successfully with response: PrepareWithdrawResponse { weight: 271, fee_sat: 2439 }
```

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -496,7 +496,22 @@ enum BuyBitcoinProvider {
     "Moonpay",
 };
 
-interface BlockingBreezServices {   
+dictionary SweepRequest {
```
roeierez, line 499: A big weird we have Sweep & Withdraw (different names for the same functionality).
If we already have Sweep maybe we can change to:
PrepareSweepRequest/PrepareSweepResponse ?

roeierez (changes requested):

libs/sdk-core/src/models.rs
```diff
@@ -95,6 +94,7 @@ pub trait NodeAPI: Send + Sync {
     /// Gets the private key at the path specified
     fn derive_bip32_key(&self, path: Vec<ChildNumber>) -> Result<ExtendedPrivKey>;
     fn legacy_derive_bip32_key(&self, path: Vec<ChildNumber>) -> Result<ExtendedPrivKey>;
+    async fn on_chain_balance(&self, funds: Option<ListFundsResponse>) -> Result<u64>;
```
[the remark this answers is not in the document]
ademar111190: removed

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -808,6 +833,62 @@ impl NodeAPI for Greenlight {
...
+            if b.reserved {
+                return a;
+            }
+            a + amount_to_msat(&b.amount.clone().unwrap_or_default())
+        });
+        return Ok(on_chain_balance);
+    }
+}
+
+async fn list_funds(node: &Greenlight) -> Result<ListFundsResponse> {
```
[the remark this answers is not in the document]
ademar111190: moved

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -808,6 +833,62 @@ impl NodeAPI for Greenlight {
...
+    let mut client = node.get_client().await?;
+    let funds = client
+        .list_funds(pb::ListFundsRequest::default())
+        .await?
+        .into_inner();
+    Ok(funds)
+}
+
+// Collect utxos from onchain funds
+async fn utxos(
```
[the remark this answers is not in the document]
ademar111190: moved

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -808,6 +833,62 @@ impl NodeAPI for Greenlight {
...
+    node: &Greenlight,
+    funds: Option<ListFundsResponse>,
```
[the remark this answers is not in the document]
ademar111190: To reuse the `ListFundsResponse` we have already instead of calling `client.list_funds` multiple times.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -808,6 +833,62 @@ impl NodeAPI for Greenlight {
...
+) -> Result<Vec<UnspentTransactionOutput>> {
+    let funds = match funds {
+        Some(f) => f,
+        None => list_funds(node).await?,
```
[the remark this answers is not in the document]
ademar111190: ok, making it mandatory.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -560,28 +534,79 @@ impl NodeAPI for Greenlight {
...
+                        vout: 0,
+                    },
+                    script_sig: Script::new(),
+                    sequence: Sequence(0),
+                    witness: Witness::default(),
+                }
+            })
+            .collect();
+
+        // remove dust
```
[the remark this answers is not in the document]
ademar111190: Yes, it is round to satoshis; I'm changing the comment to make it clear

libs/sdk-bindings/src/breez_sdk.udl
ademar111190: true; I haven't thought about that. changing names here

ademar111190: Isn't it better to change the sweep to withdraw? So we keep the naming greenlight is using? @roeierez

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -496,7 +496,22 @@ enum BuyBitcoinProvider {
...
+    string to_address;
+    u64 sat_per_kw;
```
roeierez, line 501: User should use `sat_per_vbyte` not `sat_per_kw`

libs/sdk-bindings/src/breez_sdk.udl
roeierez: Don't have any preference. Perhaps if we already have sweep we can align wtih that. (PrepareSweep)

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -496,7 +496,22 @@ enum BuyBitcoinProvider {
...
+    u64 sats_per_vbyte;
+};
+
+dictionary PrepareSweepRequest {
+    string to_address;
+    u64 sats_per_vbyte;
+};
+
+dictionary PrepareSweepResponse {
+    u64 weight;
```
roeierez, line 510: NIT: sweep_tx_weight, sweep_tx_fee_sat

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -560,28 +579,76 @@ impl NodeAPI for Greenlight {
...
+            lock_time: bitcoin::PackedLockTime(0),
+            input: txins.clone(),
+            output: tx_out,
+        };
+
+        let witness_input_size: u64 = 110;
+        let tx_weight = tx.strippedsize() as u64 * WITNESS_SCALE_FACTOR as u64
+            + witness_input_size * txins.len() as u64;
+        let fee: u64 = tx_weight * req.sats_per_vbyte / WITNESS_SCALE_FACTOR as u64;
+        if fee >= amount {
```
roeierez, line 641: This is not enough to validate the transaction. I see there is a `verify` function that can be used.

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
ademar111190: Can you point out where is it @roeierez ?
I'm mimicking swap and it is not calling verify https://github.com/breez/breez-sdk/blob/4fdeb41997df2953a04efb48520447e4e3f9cd46/libs/sdk-core/src/swap.rs#L606

roeierez:
@ademar111190 this looks good and almost ready IMO.
I am missing the React Native generated files.

dangeross:
Here's some info to get you started with the RN codegen:

https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/DEVELOPING.md

ademar111190:
thanks for the link @dangeross 
not sure why but I had a lot of problems installing things, In the end I managed to do it by disabling ipv6 on my machine

ademar111190:
@roeierez gen files added; let me know if it is missing something

roeierez (approved):
LGTM!
Pease make sure the RN generated code is up to date with the latest generator before merging.

ademar111190:
@roeierez yeap; I did use it on my last rebase.
@dangeross can you double check if there is something wrong? What I can see that is correct but my knowledge in rn is almost 0

dangeross (approved):
Looks great!

PR #296, opened 2023-07-05
Startup time optimization (phase 1)

roeierez:
This PR addresses the startup time of the sdk in all three cases (init, register, recover).
A small changes to the rust interface was made:
instead of: 
`BreezServices::start(rt(), &sdk).await?`
we now use:
`sdk.start()await?`

The runtime is initiated inside the sdk which is cleaner and also allows cleaner shutdown on disconnecting.
The start method now uses tokio tasks to spin all king of polling in the background and make sure no connections/handshakes are blocking the flow anymore.

The startup of existing node was more than 5 seconds before and now it is less then 10 milliseconds.
registering new node and recover node results in 5 seconds but I believe we can make it 2-3 seconds in the next phase.
Sorry for the long diff.

### DIFF
--- a/libs/sdk-bindings/src/uniffi_binding.rs
+++ b/libs/sdk-bindings/src/uniffi_binding.rs
@@ -138,7 +138,7 @@ pub struct BlockingBreezServices {
 
 impl BlockingBreezServices {
     pub fn start(&self) -> Result<()> {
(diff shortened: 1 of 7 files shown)


ok300 (approved):
LGTM

PR #297, opened 2023-07-05
Fix different impl member order from the trait warning

ademar111190:
It seems cargo warns when your trait and your impl have a different order in its methods. I'm ordering it to solve the warning. Also running cargo fmt.


roeierez (approved):
LGTM

PR #298, opened 2023-07-05
Initial version of detailed SdkError

ok300:
This PR allows the SDK to return typed errors.


libs/sdk-core/src/errors.rs
```diff
@@ -0,0 +1,74 @@
+use reqwest::StatusCode;
+use std::time::Duration;
+
+pub type SdkResult<T, E = ExternalSdkError> = anyhow::Result<T, E>;
+
+// TODO Rename to SdkError once existing SDKError struct is replaced
+/// Public error kinds returned by the SDK
+#[derive(Debug)]
+pub enum ExternalSdkError {
+    LnInvoiceError { err: String },
```
roeierez, line 10: For generic errors I would try and define them without additional err data.
This way they will be specific and the user will know exactly what to do. Could be it will cause us to implement more specific errors.

libs/sdk-core/src/errors.rs
```diff
@@ -0,0 +1,74 @@
...
+            _ => Self::Generic,
+        }
+    }
+}
+
+pub(crate) type InternalSdkResult<T, E = InternalSdkError> = anyhow::Result<T, E>;
+
+/// Error kinds thrown internally by SDK modules
+#[derive(Debug)]
+pub(crate) enum InternalSdkError {
```
roeierez, line 30: I am not sure we need internal sdk error structure. Every module can define its own errors if needed but I don't think it should be in a central place.

roeierez (changes requested):

libs/sdk-core/src/errors.rs
ok300: Done, I removed any internal error data for now. Instead, I tried to reflect the specific meaning of the error in the name.

libs/sdk-core/src/errors.rs
ok300: I got rid of the `InternalSdkError`.

ok300:
I updated the PR with a revised version, incorporating all the feedback.

- I started by adding a custom error to `connect()`
- When I used `Error` as the custom error name, the binding generation got into trouble (it confused it with `anyhow::Error`). So for now I used a unique name, `NewSdkError`.
- Updated internal methods used in `connect()`, such that they return the new error type. Those internal methods were used in other places, so I had to propagate this to several other methods.

The naming can surely be improved. One idea is to use a schema like `SdkModuleError`, e.g. `SdkPersistanceFailedToLoadNodeState` or `SdkInitFailedToConnectToGreenlight`.

libs/sdk-core/src/breez_services.rs
```diff
@@ -352,10 +357,10 @@ impl BreezServices {
         filter: PaymentTypeFilter,
         from_timestamp: Option<i64>,
         to_timestamp: Option<i64>,
-    ) -> Result<Vec<Payment>> {
+    ) -> SdkResult<Vec<Payment>> {
         self.persister
             .list_payments(filter, from_timestamp, to_timestamp)
-            .map_err(|err| anyhow!(err))
+            .map_err(|_| NewSdkError::SdkFailedToListPayments)
```
roeierez, line 363: Perhaps this pattern of mapping error from the storage layer to a specific error can be improved.
IMO we don't need specific error for every persistence failure (e.g FailedToListPayments, FailedToGetNodeState, etc...).
We can do the following:
1. Define one error for persistency: `PersistenceFailure`
2. Return this error from the storage layer directly which eliminate the need to map here at the BreezServices layer.

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::Error;
+use std::fmt;
+
+pub type SdkResult<T, E = NewSdkError> = Result<T, E>;
+
+/// Type of error returned by the SDK
+#[derive(Debug)]
+pub enum NewSdkError {
```
roeierez, line 8: We can use just `SdkError` for naming this enum.

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+    SdkFailedToRequestBackup,
```
roeierez, line 9: A few comments regarding all error variants:
1. No need to prefix with sdk.
2. better to use the verb at the end, e.g: BackupRequestFailed
3. Worth adding a comment on each error that describes it.
4. Missing error message, worth considering using thisError as uniffi suggests: https://mozilla.github.io/uniffi-rs/udl/errors.html

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    SdkFailedToStartBackupWatcher,
```
roeierez, line 11: In general I think we shouldn't go into too many specific errors. A structured error should let the user respond specifically and if there is not specific action that can be done then it means it is too specific.
For example I don't think we need all of `SdkNoStateSynchronizerFound`, `SdkFailedToStartBackupWatcher` `SdkInitFailedNoNodeApiOrSeedFound` as the user has not control over the flow that caused these errors. We can just use ConnectFailed and log the specific reason.

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+    SdkFailedToListPayments,
+
+    SdkFailedToSync,
+
+    // NodeState
+    SdkCannotRetrievePersistedNodeState,
+    SdkNoNodeStateFound,
+
+    // LSPs
+    SdkFailedToRetrieveLsps,
```
roeierez, line 38: All these three errors can be unified to PersistenceFailed

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,64 @@
...
+    SdkNoStateSynchronizerFound,
+
+    SdkInitFailedNoNodeApiOrSeedFound,
+
+    SdkCannotRetrievePersistedLastSyncVersion,
+
+    SdkBip32ErrorIndexOutOfBounds,
+    SdkBip32ErrorFailedToDeriveKey,
+
+    SdkFailedToInsertOrUpdatePayments,
```
roeierez, line 25: The next three errors also related to persistency: PersistenseFailed

roeierez (changes requested):

libs/sdk-core/src/error.rs
ok300: It compiles, but won't this cause issues with uniffi?

`breez_sdk.udl` would contain:

```udl
[Error]
enum SDKError {
  "Error",
};

[Error]
enum SdkError {
  ...
}
```

roeierez: We should use the new sdk error and import it. We should remove the SDKError from the uniffi_binding.rs

JssDWt:
I want to make a suggestion. Let's look at the Stripe SDK for how they handle errors. I think they have a clean structure.
`SdkError` would get a `type` and a `code`. And a `message`. The type would be something like `PersistenceError` and the code `db_not_found` for example. Or type `PaymentError` and code `no_route`. Or a type `NetworkError` with code `unauthorized`.

In some languages the error object is based on inheritance, like the [Java example](https://stripe.com/docs/error-handling?lang=java). In some languages you have to switch on the `type`, like the [node.js example](https://stripe.com/docs/error-handling?lang=node).

We can then make detailed documentation based on the error type/code, which allows developers working with the SDK to know what to expect. We'll have to document the possible errors and codes for each function call.

ok300:
Thanks @JssDWt . The goal now is to make a simple version, where errors have their own enum value and a more verbose text description. The error code is equivalent to the enum value.

Once that's in, we can see how it behaves and what's missing.

JssDWt:
> Thanks @JssDWt . The goal now is to make a simple version, where errors have their own enum value and a more verbose text description. The error code is equivalent to the enum value.

I was thinking about the goal to make structured errors that are easy to use by developers. Stripe SDK is surely a great reference for that. I think we can take their learnings and apply them here without having to go through an iterative effort.

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/breez_services.rs
```diff
@@ -732,18 +738,22 @@ impl BreezServices {
         });
     }
 
-    async fn start_backup_watcher(self: &Arc<BreezServices>) -> Result<()> {
+    async fn start_backup_watcher(self: &Arc<BreezServices>) -> SdkResult<()> {
         self.backup_watcher
             .start(self.shutdown_receiver.clone())
-            .await?;
+            .await
+            .map_err(|_| SdkError::ConnectFailed)?;
```
roeierez, line 745: In all startup flows I would use InitFailed.

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,30 @@
+use thiserror::Error;
+
+pub type SdkResult<T, E = SdkError> = Result<T, E>;
+
+/// Type of error returned by the SDK
+#[derive(Debug, Error)]
+pub enum SdkError {
+    #[error("Failed to connect to required services")]
+    ConnectFailed,
```
roeierez, line 9: Not sure we need ConnectFailed (sounds too generic). It seems that ConnectFailed is used in start up flow, so we can use InitFailed instead.
In other cases like in list_lsps we can use a more specific error (see my comment bellow).

libs/sdk-core/src/persist/settings.rs
```diff
@@ -50,13 +50,11 @@ impl SqliteStorage {
     }
 
     pub fn set_lsp_id(&self, lsp_id: String) -> Result<()> {
-        self.update_setting("lsp".to_string(), lsp_id)?;
-        Ok(())
+        self.update_setting("lsp".to_string(), lsp_id)
```
roeierez, line 53: if we change `update_settings` to return PersistenceFailure then it would propagate all the way up.

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Generic error, that doesn't fit any of the other types
+    #[error("Breez SDK error: {err}")]
+    Generic { err: String },
+
+    #[error("Failed to initialize the SDK")]
+    InitFailed,
```
roeierez, line 16: I agree with your suggestion to have String payload for every error so the user can get a proper context so let's add it to all other errors.

roeierez (changes requested):

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,30 @@
...
+    PersistenceFailure,
+
+    #[error("Failed to receive payment")]
+    ReceivePaymentFailure,
+}
+
+// TODO This won't be necessary when all service methods return SdkResult
+impl From<anyhow::Error> for SdkError {
+    fn from(_value: anyhow::Error) -> Self {
+        Self::InitFailed
```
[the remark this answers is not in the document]
ok300: Correct, done.

libs/sdk-core/src/breez_services.rs
```diff
@@ -144,10 +145,10 @@ impl BreezServices {
     ///
     /// It should be called only once when the app is started, regardless whether the app is sent to
     /// background and back.
-    async fn start(self: &Arc<BreezServices>) -> Result<()> {
+    async fn start(self: &Arc<BreezServices>) -> SdkResult<()> {
         let mut started = self.started.lock().await;
         if *started {
-            return Err(anyhow::Error::msg("BreezServices already started"));
+            return Err(SdkError::InitFailed);
```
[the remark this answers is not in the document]
ok300: Done

libs/sdk-core/src/error.rs
```diff
@@ -0,0 +1,30 @@
...
+    Generic { err: String },
+
+    #[error("Failed to initialize the SDK")]
+    InitFailed,
+
+    #[error("Failed to use the local DB for persistence")]
+    PersistenceFailure,
+
+    #[error("Failed to receive payment")]
+    ReceivePaymentFailure,
```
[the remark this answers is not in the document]
ok300: Done

libs/sdk-core/src/breez_services.rs
```diff
@@ -313,15 +315,18 @@ impl BreezServices {
         &self,
         amount_sats: u64,
         description: String,
-    ) -> Result<LNInvoice> {
+    ) -> SdkResult<LNInvoice> {
         self.payment_receiver
             .receive_payment(amount_sats, description, None)
             .await
+            .map_err(|_| SdkError::ReceivePaymentFailure)
```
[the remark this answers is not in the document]
ok300: Done

libs/sdk-core/src/breez_services.rs
```diff
@@ -313,15 +315,18 @@ impl BreezServices {
...
     }
 
     /// Retrieve the node state from the persistent storage
-    pub fn node_info(&self) -> Result<Option<NodeState>> {
-        self.persister.get_node_state()
+    pub fn node_info(&self) -> SdkResult<Option<NodeState>> {
+        self.persister
```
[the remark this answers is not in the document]
ok300: Done

libs/sdk-core/src/breez_services.rs
ok300: Done

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/persist/settings.rs
ok300: Done

libs/sdk-core/src/error.rs
ok300: Done

libs/sdk-core/src/breez_services.rs
```diff
@@ -386,10 +391,11 @@ impl BreezServices {
     }
 
     /// List available LSPs that can be selected by the user
-    pub async fn list_lsps(&self) -> Result<Vec<LspInformation>> {
+    pub async fn list_lsps(&self) -> SdkResult<Vec<LspInformation>> {
         self.lsp_api
-            .list_lsps(self.node_info()?.ok_or_else(|| anyhow!("err"))?.id)
+            .list_lsps(self.node_info()?.ok_or(SdkError::PersistenceFailure)?.id)
             .await
+            .map_err(|_| SdkError::ConnectFailed)
```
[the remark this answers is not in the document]
ok300: Done. There are other LSP-related errors, like getting the ID, which can also be caused by connection issues to the LSP.

So I added a more generic `LspConnectFailed`.

roeierez (changes requested):
One small comment, other than that LGTM.

libs/sdk-core/src/breez_services.rs
```diff
@@ -315,17 +319,27 @@ impl BreezServices {
...
-    ) -> Result<LNInvoice> {
+    ) -> SdkResult<LNInvoice> {
         self.payment_receiver
             .receive_payment(amount_sats, description, None)
             .await
     }
 
     /// Retrieve the node state from the persistent storage
-    pub fn node_info(&self) -> Result<Option<NodeState>> {
+    pub fn node_info_persisted(&self) -> SdkResult<Option<NodeState>> {
```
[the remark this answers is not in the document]
ok300: If we keep only one `node_info(&self) -> Result<Option<NodeState>>`, we have to explicitly check for `None` and throw an `SdkError`:

https://github.com/breez/breez-sdk/pull/298/commits/4600a48d45e11277f1100875dfa324c2efbbbaaa#diff-6f499d3ded5e493c1bd95b827a0a93aeb3d2bfa3d04a6f26be62d87aff026e43L394-L405

Found this pattern in at least 2 places.

I see two options:
- keep the existing  `node_info(&self) -> Result<Option<NodeState>>` but redundantly check for `None` when `node_info()` is expected to return, and throw `SdkError` accordingly
- have a second `node_info()` variant that internally checks for `None` and throws `SdkError` if empty

I can go either way, the 2nd option seemed better at the time.

Which option do you prefer?

libs/sdk-core/src/breez_services.rs
```diff
@@ -315,17 +319,27 @@ impl BreezServices {
...
         self.persister.get_node_state()
     }
 
+    /// Retrieve the node state from the persistent storage.
+    ///
+    /// Fail if it could not be retrieved or if `None` was found.
+    pub fn node_info(&self) -> SdkResult<NodeState> {
```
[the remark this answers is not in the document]
ok300: `None` is still an acceptable result, the place where it checks for the initial sync uses the other function which also returns `None`, namely `node_info_persisted()`.

All other callers did the extra `None` check and expected a value, so technically the behavior didn't change.

See below for more details, I'm open to changing it if needed.

libs/sdk-core/src/breez_services.rs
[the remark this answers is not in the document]
roeierez: What I meant is that  it doesn't make sense to have both `node_info` & `node_info_persisted` as the public API on BreezServices (I saw it also at the udl file). It is very confusing from the user point of view.
IMO we don't need at all the `node_info_persisted` and all the callers that use it right now can instead use the following: `self.persister.get_node_state()`

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -9,8 +9,12 @@ dictionary RouteHintHop {
  };
 
 [Error]
-enum SDKError {
-  "Error",
+enum SdkError {
```
roeierez, line 12: @ok300  Is the renaming from SDKError to SdkError affects the React Native implementation?

libs/sdk-core/src/breez_services.rs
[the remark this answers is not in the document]
ok300: Done

libs/sdk-bindings/src/breez_sdk.udl
ok300: Unliley, because I haven't found any references to `SDKError` (case-sensitive) anywhere.

roeierez (approved):
LGTM

libs/sdk-bindings/src/breez_sdk.udl
ok300: Searched generated RN files, no mention of `SDKError`.

dangeross:
Should check if catches are effected e.g.:
`BreezSDKModule.kt`
```
            try {
                breezServices = connect(configData, asUByteList(seed), BreezSDKListener(emitter))
                promise.resolve(readableMapOf("status" to "ok"))
            } catch (e: SdkException) {
                e.printStackTrace()
                promise.reject(TAG, e.message ?: "Error calling connect", e)
            }
``` 
`RNBreezSDK.swift`
```
            do {
                self.breezServices = try BreezSDK.connect(config: config, seed: seed, listener: BreezSDKListener(emitter: self))
                
                resolve(["status": "ok"])
            } catch SdkError.Error(let message) {
                reject(RNBreezSDK.TAG, message, nil)
            } catch let err {
                reject(RNBreezSDK.TAG, "Error calling connect", err)
            }
```

PR #304, opened 2023-07-09
Reuse greenlight connection

roeierez:
This replaces https://github.com/breez/breez-sdk/pull/273 with a lock free approach.
We now take advantage of the connect method of Greenlight to initiate both connections and pass them to the NodeAPI instance.
It turns out that tonic connections are only cloned properly under the same runtime they were created (had to learn that the hard way...)
Therefore, I had to change the code not to create our own runtime for the background tasks but use the current runtime. As a library this seems the right approach as we can't control under which runtime the sdk is used.
The relevant code for the connection reuse is here: https://github.com/breez/breez-sdk/compare/connect-api...gl-connection-reuse#diff-7466a53546c6cded460ef5502a9168b13cc930d94fdf7d0ea8166bb10c75fe7eR151
This plus the activated Greenlight proxy (constant node URI) results in a significant performance gain.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -38,8 +38,7 @@ use std::cmp::max;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Instant, SystemTime, UNIX_EPOCH};
-use tokio::runtime::Runtime;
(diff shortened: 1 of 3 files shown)


ok300 (approved):
LGTM

PR #307, opened 2023-07-10
Run cargo fmt

ademar111190:
Just run cargo fmt on some file to reduce the changes here https://github.com/breez/breez-sdk/pull/294

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -1,16 +1,16 @@
-dictionary RouteHintHop {
-    string src_node_id;
-    u64 short_channel_id;
-    u32 fees_base_msat;
-    u32 fees_proportional_millionths;
-    u64 cltv_expiry_delta;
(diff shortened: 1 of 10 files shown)


libs/sdk-bindings/src/breez_sdk.udl
ademar111190, line 1: I did format this file as well.

roeierez:
@ademar111190 Looks good but I wonder why our CI doesn't fail: https://github.com/breez/breez-sdk/blob/main/.github/workflows/main.yml#L22

ademar111190:
> @ademar111190 Looks good but I wonder why our CI doesn't fail: https://github.com/breez/breez-sdk/blob/main/.github/workflows/main.yml#L22

I don't know either, maybe a different version?

My version:

```
 $ which cargo
 /Users/ademar/.cargo/bin/cargo
 $ cargo --version 
 cargo 1.69.0 (6e9a83356 2023-04-12)
 ```

roeierez (approved):

PR #309, opened 2023-07-11
Lazy connection to greenlight to improve startup time

roeierez:
This PR goes back to initiate greenlight connection lazily.
Before this change an attempt to go lock free resulted in initiating the connection as part of the connect method which increased this method time and caused bad user experience: https://github.com/breez/c-breez/pull/596#issuecomment-1629726361

Re-adding the mutexes which solves this problem. We can get rid of the mutexes again when greenlight exposes lazy connection API but this doesn't seem to be urgent IMO.


ok300 (approved):
Looks good

PR #310, opened 2023-07-12
Adapt fee settings

roeierez:
This PR make the default fee settings a bit more relaxed and match breez mobile.
It uses 1% as maximum with exempt fee of 20 sats.
These new settings should allow payments to Alby pass through.

fixes https://github.com/breez/c-breez/issues/599


erdemyerebasmaz (approved):
LGTM

PR #313, opened 2023-07-13
Add dynamic fees feature

ok300:
This covers all "dynamic fees" related changes in one PR.

#### Changelist:
- Validate opening fee params menu & select cheapest fees for bolt11 invoices
  - Checks if fees in the menu are
    - in ascending order,
    - not expired.
  - `OpeningFeeParamsMenu` validation accepts the case when it has no entries. However, when we need channel opening fee params, and the user provided none, the fee lookup will result in an error.
  - Any LSP's menu that fail validation are removed from LSP list
 -  Return `ReceivePaymentResponse` for `receive_payment`
   - `ReceivePaymentResponse` has invoice & fee information(fee params used & calculated fees)
-  Convert `receive_payment` parameters into `ReceivePaymentRequest` struct
   - `ReceivePaymentRequest` also has `opening_fee_params` as an optional parameter.
     - Set `openingFeeParams` as optional named parameter for `receiveOnchain` & `buyBitcoin` methods on `BreezBridge`
- Use cheapest valid opening fee params by default when creating an invoice
- Use longest valid opening fee params by default when creating a swap address
  - Distinction is made by `DynamicFeeType::Longest` & `DynamicFeeType::Cheapest` `fee_type` enums passed to `choose_channel_opening_fees`


libs/sdk-core/src/models.rs
```diff
@@ -641,6 +645,180 @@ pub struct ClosedChannelPaymentDetails {
...
+                }
+            })
+            .cloned()
+            .collect();
+
+        // Of those, return the first, which is the cheapest
+        // (sorting order of fee params list was checked when the menu was initialized)
+        valid_min_48h.first().cloned().ok_or(anyhow!(
+            "The LSP doesn't support opening new channels: Dynamic fees menu contains no values"
+        ))
```
JssDWt, line 818: There's a slight difference between the menu being empty (LSP doesn't support opening channels), and the menu being empty after filtering for 48h duration (LSP doesn't support params with a duration > 48h). Maybe we can distinguish in error messages here.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1230,33 +1240,39 @@ impl Receiver for PaymentReceiver {
...
-            if amount_msats < channel_fees_msat + 1000 {
-                return Err(anyhow!(
-                    "requestPayment: Amount should be more than the minimum fees {} sats",
-                    lsp_info.channel_minimum_fee_msat / 1000
-                ));
+            // we need to open channel so we are calculating the fees for the LSP (coming either from the user, or from the LSP)
+            let ofp = lsp_info.choose_channel_opening_fees(
+                req_data.opening_fee_params,
+                DynamicFeeType::Cheapest,
+            )?;
```
JssDWt, line 1261: I think this errors in the backward compatible case.
`req_data.opening_fee_params` will be `None` then, and we actually want to send `None` to the LSP then.
So when `None` is passed as an argument, when it comes from redeem_swap, this should not error.

JssDWt (commented):
Some small things, and an annoying thing for the backward compatible case.

libs/sdk-core/src/swap.rs
```diff
@@ -128,7 +129,10 @@ impl BTCReceiveSwap {
     /// Create a [SwapInfo] that represents the details of an on-going swap.
     ///
     /// See [SwapInfo] for details.
-    pub(crate) async fn create_swap_address(&self) -> Result<SwapInfo> {
+    pub(crate) async fn create_swap_address(
+        &self,
+        channel_opening_fees: OpeningFeeParams,
+    ) -> Result<SwapInfo> {
```
roeierez, line 135: There is a problematic case when the user created a swap address with some fees and now the user invoke this API to create a swap address and get the same reused one (see bellow  `self.list_unused()` )
In this case the user gets an unused one from the db but it is attached to the old fee while now there are new fees that the user has chosen (and saw).

We may need to update the fees here for this address.

roeierez (changes requested):

libs/sdk-bindings/src/uniffi_binding.rs
```diff
@@ -259,8 +255,11 @@ impl BlockingBreezServices {
     }
 
     /// Onchain receive swap API
-    pub fn receive_onchain(&self) -> Result<SwapInfo, SDKError> {
-        rt().block_on(self.breez_services.receive_onchain())
+    pub fn receive_onchain(
+        &self,
+        opening_fee_params: Option<OpeningFeeParams>,
```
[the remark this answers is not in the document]
JssDWt: I agree.

libs/sdk-core/src/swap.rs
ok300: Done: d4fb0e11c76fac630f15abbbe400a4732d138ed9

tools/sdk-cli/src/command_handlers.rs
```diff
@@ -99,13 +96,20 @@ pub(crate) async fn handle_command(
...
-            let invoice = sdk()?.receive_payment(amount, description).await?;
+            let invoice = sdk()?
+                .receive_payment(ReceivePaymentRequest {
+                    amount_sats,
+                    description,
+                    preimage: None,
+                    opening_fee_params: None,
+                })
+                .await?;
             let mut result = serde_json::to_string(&invoice)?;
```
roeierez, line 110: Perhaps we also want to print the opening channel absolute fees in case they are exists?

libs/sdk-core/src/models.rs
```diff
@@ -637,6 +645,191 @@ pub struct ClosedChannelPaymentDetails {
...
+    }
+
+    pub(crate) fn valid_until_date(&self) -> Result<DateTime<Utc>> {
+        Utc.datetime_from_str(&self.valid_until, OPENING_FEE_PARAMS_DATETIME_FORMAT)
+            .map_err(|e| anyhow!(e))
+    }
+
+    pub(crate) fn get_channel_fees_msat_for(&self, amount_msats: u64) -> u64 {
+        max(
+            amount_msats * self.proportional as u64 / 1_000_000 / 1_000_000,
```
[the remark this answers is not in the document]
JssDWt:
```suggestion
            amount_msats * self.proportional as u64 / 1_000_000 / 1000 * 1000,
```

libs/sdk-core/src/breez_services.rs
JssDWt: This comment is no longer relevant

libs/sdk-core/src/breez_services.rs
```diff
@@ -1400,6 +1421,8 @@ impl Receiver for PaymentReceiver {
                         )?,
                         incoming_amount_msat: amount_msats as i64,
                         outgoing_amount_msat: (destination_invoice_amount_sats * 1000) as i64,
+                        tag: json!({ "apiKeyHash": api_key_hash }).to_string(),
+                        opening_fee_params: channel_opening_fee_params.clone().map(Into::into),
```
JssDWt, line 1425: Let's add a little check inside this `if open_channel_needed` statement that `channel_opening_fee_params` is never `None` here to avoid any future bugs where we may accidentally pass None.

libs/sdk-core/src/models.rs
```diff
@@ -637,6 +645,191 @@ pub struct ClosedChannelPaymentDetails {
...
+        Ok(())
+    }
+
+    pub(crate) fn get_cheapest_opening_fee_params(&self) -> Result<OpeningFeeParams> {
+        self.vals.first().cloned().ok_or(anyhow!(
+            "The LSP doesn't support opening new channels: Dynamic fees menu contains no values"
+        ))
+    }
+
+    pub(crate) fn get_longest_valid_opening_fee_params(&self) -> Result<OpeningFeeParams> {
```
JssDWt, line 808: Let's rename this to `get_48h_opening_fee_params` to reflect what it is actually doing.

libs/sdk-core/src/swap.rs
```diff
@@ -386,12 +391,14 @@ impl BTCReceiveSwap {
...
-                    swap_info.confirmed_sats,
-                    String::from("Bitcoin Transfer"),
-                    Some(swap_info.preimage),
-                )
-                .await?;
+                .receive_payment(ReceivePaymentRequestData {
+                    amount_sats: swap_info.confirmed_sats,
+                    description: String::from("Bitcoin Transfer"),
+                    preimage: Some(swap_info.preimage),
+                    opening_fee_params: None,
```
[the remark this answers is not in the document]
JssDWt: Above comment is irrelevant now.

What we need to do here:
If `swap_info` doesn't contain `opening_fee_params`, get a new `opening_fee_params` from the LSP that's valid for 48 hours and pass that to `receive_payment`. That would be the new 'backward compatible flow'.

JssDWt (commented):
Removed some outdated comments. Added a few comments about backward compatibility, as we change the way the backward compatibility works.

libs/sdk-core/src/models.rs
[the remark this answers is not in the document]
ok300: Done: 8e987644621b1ec55081634b932e3fcc35021945

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -260,6 +260,41 @@ callback interface EventListener {
...
+    OpeningFeeParams? opening_fee_params;
+};
+
+dictionary BuyBitcoinRequest {
+    BuyBitcoinProvider provider;
+    OpeningFeeParams? opening_fee_params;
+};
+
+dictionary OpeningFeeParamsMenu {
+    sequence<OpeningFeeParams> vals;
```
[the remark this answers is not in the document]
ok300: Done

libs/sdk-core/src/models.rs

libs/sdk-core/src/models.rs
```diff
@@ -637,6 +645,191 @@ pub struct ClosedChannelPaymentDetails {
...
+    pub preimage: Option<Vec<u8>>,
+    pub opening_fee_params: Option<OpeningFeeParams>,
+}
+
+/// Represents a receive payment response.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ReceivePaymentResponse {
+    pub ln_invoice: LNInvoice,
+    pub opening_fee_params: Option<OpeningFeeParams>,
+    pub setup_fees_msat: Option<u64>,
```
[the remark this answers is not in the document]

tools/sdk-cli/src/command_handlers.rs
ok300: We already print them as part of
```rust
            let invoice = sdk()?
                .receive_payment(ReceivePaymentRequest {
                    amount_sats,
                    description,
                    preimage: None,
                    opening_fee_params: None,
                })
                .await?;
            let mut result = serde_json::to_string(&invoice)?;
```

because the `invoice` type is

```rust
pub struct ReceivePaymentResponse {
    pub ln_invoice: LNInvoice,
    pub opening_fee_params: Option<OpeningFeeParams>,
    pub opening_fee_msat: Option<u64>,
}
```

libs/sdk-core/src/models.rs
ok300: Done: 423ac35d161e2c5c85a8d61ec5ab28988f34af82

libs/sdk-core/src/breez_services.rs
ok300: Done: be5f6ab922e1408d10c2ccf19c29183afd22b35b

libs/sdk-core/src/persist/swap.rs
```diff
@@ -110,6 +113,22 @@ impl SqliteStorage {
         Ok(())
     }
 
+    pub(crate) fn update_swap_fees(
```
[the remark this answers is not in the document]
ok300: Done in #355

libs/sdk-core/src/swap.rs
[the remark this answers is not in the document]
ok300: Already covovered in https://github.com/breez/breez-sdk/pull/313#discussion_r1272072796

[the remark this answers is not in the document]
ok300: As discussed, `receive_payment` already defaults internally to looking up the cheapest fee if it's called with `None` as fee arg. Since redeeming a swap happens quasi-instantly (as fast as settling a LN invoice), the cheapest dynamic fee is good enough as feedback.

libs/sdk-core/src/lsp.rs
```diff
@@ -23,29 +24,59 @@ pub struct LspInformation {
     pub fee_rate: f64,
     pub time_lock_delta: u32,
     pub min_htlc_msat: i64,
-    pub channel_fee_permyriad: i64,
     pub lsp_pubkey: Vec<u8>,
-    pub max_inactive_duration: i64,
-    pub channel_minimum_fee_msat: i64,
+    pub opening_fee_params_menu: OpeningFeeParamsMenu,
```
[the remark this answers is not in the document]
roeierez: On second thought there is an advantage for using a struct that holds the list so we can add more fields inside to help the user.
I am OK with keeping the structure but maybe rename it to `opening_fee_options` ?
This is mostly because it is exposed to the binding and reflected in the API.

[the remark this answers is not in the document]
ok300: Done with cd431b8957c6c909a3560f97480e4d091c624d60

The struct is also useful because we can have a validation method on it (validating if the list is ordered, etc, as opposed to validating each entry, which we do separately).

PR #315, opened 2023-07-13
Add missing gRPC fields for `PaymentInformation`

ok300:
- Update the `opening_fee_params` field ID
- Add the missing `tag` field
- Populate the `tag` field with the `api_key` when calling `LspAPI::register_payment`

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1265,17 +1265,14 @@ impl Receiver for PaymentReceiver {
             )?;
             channel_opening_fee_params = Some(ofp.clone());
             channel_fees_msat = Some(ofp.get_channel_fees_msat_for(amount_msats));
-            match channel_fees_msat {
-                Some(channel_fees_msat) => {
-                    info!("zero-conf fee calculation option: lsp fee rate (proportional): {}:  (minimum {}), total fees for channel: {}",
-                ofp.proportional, ofp.min_msat, channel_fees_msat);
+            if let Some(channel_fees_msat) = channel_fees_msat {
+                info!("zero-conf fee calculation option: lsp fee rate (proportional): {}:  (minimum {}), total fees for channel: {}",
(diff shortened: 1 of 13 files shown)


roeierez (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -1361,6 +1363,7 @@ impl Receiver for PaymentReceiver {
                         destination: hex::decode(parsed_invoice.payee_pubkey.clone())?,
                         incoming_amount_msat: amount_msats as i64,
                         outgoing_amount_msat: (destination_invoice_amount_sats * 1000) as i64,
+                        tag: self.config.api_key.clone().unwrap_or_default(),
```
JssDWt, line 1366: It should be a json object like this
```
// apikeyhash = sha256(apikey)

{
    "apiKeyHash": apikeyhash
}
```

JssDWt (commented):
Looks good. Nit on the tag that should be a json object.

libs/sdk-core/src/breez_services.rs
ok300: @erdemyerebasmaz  pointed out that in the lspd, the field expected is

```go
	expected := "{\"apiKey\": \"11111111111111\"}"
```

as per https://github.com/breez/lspd/pull/62/files#diff-7098b568a089d7880d3c1f05c0ae0e995358eb6b84f9a60710020a57e0843fc6R13

Is that the correct format? Or the one you mentioned above with `apiKeyHash`?

JssDWt: It's the mentioned `apiKeyHash`. That test is 'wrong'. 
Here's the breezmobile code that sets the tag.
https://github.com/breez/breez/blob/a7622f5d1437a905eee41ef2c4bfc4efc40d9ccf/account/payments.go#L1552-L1565

JssDWt (approved):

erdemyerebasmaz (approved):

PR #320, opened 2023-07-15
Simplify version management

ok300:
Both packages of the workspace now derive their version from the workspace version.

The version in `libs/Cargo.toml` now defines the version of both `sdk-core` and `sdk-bindings`.


roeierez (approved):
Nice! LGTM

PR #322, opened 2023-07-15
Create new Bindings for Kotlin Multiplatform

angelix:
### Kotlin + Multiplatform

For most users, we recommend integrating the Breez SDK as Gradle dependency from [our Maven repository](https://mvn.breez.technology/releases).

To do so, add the following to your Gradle dependencies:

Set **Breez** maven repository
``` groovy
repositories {
  maven {
      url("https://mvn.breez.technology/releases")
  }
}
```
**Kotlin Multiplatform dependency**
``` kotlin
val commonMain by getting {
  dependencies {
    implementation("technology.breez:breez-sdk-kmp:<version>")
  }
}
```


libs/sdk-bindings/makefile
```diff
@@ -121,7 +121,26 @@ bindings-android: kotlin
 	cd bindings-android && ./gradlew assemble
 	mkdir -p ffi/android
 	cp bindings-android/lib/build/outputs/aar/lib-release.aar ffi/android
-		
+
+bindings-kotlin-multiplatform: ios-universal kotlin
+	cp -r ffi/kotlin/jniLibs/ bindings-kotlin-multiplatform/breez-sdk-kmp/src/androidMain/jniLibs/
+	cp -r kmm/* bindings-kotlin-multiplatform/breez-sdk-kmp/src/
+
+	sed -i '' -e "s/\"breez_sdk\"/\"breez_sdk_bindings\"/" bindings-kotlin-multiplatform/breez-sdk-kmp/src/jvmMain/kotlin/breez_sdk/UniFFILib.kt
```
angelix, line 129: @roeierez this is the only thing i couldn't manage to find a solution. I reside to `sed` for text replacement.

libs/sdk-bindings/build.rs
```diff
@@ -1,6 +1,17 @@
...
 
 fn main() -> Result<()> {
-    uniffi_build::generate_scaffolding("./src/breez_sdk.udl").unwrap();
+    let udl_file = "./src/breez_sdk.udl";
+    let out_dir = Utf8Path::new("kmm");
+    uniffi_build::generate_scaffolding(udl_file).unwrap();
+    uniffi_bindgen::generate_external_bindings(
+        KotlinBindingGenerator {},
+        udl_file,
+        None::<&Utf8Path>,
```
roeierez, line 12: This seems to be the path for the uniff.toml file. Since it is passed with None I am guessing it uses some defaults.
Did you consider pointing to the uniffi.toml we have and add the kmp configuration? Then you can control the library name from there.

roeierez (changes requested):
This looks really great. I dropped a comment that might help with the binary name. Other than that happy to merge this in. 
Thanks a lot!

libs/sdk-bindings/makefile
roeierez: Here is the line where the binary name is taken: https://gitlab.com/trixnity/uniffi-kotlin-multiplatform-bindings/-/blob/main/src/lib.rs?ref_type=heads#L65

It takes the namespace name as the default.
You can override it by setting it on the config as can be seen here: https://gitlab.com/trixnity/uniffi-kotlin-multiplatform-bindings/-/blob/main/src/lib.rs?ref_type=heads#L45

libs/sdk-bindings/build.rs
angelix: Fixed in 52bbe0b0a75d115194d059543cf848030937fd3e

libs/sdk-bindings/makefile
angelix: I added the following to `uniffi.toml` and worked.

```
[bindings.kotlin-native]
package_name = "breez_sdk"
cdylib_name = "breez_sdk_bindings"
```

The key is mentioned here: https://gitlab.com/trixnity/uniffi-kotlin-multiplatform-bindings/-/blob/main/src/lib.rs?ref_type=heads#L56

roeierez (approved):
LGTM

PR #330, opened 2023-07-18
Cache cargo crates and target dependency builds etc.

lnd3v:
- Use the Swatinem/rust-cache@v2 to cache ~/.cargo
- Also cache libs/target and tools/sdk-cli/target

Workflow run now takes 13 minutes.


roeierez (approved):
LGTM! Great improvement.

PR #333, opened 2023-07-18
Method naming convention discussion

erdemyerebasmaz:
As we're releasing packages & plugins for Breez SDK, it's time we have a standardized naming convention to reduce differences between bindings and rethink API's in the perspective of end-user and documentation.

_This PR is to open up these changes for discussion and the changes proposed are not final._

### Changelist:
- Grouped up API's by concept, (Breez Services, LSP, LNURL, Fiat Currency etc.)
- Follow the naming convention of `VerbObject` / `VerbNoun`,
  - There's a downside I noticed with this, since we have a single API class there can be multiple Objects that use the same verb, code completion can be off-putting and confusing. Although, an `ObjectVerb` / `ConceptVerb` naming convention is unorthodox, it's really a breeze to code and is not disruptive for some cases:
  Screenshot 2023-07-18 at 14 41 47Screenshot 2023-07-18 at 14 43 15
- Use [imperative mood](https://www.grammar-monster.com/glossary/imperative_mood.htm) on `Verb` portion of method names,
- Use `Fetch` only on API's that has a network call,
- Renamed `BreezBridge` to `BreezSDK`,
  - `BreezSDK` api's follows the same order of bindings.rs, 

 ---
### Explanation: 
I tried to follow an order that makes sense implementation-wise i.e.,
- Requirements of the API has to be met before it's used e.g., 
  - Node should be connected & Breez Services should be initialized before connection to an LSP. More practical examples:
    - stopNode method must not be above connect
    - getLspId method must not be above connectLSP
- Verification tools have to be present upon first use
  - I'm trying to keep UI in mind and have some conflicting ideas here because of it, which comes first: payment list or the payment?
    - Should users receive a payment and then call listPayments or create a payment list UI using listPayments and see with their own eyes that the payment is shown upon receiving a payment?

and takes our SDK's payment type priority(**lightning=lnurl>on-chain**) into consideration. Which I believe should be reflected on SDK docs as well.

 ### API concept groupings by order:
- Breez Services API's
  - Breez Services Helper API's
- LSP API's
- Backup API's
- Parse API's
- Payment* API's. (Payment as in transaction object, not paying/receiving - This should swap places with Parse API's)
- Lightning Payment API's
- LNURL API's
- Fiat Currency API's
- On-Chain Swap API's
- Refundables API's
- In Progress Swap API's
- Swap Fee API's
- CLI API's
- Helper Methods


libs/sdk-core/src/binding.rs
```diff
@@ -30,68 +30,65 @@ static NOTIFICATION_STREAM: OnceCell<StreamSink<BreezEvent>> = OnceCell::new();
...
+pub fn get_node_state() -> Result<Option<NodeState>> {
+    block_on(async { get_breez_services()?.node_info() })
+}
 
-impl EventListener for BindingEventListener {
-    fn on_event(&self, e: BreezEvent) {
-        if let Some(stream) = NOTIFICATION_STREAM.get() {
-            stream.add(e);
+/// Cleanup node resources and stop the signer.
+pub fn stop_node() -> Result<()> {
```
roeierez, line 64: The equivalent sdk method is disconnect.

libs/sdk-core/src/binding.rs
```diff
@@ -30,68 +30,65 @@ static NOTIFICATION_STREAM: OnceCell<StreamSink<BreezEvent>> = OnceCell::new();
...
-///
-pub fn connect(config: Config, seed: Vec<u8>) -> Result<()> {
-    block_on(async move {
-        let breez_services =
-            BreezServices::connect(config, seed, Box::new(BindingEventListener {})).await?;
-        BREEZ_SERVICES_INSTANCE
-            .set(breez_services)
-            .map_err(|_| anyhow!("static node services already set"))?;
+/// See [breez_services::mnemonic_to_seed]
+pub fn convert_mnemonic_to_seed(phrase: String) -> Result<Vec<u8>> {
```
roeierez, line 77: `mnemonic_to_seed` ?

libs/sdk-core/src/binding.rs
```diff
@@ -107,17 +104,82 @@ pub fn breez_log_stream(s: StreamSink<LogEntry>) -> Result<()> {
...
+
+/*  Backup API's */
+
+/// See [BreezServices::backup]
+pub fn backup() -> Result<()> {
+    block_on(async { get_breez_services()?.backup().await })
+}
+
+/// See [BreezServices::backup_status]
+pub fn get_backup_status() -> Result<BackupStatus> {
```
roeierez, line 145: `backup_status` ?

libs/sdk-core/src/breez_services.rs
```diff
@@ -360,8 +360,8 @@ impl BreezServices {
             .map_err(|err| anyhow!(err))
     }
 
-    /// Fetch a specific payment by its hash.
-    pub async fn payment_by_hash(&self, hash: String) -> Result<Option<Payment>> {
+    /// Get a specific payment by its hash.
+    pub async fn get_payment_by_hash(&self, hash: String) -> Result<Option<Payment>> {
```
roeierez, line 364: For rust it is better to leave `payment_by_hash`. In general we are not using the `get` notation in the sdk so let's be consistent.

roeierez (changes requested):
Great for lifting this! Looks good. My comments are only to match the method name in dart to the rust name.

libs/sdk-core/src/binding.rs
```diff
@@ -30,68 +30,65 @@ static NOTIFICATION_STREAM: OnceCell<StreamSink<BreezEvent>> = OnceCell::new();
...
-    }
-    fn flush(&self) {}
+/// See [BreezServices::sync]
+pub fn sync_node() -> Result<()> {
+    block_on(async { get_breez_services()?.sync().await })
 }
 
-struct BindingEventListener;
+/// See [BreezServices::node_info]
+pub fn get_node_state() -> Result<Option<NodeState>> {
```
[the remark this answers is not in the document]
erdemyerebasmaz: nodeInfo breaks the VerbObject convention I suggested, which is not final but I think this naming is better suited as this method returns the NodeState.
This applies to your other suggestions as well.

This PR isn't only for bringing naming of Dart and rust methods in line but thinking of method naming convention as a whole.

roeierez:
@erdemyerebasmaz what is the status of this? Is it needed?

erdemyerebasmaz:
> @erdemyerebasmaz what is the status of this? Is it needed?

@breez/breez-developers 

This PR and the example changes were not aimed to get merged but be demonstrative of the items I pointed out as a starting point for a discussion. The status is determined by the participants and I don't know where we miscommunicated but there was no interest :)

Setting a naming convention as early as possible is key to keep things organized, especially when we're dealing with this many bindings & generators. We're collaborating with more and more partners each day and we should be conscious of the developer experience. The pictures that show the difference between following `VerbObject / VerbNoun` or a `ObjectVerb / ConceptVerb` standard is a great example of this and something as mundane as naming causing breaking changes if almost never(_unless there was functional change that called for it etc._) a good look in later stages of a project.

It also makes things much more clear to us as we'll try to optimize & debug certain issues. There was a case not too recently where we were trying to figure out poor performance of a function and wondering if it did have a network request, which would have been made obvious if we had a distinction of `Get` & `Fetch` prefix.

PR #335, opened 2023-07-18
Use framework for ios

roeierez:
This PR now use an iOS framework as dependency for flutter iOS platform.
The advantage is that we can now point to to our external pod deployment which allows us to deploy the flutter plugin to our git and later to pub.
I changed the "make ios" to build the local pod for development and the only change for developers to work with the local pod is to rename the` ios/breez_sdk.podspec.dev` to `ios/breez_sdk.podspec`


erdemyerebasmaz (approved):

erdemyerebasmaz:
> I changed the "make ios" to build the local pod for development and the only change for developers to work with the local pod is to rename the` ios/breez_sdk.podspec.dev` to `ios/breez_sdk.podspec`

Could you add this information to README.md?

PR #340, opened 2023-07-19
Add swaps_Info and reverse_swaps_info on restore

roeierez:
We missed reconstructing the cached tables on restore. We only need to add initial entries and on next sync they will be updated to the current state.


ok300:
LGTM.

This worked, but only when explicitly calling `backup` on both instances (the one where a swap was created and the one that is to be "synchronized" to).

The ongoing swap didn't sync after restarting the instances several times. It only worked with `backup`.

roeierez:
This is as expected, we don't sync periodically currently but only when the user calls backup and after restore.

PR #343, opened 2023-07-21
CLI app: Log to file

ok300:
In the CLI app, redirect log output to a file.


tools/sdk-cli/src/main.rs
```diff
@@ -82,3 +80,40 @@ fn show_results(res: Result<String>) {
         Err(err) => error!("Error: {}", err),
     }
 }
+
+fn init_logging() {
+    let target = Box::new(
+        OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open("cli.log")
```
roeierez, line 89: @ok300  Perhaps we create this file under the working dir passed from the user?

tools/sdk-cli/src/main.rs
```diff
@@ -82,3 +80,40 @@ fn show_results(res: Result<String>) {
         Err(err) => error!("Error: {}", err),
     }
 }
+
+fn init_logging() {
```
roeierez, line 84: I was thinking if the sdk logs to a file by default then the cli doesn't need to configure anything and we get the log file for any user of the sdk (not just the cli).

roeierez (changes requested):

tools/sdk-cli/src/main.rs
ok300: Done

tools/sdk-cli/src/main.rs
ok300: Good idea, done

libs/sdk-core/src/breez_services.rs
```diff
@@ -929,6 +933,47 @@ impl BreezServices {
             }
         });
     }
+
+    fn init_logging(working_dir: &str) -> SdkResult<()> {
```
roeierez, line 937: Will this work side by side with the binding logger that we initialize in uniffi_binding.rs? Or one will override another?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: By looking at the failed CI tests

```
thread '<unnamed>' panicked at 'Builder::init should not be called after logger initialized: SetLoggerError(())', /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/env_logger-0.9.3/src/lib.rs:816:14
```

it looks like my change conflicts with the `uniff_binding` log stream.

Looking into ways to make them compatible.

ok300: The main way to support multiple global loggers is https://crates.io/crates/multi_log . However that crate is not maintained since 5 years.

Instead I'm looking into using `try_init` which should fail gracefully if another global logger, like the `uniffi` logstream, got initialized earlier.

roeierez:
@ok300 this approach will make only one of the loggers active which is not what we want.
We want to support logger listener in the bindings in addition to write the logs to a file.
I can think of an approach that will work: 
One logger in sdk-core that logs to a file and if it has a  log listener also trigger the listener.

ok300:
@roeierez is this closer to what you had in mind?

A few notes on the changes:
* Turns out there are two areas that have additional log handling: `binding.rs` and `uniffi_binding.rs`
* Since we need the new log stream listener already when building `BreezServices` (which happens on `connect`), I added it as an extra arg to `connect`.
* `uniffi_binding` was already using `LogStream` as a log event listener, so I re-used that wherever possible.
* `tokio::test`s were a bit tricky, they don't work well with globals, so I added an expectation of failure when setting the global logger in tests
* `uniffi_binding::set_log_stream` cannot allow setting the log stream handler after `connect`, so it was removed in favor of adding the `LogStream` as extra `connect` arg
* `binding.rs::breez_log_stream` saves the `LogStream` in a global variable of `binding.rs`, so to keep it simple, for now I just used that as arg to `connect`

There are a few open points (fixing tests for bindings, etc) but I wanted to get some feedback on the direction first.

roeierez:
@ok300 yes it is closer indeed. I would try to not overload the connect with extra arguments unless really necessary.
Did you consider leave the API as it was before and move the global listener to the sdk core?

ok300:
Tried a different approach:
* `binding.rs` and `uniff_binding.rs` each store their own optional log stream in a `OnceCell`
* this optional custom logger is used when internally calling `BreezeServices::connect`, while keeping the binding exposed `connect` unchanged

roeierez:
> Tried a different approach:
> 
> * `binding.rs` and `uniff_binding.rs` each store their own optional log stream in a `OnceCell`
> * this optional custom logger is used when internally calling `BreezeServices::connect`, while keeping the binding exposed `connect` unchanged

This approach creates two different connect prototypes, one for the binding and one for the rust while IMO the binding API should reflect exactly the rust one.
These iterations made me thinking that initializing the logger at the sdk-core level might not be a good idea as rust users will just use rust logger API to get full control of the logging behavior. I am now thinking I was wrong about pushing to apply a logging behavior at the sdk-core. Also the sdk-core is used as a library inside another application where probably a logger is already implemented there.
I do think we need at the binding languages to implement both log listener and file based logger so I am thinking now exposing the function that initializes the logger (`init_logging` that you have implemented) and from the `set_log_stream` binding API invoke that function passing it the working_dir and the listener.
We can do the same for dart and for the cli. What do you think?

ok300:
Good point, I tweaked the changes as follows:
* `init_logging` can take an optional `app_logger`, such that
  * it covers our `binding` and `uniffi_binding` use-cases
  * it can be used by apps who have their own logger, but still wish to have the SDK log to file
  * it's compatible with apps that have global loggers (they can simply not call this method)
* `binding` and `uniffi_binding` expose it as before, but since setting a logger needs the SDK instance, it's internally called as part of the binding `connect`
* same logic for the CLI (weird to have it as standalone CLI command, so calling it as part of `connect`)

One downside is that log statements coming out of `connect` are not logged, because the logger is initialized on the SDK instance returned by `connect`.

One minor TODO left is to restrict the SDK "log to file" logger to a whitelist of modules (SDK, GL, ..) such that rust apps that call this method don't have their logs included in the SDK log file.

What do you think? Does it make sense, or is it unnecessarily complicated?

I could make it simpler by making the `init_logging` argument non-optional.

roeierez (changes requested):
I think this is the right direction! Almost ready.

libs/sdk-core/src/breez_services.rs
```diff
@@ -929,6 +937,98 @@ impl BreezServices {
...
+    /// If the application is to use it's own logger, but would also like the SDK to log SDK-specific
+    /// log output to a file in the configured `working_dir` (see [Config]), then do not register the
+    /// app-specific logger as a global logger and instead call this method with the app logger as an arg.
+    ///
+    /// ### Errors
+    ///
+    /// An error is thrown if the log file cannot be created in the working directory.
+    ///
+    /// An error is thrown if a global logger is already configured.
+    pub fn init_logging(&self, app_logger: Option<Box<dyn log::Log>>) -> SdkResult<()> {
```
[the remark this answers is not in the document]
ok300: Maybe `init_sdk_logging` is better?

`set_logger` sounds like a setter for the `app_logger`, but it also initializes the SDK logging-to-file when called.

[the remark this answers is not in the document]
roeierez: I see. So let's leave Init_logging. As you said it better describes the goal here.

libs/sdk-bindings/src/uniffi_binding.rs
```diff
@@ -79,12 +68,22 @@ pub fn connect(
 ) -> SdkResult<Arc<BlockingBreezServices>> {
     rt().block_on(async move {
         let breez_services = BreezServices::connect(config, seed, event_listener).await?;
+
+        let uniffi_logger: Option<Box<dyn log::Log>> = match LOG_STREAM.get() {
+            None => None,
+            Some(_) => Some(Box::new(BindingLogger {})),
+        };
+        breez_services.init_logging(uniffi_logger)?;
```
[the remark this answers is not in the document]
roeierez: Yes. We do. The sdk supports multiple instances with different working dir, one for each node id.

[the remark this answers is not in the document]
ok300: I've updated it with:
- no more dependency on a `Config` arg, now it takes `log_dir` directly
- in Rust apps: can and should be called separately from `connect`
- in our bindings: automatically called within `connect`

I've tested it with a Rust app with two SDK instances, both ended up logging to the same file.

Pseudo-code:

```rust
BreezServices::init_logging("logging-dir", None)?;

let sdk_1 = ...;
sdk_1.connect().await?;
info!("[sdk_1] Node info: {}", sdk_1.node_info()?);

let sdk_2 = ...;
sdk_2.connect().await?;
info!("[sdk_2] Node info: {}", sdk_2.node_info()?);

info!("[sdk_1] Node info: {}", sdk_1.node_info()?);
```

The `info` statements, as well as the SDK logs from both instances, ended up in the same log file in `logging-dir`.

roeierez (approved):
LGTM!

dangeross (approved):
I was having issues yesterday when calling `set_log_stream` twice with the Golang bindings was causing a panic. I think this might gracefully handle this 👍

ok300:
@dangeross yes that's the goal.

ademar111190:
Thank you very much @ok300 ; that makes the CLI so much better!!!

PR #348, opened 2023-07-24
Fix Jitpack build

cnixbtc:
When we declare the JNA dependency in our Gradle project, we do it like so:

```
implementation("net.java.dev.jna:jna:5.8.0@aar")
```

Note the `@aar`. From my investigations, only the AAR version of JNA inlcudes the Android platform builds (aarch64, armv7, x86-64, x86) of the necessary `jnidispatch` native library.

But due what seems to be an [issue with Jitpack](https://github.com/jitpack/jitpack.io/issues/5752), the information about using the AAR instead of the JAR is dropped by Jitpack from our module definition file.

From what I can tell, that means that when using Jitpack, the JAR version of JNA will be used and therefore the versions of `jnidispatch` that are needed on the Android platform are missing. Therefore we get the following error when running Android apps:

```
(libjnidispatch.so) not found in resource path (.)
```

A solution to this (based on [this](https://stackoverflow.com/a/54663679/8034839) SO answer) is to manually download and include `jnidispatch` in the build. That way the necessary native libraries for the Android platform can be found. We already explicitly include `libc++_shared.so` so I think including `libjnidispatch.so` should be fine.

Size-wise, this adds about 400KB to the package.

When installing the Breez SDK from other package repositories (our custom one, GitHub), this is not an issue. There, the information to use the AAR version of JNA is kept and we don't run into the "libjnidispatch.so not found" issue. From my tests, manually including `jnidispatch` doesn't break builds installed from these repositories either. It doesn't seem to cause any duplicate symbols or other errors.

## How to Test

Add the Jitpack repository

```
repositories {
    ...
    maven { url 'https://jitpack.io' }
}
```

Then add the Breez SDK. The `c6f838bd89` is equivalent to the 0.1.3 release + this fix:

```
dependencies {
    implementation 'com.github.breez:breez-sdk:c6f838bd89'
}

```

## Why Jitpack

In React Native projects, Jitpack is a default repository. That means users don't need to manually add a repository when installing the Breez SDK plugin.

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/build.gradle.kts
+++ b/libs/sdk-bindings/bindings-android/lib/build.gradle.kts
@@ -45,7 +45,7 @@ publishing {
     repositories {
         maven {
             name = "breezReposilite"
-            url = uri("https://mvn.breez.technology/releases")
+            url = uri("https://mvn.breez.technology/snapshots")
             credentials(PasswordCredentials::class)
             authentication {
                 create<BasicAuthentication>("basic")
--- a/libs/sdk-bindings/bindings-android/lib/build.gradle.kts
+++ b/libs/sdk-bindings/bindings-android/lib/build.gradle.kts
@@ -45,7 +45,7 @@ publishing {
     repositories {
         maven {
             name = "breezReposilite"
-            url = uri("https://mvn.breez.technology/snapshots")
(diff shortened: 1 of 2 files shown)


cnixbtc:
Might have a better solution. Will reopen if needed.

PR #355, opened 2023-07-26
Store swap dynamic fees in separate table

ok300:


libs/sdk-core/src/persist/swap.rs
```diff
@@ -113,19 +120,42 @@ impl SqliteStorage {
...
         &self,
         bitcoin_address: String,
         channel_opening_fees: Option<OpeningFeeParams>,
     ) -> Result<()> {
-        self.get_connection()?.execute(
+        let mut con = self.get_connection()?;
+        let tx = con.transaction()?;
+
+        tx.execute(
             "UPDATE swaps_info SET channel_opening_fees=:channel_opening_fees where bitcoin_address=:bitcoin_address",
```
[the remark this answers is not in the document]
ok300: Done

roeierez (approved):

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -428,6 +428,18 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
...
+        "
+        CREATE TABLE IF NOT EXISTS sync.swaps_fees (
+         bitcoin_address TEXT PRIMARY KEY NOT NULL,
+         created_at TEXT DEFAULT CURRENT_TIMESTAMP NOT NULL,
+         channel_opening_fees TEXT NOT NULL
+        ) STRICT;
+        ",
+        // Remove dynamic fee from swaps_info (fees already stored in swaps_fees)
+        "
+        ALTER TABLE swaps_info DROP COLUMN channel_opening_fees;
```
roeierez, line 441: @ok300  If we didn't merge this branch and the feature yet then we can remove the migration line that added the channel_opening_fees to the swaps_info table and remove this line that drops it.

ok300: Good catch, done

libs/sdk-core/src/persist/sync.rs
```diff
@@ -201,78 +201,255 @@ impl SqliteStorage {
...
+        // The local DB doesn't have a swaps_fees row matching one or more entries in the remote
+        tx.execute(
+            "
+        INSERT OR REPLACE INTO sync.swaps_fees
+         SELECT bitcoin_address, created_at, channel_opening_fees FROM remote_sync.swaps_fees
+          WHERE
+           bitcoin_address NOT IN (SELECT bitcoin_address FROM sync.swaps_fees)
+         ;",
+            [],
+        )?;
```
JssDWt, line 241: I think changing the where clause to this, potentially has less unintended consequences with comparing a datetime with null. And it makes the second query obsolete.

```suggestion
          sync.swaps_fees.created_at is null or remote_sync.swaps_fees.created_at > sync.swaps_fees.created_at
```

JssDWt (approved):
Approved!
One suggestion for the remote sync query.

libs/sdk-core/src/persist/sync.rs
JssDWt: (I selected too many lines to delete. 239-241 should not be deleted)

ok300: Good point, that way the two queries can be merged into one.

Done in aa35a5f1d3e4760bca4d96000b21ade483fad458

JssDWt (approved):

PR #368, opened 2023-07-31
ci: publish android as part of all platforms

JssDWt:
Same CI structure for android as for golang and C#.


ok300 (approved):

.github/workflows/publish-android.yml
```diff
@@ -2,46 +2,84 @@ name: Publish Android Bindings
...
+        description: 'password for gradlew publish'
+        required: true
 
 jobs:
-  publish-android:
+  build-package:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
```
cnixbtc, line 15: @JssDWt One small thing: Publishing the Android bindings to GitHub packages requires these permissions (at least according to [this guide by GitHub](https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven#publishing-packages-to-github-packages)). If we see the publishing step to GitHub packages fail, then we'll most likely need to re-add these three lines.

JssDWt: Good point. I thought, let me remove some stuff I don't understand the purpose for, and it still worked. So I just left it there. If it breaks, we'll change it.

cnixbtc: Yep, that's totally valid. Just wanted to give a pointer to make the fix easy in case it does break on release. Great CI improvements overall. :)

PR #369, opened 2023-07-31
ci: add kotlin mpp to all platforms

JssDWt:
Same procedure as the golang/C#/android publishes, the kotlin-mpp gets the same structure for publish.

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -1,0 +1,62 @@
+name: Build iOS
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'commit/tag/branch reference'
+        required: true
+        type: string
+  workflow_call:
+    inputs:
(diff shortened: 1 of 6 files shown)


ok300 (approved):

.github/workflows/build-ios.yml
```diff
@@ -0,0 +1,62 @@
...
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest-stable
+
+    - name: Build Breez
+      working-directory: libs/sdk-bindings
+      run: cargo build --release --target ${{ matrix.target }}
+    
+    - name: Archive release
+      uses: actions/upload-artifact@v3
```
dangeross, line 59: I think you could archive the dylib outputs also a create universal libs, but that can be done in a another PR

dangeross (approved):
LGTM!

PR #370, opened 2023-08-02
Map GL status ONCHAIN to models::ChannelState::Closed

ok300:
Now two different GL states map a channel state to `models::ChannelState::Closed`: `ONCHAIN` and `CLOSED`.

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -30,10 +30,7 @@ use tokio::sync::{mpsc, Mutex};
 use tonic::Streaming;
 
 use crate::invoice::parse_invoice;
-use crate::models::{
-    Config, GreenlightCredentials, LnPaymentDetails, Network, NodeAPI, NodeState, PaymentDetails,
-    PaymentType, SyncResponse, UnspentTransactionOutput,
-};
+use crate::models::*;
 use crate::persist::db::SqliteStorage;
 use crate::{Channel, ChannelState, NodeConfig};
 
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -929,16 +923,16 @@ fn parse_amount(amount_str: String) -> Result<pb::Amount> {
...
+impl From<pb::Channel> for Channel {
     fn from(c: pb::Channel) -> Self {
         let state = match c.state.as_str() {
-            "OPENINGD" | "CHANNELD_AWAITING_LOCKIN" => crate::models::ChannelState::PendingOpen,
-            "CHANNELD_NORMAL" => crate::models::ChannelState::Opened,
-            "CLOSED" => crate::models::ChannelState::Closed,
-            _ => crate::models::ChannelState::PendingClose,
+            "OPENINGD" | "CHANNELD_AWAITING_LOCKIN" => ChannelState::PendingOpen,
+            "CHANNELD_NORMAL" => ChannelState::Opened,
+            "ONCHAIN" | "CLOSED" => ChannelState::Closed,
```
roeierez, line 931: According to core lightning docs there is no status CLOSED https://docs.corelightning.org/reference/lightning-listpeers

roeierez (changes requested):
Looks good. Just need to remove the CLOSED status as far as I understand the docs

libs/sdk-core/src/greenlight/node_api.rs
ok300: If I search that link for `"CLOSED"` I find:

> The objects in the channels array will have at least these fields:
> - state: Any of these strings: 
>   ...
>   -  "CLOSED": The channel closure has been confirmed deeply. The channel will eventually be removed from this array.

Is this not the same status we're interested in?

JssDWt: @ok300 I agree

JssDWt (approved):

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Right .  My miss.

roeierez (approved):

Issue #371, opened 2023-08-03
Python and Swift language binding tests are very flaky on Linux


JssDWt:
Running the language binding tests on Linux, using `cargo test` in the `libs/sdk-bindings` folder. When doing that on linux, the python and swift tests fail, as a guesstimate, 75% of the time. Running on Darwin seems to be more stable.

Related issue: https://github.com/mozilla/uniffi-rs/issues/1553

I'm not sure whether this is a problem with the language bindings, or with the tests executing the language bindings. All tests seems to fail around async code and threading.

PR #372, opened 2023-08-03
ci: speed up pull request ci

JssDWt:
Speed up the pull request CI, to give more instant feedback about the state of your PR.
The main change is to only compile for the platform the CI is running on. So skip the makefile entirely.

Here's some status screenshots about CI runtimes in the best and worst case respectively. The main effect on the runtime is the presence of rust cache. The first image is a run with an active rust cache with no changes. The second image is where the rust cache is missing. In both scenarios there's a significant improvement in the time to feedback.
Scherm­afbeelding 2023-08-03 om 13 03 43
Scherm­afbeelding 2023-08-03 om 13 21 57

Downsides:
- The makefile is no longer tested out-of-the-box
- This change uses more rust cache resources. Each job has its own rust cache. We'll have to see how that plays out.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,52 +17,151 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3      
-      - run: |
+      - name: Run Format
+        run: |
           cd libs
           cargo fmt -- --check
           cd ../tools/sdk-cli
           cargo fmt -- --check
 
-  build:
-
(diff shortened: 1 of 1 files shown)


JssDWt:
Note: When trying this CI change, I tried to run the language bindings tests on Linux, that was nearly impossible, because almost always python or swift failed. It seems to be much more stable on Darwin. Issue here: https://github.com/breez/breez-sdk/issues/371

dangeross (approved):

PR #375, opened 2023-08-04
ci: build language bindings in a separate workflow

JssDWt:
Add a separate workflow for language bindings, so the compiled sdk can be reused to create language bindings for several languages.

Add a `publish` parameter. If `publish` is false, the packages won't be published to maven, nuget, git, etc.

Add a setup stage for the main `publish-all-platforms.yml`. This makes it easier to test the file if you make changes, because you can set `windows` to true on push for example when testing the workflow.


dangeross (approved):
LGTM! This CI is looking awesome

ok300 (approved):

PR #378, opened 2023-08-04
Use Breez SDK bindings from Jitpack in React Native plugin

cnixbtc:
Finally uses the published Breez SDK Android bindings instead of bundling them with the RN plugin.

There were a couple of issues to fix with Jitpack and also with RN itself. I tried to document it in code but please feel free to ask for details if it is not clear enough. 🙏 

⚠️ Note that with this change, the RN Plugin will only work seamlessly with React Native versions >= 0.71.0 due to a bug in the way RN treats plugins that bundle `libc++_shared.so` (like we do). There is a workaround for people on older RN versions but it requires an additional manual step. I documented that in the readme.

I tested both variants (RN 0.71.0 in a fresh project and RN 0.68.6 in the example project) and they worked fine on my machine. Would be cool if others could test too!

Let me know what you think!

---

After this I will try to investigate the issues we saw with RN + Cocoapods for iOS builds. Once these are fixed we can move our RN publishing to NPM again.


roeierez (approved):
LGTM. Let's wait for @dangeross to review.as well before merging.

dangeross:
We've upgraded the react-native dependency in the package and example project to 0.70.6. Hopefully it's not a big leap to upgrade again to 0.71.0 if needed. I'll do a full review tonight

dangeross (approved):
LGTM! 

I do the same workaround in Satimoto because of conflicts in dependencies. I think ideally we bump the react-native to 0.71.0 to avoid extra steps.

cnixbtc:
> I think ideally we bump the react-native to 0.71.0 to avoid extra steps. 

Let me see if I can update it in the example project 👍 Will do a separate PR though if that is fine with you guys.

Should we bump the peer dependency requirement as well or is that too eager? 🤔

roeierez:
Will this force users to use this RN version? If so I am not sure it is something we want since we have a simple work around for older versions.

PR #386, opened 2023-08-07
Add Android example

ademar111190:
Add an Android example for using breez sdk.

Related: https://github.com/breez/breez-sdk-docs/pull/32


roeierez:
@ademar111190 Looks like a great job you did here!
My concerns are mainly having to maintain another app. We are moving fast and frequently introducing breaking changes and each of these changes will require changing this app.

ademar111190:
Thanks @roeierez 

The android example is using the lib with a specific version set https://github.com/breez/breez-sdk/pull/386/files#diff-3d7e80965408f4fba485928006729381607bb60506ab69f7a04aca387c9dbfe6R9
So no need to change the app until we want to change. It will break nothing in the project.

I'm giving the developers a real app where they can see the lib  being used so they focus on the problem they are working on instead of figuring out details about how to integrate the lib. 

e.g. look into `HomeViewModel` in special how I'm dealing with multi-threading; android is restrictive about using the main thread and our lib is threading blocking; or the types conversion (uLong / uInt / etc …) these things are hard to grasp if the developer had never worked with a language with these types (the case for most android developers); Just to mention two things.

So that is my point, this example is beneficial for outside developers even if it is not working anymore because breaking changes as there are a lot of implementation details that are solved by me (who has the library context and native android experience)

roeierez:
> Thanks @roeierez
> 
> The android example is using the lib with a specific version set https://github.com/breez/breez-sdk/pull/386/files#diff-3d7e80965408f4fba485928006729381607bb60506ab69f7a04aca387c9dbfe6R9 So no need to change the app until we want to change. It will break nothing in the project.
> 
> I'm giving the developers a real app where they can see the lib being used so they focus on the problem they are working on instead of figuring out details about how to integrate the lib.
> 
> e.g. look into `HomeViewModel` in special how I'm dealing with multi-threading; android is restrictive about using the main thread and our lib is threading blocking; or the types conversion (uLong / uInt / etc …) these things are hard to grasp if the developer had never worked with a language with these types (the case for most android developers); Just to mention two things.
> 
> So that is my point, this example is beneficial for outside developers even if it is not working anymore because breaking changes as there are a lot of implementation details that are solved by me (who has the library context and native android experience)

Fair enough.
So currently the question is where to push it. The `binding-android` directory is used to build the android maven package and I don't think it is the right place to add the app.
How about creating a top level directory in the repository `examples` and push the app under `/examples/android` ?

ademar111190:
@roeierez Initially I had added this sample to https://github.com/breez/breez-sdk-docs/pull/32 then I removed from there and opened a PR here
I'm not sure if here is the right place, maybe we should have a separated repo only for examples?
Let me know what you think; If adding here is the right answer, I totally agree by moving to `/examples/android`.

roeierez:
I think both new repo or examples folder are good options. Maybe a new repo is a better start and if we see the maintenance is not a burden we can move it later to examples folder?

ademar111190:
@roeierez sounds good to me.

roeierez:
> @roeierez sounds good to me.

@ademar111190 here you are: https://github.com/breez/breez-sdk-android-example

ademar111190:
Moved the PR to https://github.com/breez/breez-sdk-examples/pull/1

PR #388, opened 2023-08-09
Update docs for dynamic_fees structs and service methods

ok300:


roeierez (approved):

libs/sdk-core/src/lib.rs
```diff
@@ -159,7 +160,8 @@ mod crypt;
 pub mod error;
 mod fiat;
 mod greenlight;
-mod grpc;
+// We expose grpc as pub to be able to reference and re-use its internal field and struct docs
+pub mod grpc;
```
JssDWt, line 164: I think it's best to put the comments on our external models instead. Since grpc is purely for internal use/communication with the lsp, we may keep it private.

ok300: Done in eec2d1356fcc769cadc46d5da49553f9403e4f73, made `grpc` private again and moved the comments from the `proto` to our model structs.

PR #391, opened 2023-08-10
Fix Cocoapod integration in RN plugin

cnixbtc:
It seems that this is all that was missing in the RN plugin to make pulling in the bindings from Cocoapods work. We've had that line in before but I must've accidentally deleted it when setting up the second podfile (the one to use the locally built bindings during development) in [this commit](https://github.com/breez/breez-sdk/pull/263/commits/bb5775ff54cabf4ba101211bbffc27ac60dbf24c).

Anyway, what I did to test was to checkout the 0.1.4 tag to make sure the RN plugin and the bindings line up. Then I applied this PR's diff and:

- ✅ Ran the example app as is without the locally built bindings present (pulling in BreezSDK from Cocoapods).
- ✅ Ran the example app using the plugin's second dev podspec which makes use of the locally built bindings for development.
- ❌ Removed the line added in this PR and reproduced the exact same error we observed before ("react-native-breez-sdk doesn't seem to be linked)

@dangeross Would be much appreciated if you could double check my testing just to rule out any weirdness with my machine (which we've seen before when it came to Cocoapods). 🙏 

Let me know what you think!

After this we should be ready to publish to npm without having to include full copies of the Android/iOS bindings in the package. I'll take care of updating the CI then.

### DIFF
--- a/libs/sdk-react-native/breez_sdk.podspec
+++ b/libs/sdk-react-native/breez_sdk.podspec
@@ -14,6 +14,7 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/breez-sdk/breez-sdk.git", :tag => "#{s.version}" }
 
   s.source_files = "ios/**/*.{h,m,mm,swift}"
+  s.exclude_files = "ios/bindings-swift/**/*.{h,m,mm,swift}"
 
   s.dependency "React-Core"
   s.dependency "BreezSDK", package["version"]
--- a/libs/sdk-react-native/breez_sdk.podspec
+++ b/libs/sdk-react-native/breez_sdk.podspec
@@ -13,6 +13,8 @@ Pod::Spec.new do |s|
   s.platforms    = { :ios => "11.0" }
   s.source       = { :git => "https://github.com/breez-sdk/breez-sdk.git", :tag => "#{s.version}" }
 
+  s.source_files = "ios/**/*.{h,m,mm,swift}"
+
   s.dependency "React-Core"
   s.dependency "BreezSDK", package["version"]
 end
--- a/libs/sdk-react-native/example/package.json
+++ b/libs/sdk-react-native/example/package.json
@@ -9,7 +9,7 @@
     "start": "react-native start",
     "test": "jest",
     "lint": "eslint .",
-    "pods": "cd ios && pod install",
+    "pods": "cd ios && pod install --repo-update",
     "rebuild": "rm -rf node_modules && yarn && yarn pods"
   },
   "dependencies": {


dangeross (commented):
It looks good for the pods, but I get some errors

**Pods**

it didn't work for me by just adding the PR diff to 0.1.4, I initially got this error:
```
[!] CocoaPods could not find compatible versions for pod "BreezSDK":
  In Podfile:
    breez_sdk (from `../node_modules/@breeztech/react-native-breez-sdk`) was resolved to 0.1.4, which depends on
      BreezSDK (= 0.1.4)

None of your spec sources contain a spec satisfying the dependency: `BreezSDK (= 0.1.4)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
```
but after running `pod install --repo-update` it was able to find BreezSDK 0.1.4 and install the pods.

Is it worth adding `pod install --repo-update` into the package.json scripts just in case?

**Build**

After the pods are installed is get errors like this during the build:
```
Cannot convert value of type 'BreezSDK.InputType' to expected argument type 'breez_sdk.InputType'
```
I think this is because outside of `RNBreezSDK.swift` the Breez SDK enums are not explicitly referenced using the import i.e. `InputType` and not `BreezSDK.InputType` which causes the compiler to use the BreezSDK.swift found in ios/bindings-swift. So maybe we need to either explicitly reference BreezSDK or remove `ios/bindings-swift` from the podspec source files.

Also related to the ios/bindings-swift, and I know this is an edge case, but mine contained a newer BreezSDK.swift file (from a previous SDK build) which was also pulled into the iOS build with `"ios/**/*.{h,m,mm,swift}"` and caused the build to fail. I rebuild to 0.1.4 then got the above errors.

roeierez (approved):
Amazing!
Looking forward to publish to NPM

cnixbtc:
Awesome, thanks for testing and sharing your findings @dangeross! 🙏 

> Is it worth adding pod install --repo-update into the package.json scripts just in case?

Sure, no harm in doing so. Will update. 👍 

> which causes the compiler to use the BreezSDK.swift found in ios/bindings-swift. So maybe we need to either explicitly reference BreezSDK or remove ios/bindings-swift from the podspec source files.

Ah yeah that's a good catch. It wouldn't be a problem for the CI or for people using the plugin from NPM since in these cases there wouldn't be any locally built bindings. But for local testing I'd be annoying. It's an easy fix I believe. As you say, we can just [exclude](https://guides.cocoapods.org/syntax/podspec.html#exclude_files) the `ios-bindings/` directory in podspec. Will update that as well 👍

cnixbtc:
Made the two changes @dangeross, please have another look. 🙏 

Tested your scenario of having new, locally built bindings present and could now successfully run the example project.

dangeross (approved):
Great! 🎉 we should be able to test this with an npm publish pretty soon

cnixbtc:
> Great! 🎉 we should be able to test this with an npm publish pretty soon

We could already try publishing 0.1.4 to npm. It's not there yet due to the filesize limitation we ran into. But now it should work if all goes well. That way we could test without needing to bump the core SDK version.

Basically what we'd need to do is checkout 0.1.4, apply this patch and run npm publish.

What do you guys think @roeierez @dangeross?

PR #392, opened 2023-08-11
rn: Update README for local Android dev

dangeross:


erdemyerebasmaz (approved):

libs/sdk-react-native/example/README.md
```diff
@@ -49,7 +49,18 @@ yarn ios
 
 ## Development
 
-To develop locally along with the SDK, you need to switch the podspec from the cocoapods dependencies to the local dependencies.
+To develop locally along with the SDK and test using the example app, you need to make the following changes.
+
+### Android
+
+In `libs/sdk-react-native/android/build.gradle` comment out the line:
```
cnixbtc, line 56: This assumes that the developer re-installs dependencies in the example project afterwards while below part (for iOS) does the change in the already installed `node_modules`. I don't have a preference, both work but we should probably make it consistent.

Similar documentation can also be found here: https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/DEVELOPING.md#using-the-locally-built-bindings

dangeross: Oh, this document is way better! I forgot about it, sorry. I will refactor the README to reference this file

ok300: Would it be possible to replace the Android commenting out part with a conditional check in gradle?

For example, it can be based on the app name ("example").

Or based on an app-specific gradle var, like `isLocal`: https://www.baeldung.com/gradle-conditional-dependencies#configuring-conditional-dependency

dangeross: It's a good idea instead of commenting out. We want the default option for the example app to use the pre-built bindings, then to easily switch to locally built for developing

cnixbtc (approved):

PR #394, opened 2023-08-15
Add `SdkError::LspOpenChannelNotSupported`

ok300:
Add a new error type to allow callers to more reliably identify this specific LSP scenario, like https://github.com/breez/c-breez/issues/618.

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -30,6 +30,7 @@ use crate::lsp::LspInformation;
 use crate::models::Network::*;
 use crate::{LNInvoice, LnUrlErrorData};
 
+use crate::error::{SdkError, SdkResult};
 use strum_macros::{Display, EnumString};
 
 /// Different types of supported payments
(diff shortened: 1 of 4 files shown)


ok300:
Yes, `receive_payment` returns `SdkResult<ReceivePaymentResponse>` so that should preserve and propagate the new error type.

ubbabeck:
> Yes, `receive_payment` returns `SdkResult<ReceivePaymentResponse>` so that should preserve and propagate the new error type.

Awesome thanks

libs/sdk-core/src/models.rs
```diff
@@ -833,9 +841,12 @@ impl OpeningFeeParamsMenu {
...
         // Of those, return the first, which is the cheapest
         // (sorting order of fee params list was checked when the menu was initialized)
-        valid_min_48h.first().cloned().ok_or(anyhow!(
-            "The LSP doesn't support opening fees with a validity > 48 hours"
-        ))
+        valid_min_48h
+            .first()
+            .cloned()
+            .ok_or_else(|| SdkError::LspOpenChannelNotSupported {
+                err: "The LSP doesn't support opening fees with a validity > 48 hours".into(),
```
roeierez, line 848: NIT: maybe:
`The LSP doesn't support opening fees that are valid for at least 48 hours`

roeierez (approved):

libs/sdk-core/src/models.rs

PR #396, opened 2023-08-19
Add optional logging of reqwest raw payloads

ok300:
Adds the ability to log
- every REST endpoint we call GET on
- the raw payload received, independent of JSON parsing

This allows us (or clients) to inspect the received raw response.


JssDWt (approved):
LGTM

Amazing that reqwest doesn't have interceptor functionality yet.

PR #397, opened 2023-08-20
Add utils for reverse swap fees and amounts

ok300:
Add utilities around the `send_onchain` feerates:
* `ReverseSwapPairInfo::estimate_amount_sat_received` lets callers estimate how many sats will arrive, based on the sent amount and current fees
* `ReverseSwapPairInfo::estimate_amount_sat_to_send` lets callers estimate how much to send in order to receive a target amount
* `ReverseSwapPairInfo::feerate` estimates which feerate was used in this `ReverseSwapPairInfo`, to give the caller a reference point for the claim tx feerate argument of `send_onchain` (if the fees and estimated amounts are ok, keep the feerate, else increase or decrease)

These are all methods on the `ReverseSwapPairInfo` struct, so no binding or service method was needed.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -288,6 +288,10 @@ dictionary OpeningFeeParams {
     string promise;
 };
 
+dictionary ReverseSwapFeesRequest {
+    u64? send_amount_sat;
+};
+
 dictionary ReceivePaymentRequest {
     u64 amount_sats;
     string description;
@@ -551,7 +555,7 @@ interface BlockingBreezServices {
    string refund(string swap_address, string to_address, u32 sat_per_vbyte);
 
    [Throws=SdkError]
-   ReverseSwapPairInfo fetch_reverse_swap_fees(u64? send_amount_sat);
+   ReverseSwapPairInfo fetch_reverse_swap_fees(ReverseSwapFeesRequest req);
 
    [Throws=SdkError]
    sequence<ReverseSwapInfo> in_progress_reverse_swaps();
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -379,6 +379,7 @@ dictionary ReverseSwapPairInfo {
     f64 fees_percentage;
     u64 fees_lockup;
     u64 fees_claim;
+    u64? fees_total_estimated;
 };
 
 dictionary ReverseSwapInfo {
@@ -550,7 +551,7 @@ interface BlockingBreezServices {
    string refund(string swap_address, string to_address, u32 sat_per_vbyte);
 
    [Throws=SdkError]
-   ReverseSwapPairInfo fetch_reverse_swap_fees();
+   ReverseSwapPairInfo fetch_reverse_swap_fees(u64? send_amount_sat);
 
    [Throws=SdkError]
    sequence<ReverseSwapInfo> in_progress_reverse_swaps();
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -383,7 +383,7 @@ dictionary ReverseSwapPairInfo {
     f64 fees_percentage;
     u64 fees_lockup;
     u64 fees_claim;
-    u64? fees_total_estimated;
+    u64? total_estimated_fees;
 };
 
 dictionary ReverseSwapInfo {
--- a/libs/sdk-bindings/src/uniffi_binding.rs
+++ b/libs/sdk-bindings/src/uniffi_binding.rs
@@ -277,8 +277,11 @@ impl BlockingBreezServices {
         .map_err(|e| e.into())
     }
 
-    pub fn fetch_reverse_swap_fees(&self) -> SdkResult<ReverseSwapPairInfo> {
-        rt().block_on(self.breez_services.fetch_reverse_swap_fees())
+    pub fn fetch_reverse_swap_fees(
+        &self,
+        send_amount_sat: Option<u64>,
+    ) -> SdkResult<ReverseSwapPairInfo> {
+        rt().block_on(self.breez_services.fetch_reverse_swap_fees(send_amount_sat))
             .map_err(|e| e.into())
     }
 
--- a/libs/sdk-bindings/src/uniffi_binding.rs
+++ b/libs/sdk-bindings/src/uniffi_binding.rs
@@ -16,9 +16,9 @@ use breez_sdk_core::{
     LocalizedName, LogEntry, LogStream, LspInformation, MessageSuccessActionData, MetadataItem,
     Network, NodeConfig, NodeState, OpeningFeeParams, OpeningFeeParamsMenu, Payment,
     PaymentDetails, PaymentFailedData, PaymentType, PaymentTypeFilter, Rate, ReceiveOnchainRequest,
-    ReceivePaymentRequest, ReceivePaymentResponse, RecommendedFees, ReverseSwapInfo,
-    ReverseSwapPairInfo, ReverseSwapStatus, RouteHint, RouteHintHop, SignMessageRequest,
-    SignMessageResponse, SuccessActionProcessed, SwapInfo, SwapStatus, Symbol,
+    ReceivePaymentRequest, ReceivePaymentResponse, RecommendedFees, ReverseSwapFeesRequest,
+    ReverseSwapInfo, ReverseSwapPairInfo, ReverseSwapStatus, RouteHint, RouteHintHop,
+    SignMessageRequest, SignMessageResponse, SuccessActionProcessed, SwapInfo, SwapStatus, Symbol,
     UnspentTransactionOutput, UrlSuccessActionData,
 };
 
@@ -279,9 +279,9 @@ impl BlockingBreezServices {
 
     pub fn fetch_reverse_swap_fees(
         &self,
-        send_amount_sat: Option<u64>,
+        req: ReverseSwapFeesRequest,
     ) -> SdkResult<ReverseSwapPairInfo> {
-        rt().block_on(self.breez_services.fetch_reverse_swap_fees(send_amount_sat))
+        rt().block_on(self.breez_services.fetch_reverse_swap_fees(req))
             .map_err(|e| e.into())
     }
 
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -655,6 +655,11 @@ pub struct ClosedChannelPaymentDetails {
     pub funding_txid: String,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ReverseSwapFeesRequest {
+    pub send_amount_sat: Option<u64>,
+}
+
 /// Represents a receive payment request.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ReceivePaymentRequest {
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -149,7 +149,7 @@ pub struct ReverseSwapPairInfo {
     /// [`crate::ESTIMATED_CLAIM_TX_VSIZE`] vbytes
     pub fees_claim: u64,
     /// Estimated total fees in sats, based on the given send amount. Only set when the send amount is known.
-    pub fees_total_estimated: Option<u64>,
+    pub total_estimated_fees: Option<u64>,
 }
 
 /// Details of past or ongoing reverse swaps, as stored in the Breez local DB
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -28,7 +28,7 @@ use crate::grpc::{self, PaymentInformation, RegisterPaymentReply};
 use crate::lnurl::pay::model::SuccessActionProcessed;
 use crate::lsp::LspInformation;
 use crate::models::Network::*;
-use crate::{LNInvoice, LnUrlErrorData, ESTIMATED_CLAIM_TX_VSIZE};
+use crate::{LNInvoice, LnUrlErrorData};
 
 use crate::error::{SdkError, SdkResult};
 use strum_macros::{Display, EnumString};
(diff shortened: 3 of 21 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -142,12 +142,38 @@ pub struct ReverseSwapPairInfo {
...
-    /// Estimated miner fees in sats for claiming funds
+    /// Estimated miner fees in sats for claiming funds, assuming a transaction virtual size of
+    /// [`crate::ESTIMATED_CLAIM_TX_VSIZE`] vbytes and [ReverseSwapPairInfo::feerate]
     pub fees_claim: u64,
 }
 
+impl ReverseSwapPairInfo {
+    /// Estimates the net amount received at the destination onchain address, for these fees and a given
+    /// `send_amount_sat`
+    pub fn estimate_amount_sat_received(&self, send_amount_sat: u64) -> Result<u64> {
```
roeierez, line 156: How would we expose that to the bindings (uniffi/dart/react native) ?

roeierez (changes requested):
Thanks @ok300. I agree we need to improve this part of fee handling but not sure the implementation is good for the binding layers as is.
Do you think we can add optional parameter to the `fetch_reverse_swap_fees` (amount) and if passed we can calculate the absolute fee amount including all these factors and return it as part of the result?

libs/sdk-core/src/models.rs
ok300: I replaced the `impl ReverseSwapPairInfo` with an extra optional arg in the service method `fetch_reverse_swap_fees`.

libs/sdk-bindings/src/uniffi_binding.rs
```diff
@@ -277,8 +277,11 @@ impl BlockingBreezServices {
         .map_err(|e| e.into())
     }
 
-    pub fn fetch_reverse_swap_fees(&self) -> SdkResult<ReverseSwapPairInfo> {
-        rt().block_on(self.breez_services.fetch_reverse_swap_fees())
+    pub fn fetch_reverse_swap_fees(
+        &self,
+        send_amount_sat: Option<u64>,
```
roeierez, line 282: Let's take the opportunity to change to one struct argument? `ReverseSwapFeesRequest` ?

roeierez (changes requested):
Looks good.

libs/sdk-bindings/src/uniffi_binding.rs
ok300: Good idea, done.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -379,6 +383,7 @@ dictionary ReverseSwapPairInfo {
     f64 fees_percentage;
     u64 fees_lockup;
     u64 fees_claim;
+    u64? fees_total_estimated;
```
roeierez, line 386: NIT: total_estimated_fee

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
```diff
@@ -94,6 +94,11 @@ fun asCheckMessageRequest(reqData: ReadableMap): CheckMessageRequest? {
     return null
 }
 
+fun asReverseSwapFeesRequest(reqData: ReadableMap): ReverseSwapFeesRequest? {
+    val sendAmountSats = reqData.getDouble("sendAmountSats")
```
roeierez, line 98: The sendAmountSats is defined as optional in the udl. I don't know how the code handle now empty value.
Does it throw error? Does it use default value? In both cases it is not the correct behavior.
Maybe @dangeross can help here?

roeierez (changes requested):
Some feedback mainly around RN. I think @dangeross should review as well.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
```diff
@@ -94,6 +94,11 @@ fun asCheckMessageRequest(reqData: ReadableMap): CheckMessageRequest? {
...
+    return ReverseSwapFeesRequest(sendAmountSats)
+}
```
dangeross, line 100: If sendAmountSat is optional, getDouble will return 0 if not set. 
asReverseSwapFeesRequest always returns a ReverseSwapFeesRequest
```
fun asReverseSwapFeesRequest(reqData: ReadableMap): ReverseSwapFeesRequest {
    val sendAmountSats = if (hasNonNullKey(reqData, "sendAmountSats")) reqData.getDouble("sendAmountSats") else null
    return ReverseSwapFeesRequest(sendAmountSats)
}
```

libs/sdk-react-native/ios/RNBreezSDK.swift
```diff
@@ -414,13 +414,17 @@ class RNBreezSDK: RCTEventEmitter {
...
-    @objc(fetchReverseSwapFees:rejecter:)
-    func fetchReverseSwapFees(_ resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
-        do {
-            let fees = try getBreezServices().fetchReverseSwapFees()
-            resolve(fees)        
-        } catch let err {
-            rejectErr(err: err, reject: reject)
+    @objc(fetchReverseSwapFees:resolver:rejecter:)
+    func fetchReverseSwapFees(_ reqData:[String: Any], resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
+        if let reverseSwapFeesRequest = BreezSDKMapper.asReverseSwapFeesRequest(reqData: reqData) {
```
dangeross, line 419: asReverseSwapFeesRequest will always return a ReverseSwapFeesRequest

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -544,14 +544,20 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
...
         executor.execute {
-            try {
-                val reverseSwapFees = getBreezServices().fetchReverseSwapFees()
-                promise.resolve(readableMapOf(reverseSwapFees))
-            } catch (e: SdkException) {
-                e.printStackTrace()
-                promise.reject(e.javaClass.simpleName, e.message, e)
+            val reverseSwapFeesRequest = asReverseSwapFeesRequest(reqData)
+
+            if (reverseSwapFeesRequest == null) {
```
dangeross, line 551: asReverseSwapFeesRequest will always return a ReverseSwapFeesRequest

dangeross (changes requested):

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
dangeross: Commented below on this

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
ok300: Done in b353354db83fbba408645f2007ca2172cbfeba2c

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -211,6 +211,15 @@ class BreezSDKMapper {
...
         return nil
     }
+
+    static func asReverseSwapFeesRequest(reqData: [String: Any?]) -> ReverseSwapFeesRequest? {
+        if let sendAmountSat = reqData["sendAmountSat"] as? UInt64
+        {
+            return ReverseSwapFeesRequest(sendAmountSat: sendAmountSat)
+        }
+
+        return nil
```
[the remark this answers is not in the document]
ok300: Done in b353354db83fbba408645f2007ca2172cbfeba2c

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
ok300: Done in b353354db83fbba408645f2007ca2172cbfeba2c

ok300:
Thanks @dangeross , I added your suggestions and fixes.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
```diff
@@ -94,8 +94,8 @@ fun asCheckMessageRequest(reqData: ReadableMap): CheckMessageRequest? {
     return null
 }
 
-fun asReverseSwapFeesRequest(reqData: ReadableMap): ReverseSwapFeesRequest? {
-    val sendAmountSats = reqData.getDouble("sendAmountSats")
+fun asReverseSwapFeesRequest(reqData: ReadableMap): ReverseSwapFeesRequest {
+    val sendAmountSats = if (hasNonNullKey(reqData, "sendAmountSats")) reqData.getDouble("sendAmountSats") else null
```
dangeross, line 98: Sorry, I didn't realise it should be a ULong
```
val sendAmountSats = if (hasNonNullKey(reqData, "sendAmountSats")) reqData.getDouble("sendAmountSats").toULong() else null
```
or maybe this is clearer
```
fun asReverseSwapFeesRequest(reqData: ReadableMap): ReverseSwapFeesRequest {
    if (hasNonNullKey(reqData, "sendAmountSats")) {
        val sendAmountSats = reqData.getDouble("sendAmountSats")
        return ReverseSwapFeesRequest(sendAmountSats.toULong())
    }
    
    return ReverseSwapFeesRequest(null)
}
```

libs/sdk-react-native/ios/RNBreezSDK.swift
```diff
@@ -414,11 +414,12 @@ class RNBreezSDK: RCTEventEmitter {
...
 
-    @objc(fetchReverseSwapFees:rejecter:)
-    func fetchReverseSwapFees(_ resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
+    @objc(fetchReverseSwapFees:resolver:rejecter:)
+    func fetchReverseSwapFees(_ reqData:[String: Any], resolver resolve: @escaping RCTPromiseResolveBlock, rejecter reject: @escaping RCTPromiseRejectBlock) -> Void {
         do {
-            let fees = try getBreezServices().fetchReverseSwapFees()
-            resolve(fees)        
+            let reverseSwapFeesRequest = BreezSDKMapper.asReverseSwapFeesRequest(reqData: reqData)
+            let fees = try getBreezServices().fetchReverseSwapFees(reqData: reverseSwapFeesRequest)
```
dangeross, line 421: Should be `req` instead of `reqData`

dangeross (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
roeierez: It doesn't sound like english to me. I prefer to make it right even if it is not consistent. I don't mind to change other fields as well to be consistent.

libs/sdk-react-native/ios/RNBreezSDK.swift
ok300: Done in 6771091833e2e23095f676ecb07d7dfe507683b4

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
ok300: Done in 6771091833e2e23095f676ecb07d7dfe507683b4

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
```diff
@@ -680,7 +689,8 @@ fun readableMapOf(reverseSwapPairInfo: ReverseSwapPairInfo): ReadableMap {
             "feesHash" to reverseSwapPairInfo.feesHash,
             "feesPercentage" to reverseSwapPairInfo.feesPercentage,
             "feesLockup" to reverseSwapPairInfo.feesLockup,
-            "feesClaim" to reverseSwapPairInfo.feesClaim
+            "feesClaim" to reverseSwapPairInfo.feesClaim,
+            "feesTotalEstimated" to reverseSwapPairInfo.feesTotalEstimated
```
dangeross, line 693: Needs renaming to `totalEstimatedFees`

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -681,7 +686,8 @@ class BreezSDKMapper {
             "feesHash": reverseSwapPairInfo.feesHash,
             "feesPercentage": reverseSwapPairInfo.feesPercentage,
             "feesLockup": reverseSwapPairInfo.feesLockup,
-            "feesClaim": reverseSwapPairInfo.feesClaim
+            "feesClaim": reverseSwapPairInfo.feesClaim,
+            "feesTotalEstimated": reverseSwapPairInfo.feesTotalEstimated
```
dangeross, line 690: Needs renaming to `totalEstimatedFees`

libs/sdk-react-native/src/index.tsx
```diff
@@ -416,6 +420,7 @@ export type ReverseSwapPairInfo = {
     feesPercentage: number
     feesLockup: number
     feesClaim: number
+    feesTotalEstimated?: number
```
dangeross, line 423: Needs renaming to `totalEstimatedFees`

dangeross (changes requested):
Just the RN `totalEstimatedFees` renaming then good to go

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt
ok300: Done in f02c462fb1406ad4f07a94a1a9dfb440e140bec1

libs/sdk-react-native/ios/BreezSDKMapper.swift
ok300: Done in f02c462fb1406ad4f07a94a1a9dfb440e140bec1

libs/sdk-react-native/src/index.tsx
ok300: Done in f02c462fb1406ad4f07a94a1a9dfb440e140bec1

roeierez (approved):

PR #398, opened 2023-08-23
Add correct routing hint in reverse swaps invoice

ok300:
This PR changes the routing hint used when creating a new reverse swap. Before it was the LSP node ID, now it is the reverse routing node ID.

Changes include:
* renamed existing `ReverseSwapperAPI` to `ReverseSwapServiceAPI` for consistency. In the case of normal swaps, the `SwapperAPI` is used for `BreezServer` API calls. Therefore, for consistency reasons, it makes sense to reserve the `ReverseSwapperAPI` name for `BreezServer` calls as well.
* introduced `ReverseSwapperAPI` which exposes a simple call to the `BreezServer` API, a [getter for the reverse routing node](https://github.com/breez/breez/blob/7c8560ca2dc18c65e3ccc2b0cd470328ba1d044c/breez/breez.proto#L58)

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1068,15 +1068,15 @@ impl BreezServices {
             .parse_filters(
                 r#"
                 info,
-                gl_client=info,
+                gl_client=warn,
                 h2=warn,
                 hyper=warn,
                 breez_sdk_core::reverseswap=info,
-                lightning_signer=info,
+                lightning_signer=warn,
                 reqwest=warn,
                 rustls=warn,
                 rustyline=warn,
-                vls_protocol_signer=info
+                vls_protocol_signer=warn
             "#,
             )
             .format(|buf, record| {
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -584,19 +584,15 @@ impl BreezServices {
         sat_per_vbyte: u64,
     ) -> Result<ReverseSwapInfo> {
         match self.in_progress_reverse_swaps().await?.is_empty() {
-            true => {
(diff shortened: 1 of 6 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -584,12 +585,13 @@ impl BreezServices {
     ) -> Result<ReverseSwapInfo> {
         match self.in_progress_reverse_swaps().await?.is_empty() {
             true => {
+                let reverse_routing_node = self.btc_send_swapper.reverse_swapper_api.fetch_reverse_routing_node().await?;
```
roeierez, line 588: If we already have the reverse_swapper_api inside the btc_send_swapper then we should use it internally to get the reverse routing node rather then fetch it at the BreezServices level.

libs/sdk-core/src/models.rs
```diff
@@ -387,9 +388,27 @@ impl TryFrom<i32> for ReverseSwapStatus {
     }
 }
 
-/// Trait covering functionality involving swaps
+/// Trait covering Breez Server reverse swap functionality
 #[tonic::async_trait]
 pub(crate) trait ReverseSwapperAPI: Send + Sync {
```
roeierez, line 393: Perhaps to better differentiate between the two similar named traits we can rename this to: `ReverseSwapperRoutingAPI` ?

libs/sdk-core/src/breez_services.rs
ok300: Done in 88946bc63ca7c7b6510fb405d1d8864fbbcfa561

libs/sdk-core/src/models.rs
ok300: Done in f107829294d848f9b36f63af9ce713a593d3fdef

roeierez (approved):

PR #401, opened 2023-08-23
fix sync db swap migration

roeierez:
The migration for the sync database should go to the sync migration rather than the main db migration so the remote backups can be migrated as well.
I believe this was the issue with the error "table swap_fees not found" that we saw on some devices.
I left the previous migration in the main db so already migrated databases won't fail.

### DIFF
--- a/libs/sdk-core/src/persist/migrations.rs
+++ b/libs/sdk-core/src/persist/migrations.rs
@@ -423,15 +423,15 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
          data BLOB NOT NULL,
          created_at TEXT DEFAULT CURRENT_TIMESTAMP
         ) STRICT;
-       ",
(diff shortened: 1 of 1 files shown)


dangeross:
This is the specific error that you mentioned:
```
event backupFailed: {"error": "no such table: remote_sync.swaps_fees"}
```

dangeross (approved):
From what i understand the current_sync_migrations() are used to migrate the local sync db, then remote sync db. After that local is synced to remote.

ok300:
Right now there are 2 identical tables being created (and maintained).

That means any future changes to the table structure have to be applied 2x, once to each table, no?

roeierez:
Only one table is created, it is the same table and we just moved the statement to the right place. The new place is right because we apply these migrations on sync as well. Before this change the migration wasn't executed on the remote sync and that's why the backup failed.

PR #402, opened 2023-08-23
Skip re-encoding invoice when not needed

### LINKED ISSUE #399: bolt11 in event `INVOICE_PAID` event is not correct
1) Create invoice with `bolt11_a, payment_hash_a`
2) Pay that invoice from any other wallet
3) Receive `INVOICE_PAID` with `bolt11_b, payment_hash_b` and notice that `bolt11_a != bolt11_b` but `payment_hash_a == payment_hash_b`

Seems like something is changing the invoice retroactively? But still signing it correctly.

Two examples

1. The one I create: https://lightningdecoder.com/lnbc10n1pjwvrntdq2f38xy6t5wvpp50udjtr576fjgm3drvt5zhqwax7ap3c4h3ehmwug9cmngdpwy8l9qxqyjw5qsp58s0yxg6szwhfxeg5s3kxvkt9eaaugejtdzpttsef5qwxg06evegs9qrsgqcqpxrzjqtypret4hcklglvtfrdt85l3exc0dctdp4qttmtcy5es3lpt6uts6s0yarww0ua5wyqqqqlgqqqqqzsqygqv750vuv5zn734feczlf95t4v8sdzyjj67lzvwypjr627vcj9zypddk5yyawyahztvgfh6dxt9kg8zxfn42w4kjqlp0sc9l764d24aqpwzdh5s

2. The one I get from the event listener: https://lightningdecoder.com/lnbc10n1pjwvrntsp58s0yxg6szwhfxeg5s3kxvkt9eaaugejtdzpttsef5qwxg06evegspp50udjtr576fjgm3drvt5zhqwax7ap3c4h3ehmwug9cmngdpwy8l9qdq2f38xy6t5wvxqyjw5qcqpxrzjqtypret4hcklglvtfrdt85l3exc0dctdp4qttmtcy5es3lpt6uts6s0yarww0ua5wyqqqqlgqqqqqzsqyg9qxpqysgq76dstp34j9kgtsnsvmwftalxvqt02jdwqxzqjagwaza0f2p4knxkzdhdx6xg3xvte545h4nsd3jayxpjhppca6qrehj7zdad4f25zjsp2ym7y6

roeierez:
When opening a channel or adding routing hints we may need to re-encode the invoice before sending it to the payer. 
If the above is not the case no reason to re-encode it.
fixed #399

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1553,15 +1553,18 @@ impl Receiver for PaymentReceiver {
             });
         }
 
-        // create the large amount invoice
-        let raw_invoice_with_hint =
-            add_lsp_routing_hints(invoice.bolt11.clone(), lsp_hint, amount_sats * 1000)?;
+        // We only create a new invoice if we need to add the lsp hint or change the amount
+        if lsp_hint.is_some() || amount_sats != destination_invoice_amount_sats {
+            // create the large amount invoice
+            let raw_invoice_with_hint =
+                add_lsp_routing_hints(invoice.bolt11.clone(), lsp_hint, amount_sats * 1000)?;
 
-        info!("Routing hint added");
-        let signed_invoice_with_hint = self.node_api.sign_invoice(raw_invoice_with_hint)?;
-        info!("Signed invoice with hint = {}", signed_invoice_with_hint);
+            info!("Routing hint added");
+            let signed_invoice_with_hint = self.node_api.sign_invoice(raw_invoice_with_hint)?;
+            info!("Signed invoice with hint = {}", signed_invoice_with_hint);
 
-        parsed_invoice = parse_invoice(&signed_invoice_with_hint)?;
+            parsed_invoice = parse_invoice(&signed_invoice_with_hint)?;
+        }
 
         // register the payment at the lsp if needed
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1553,15 +1553,18 @@ impl Receiver for PaymentReceiver {
             });
         }
 
-        // create the large amount invoice
-        let raw_invoice_with_hint =
-            add_lsp_routing_hints(invoice.bolt11.clone(), lsp_hint, amount_sats * 1000)?;
+        // We only create a new invoice if we need to add the lsp hint or change the amount
+        if lsp_hint.is_some() || amount_sats != destination_invoice_amount_sats {
```
dangeross, line 1557: It seems to me the first created invoice never has routing hints, so won't this always happen as the LSP routing hint **has** to be added?

roeierez: For the next invoices (when there is enough liquidity) we won't enter this block as gl will populate the hints.
Only for invoices that opens channels will be encoded.
On some rate cases gl doesn't populate the hints even though the channel exists and this is why we add them manually

dangeross (approved):
Looks good. Thanks for clarifying

PR #407, opened 2023-08-26
Add CI for publishing React Native plugin

cnixbtc:
Adds CI for uploading the RN plugin to NPM and integrates that into the main "publish all packages" workflow.

⚠️ Note that the [iOS publishing CI is still external](https://github.com/breez/breez-sdk-swift/blob/main/.github/workflows/publish-swift-package.yaml) and that the RN plugin also requires publishing the respective Cocoapod. That means that while we can publish the RN plugin to NPM independently from the iOS workflow being run or not, for the plugin to actually function one also needs to run the iOS publishing workflow. 
👉  We could (and probably should) of course also bring the iOS CI into the main repo to automate all of this this.

This PR also fixes an issue where the Android package dependency was hardcoded in the RN plugin. Now it is dynamically loaded from the RN plugin's `package.json`. This is [also what we do](https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/breez_sdk.podspec#L20) for the Cocoapod dependency.

### DIFF
--- a/.github/workflows/publish-all-platforms.yml
+++ b/.github/workflows/publish-all-platforms.yml
@@ -65,7 +65,7 @@ jobs:
       golang-package-version: ${{ inputs.golang-package-version || '0.0.2' }}
       maven-package-version: ${{ inputs.maven-package-version || '0.0.2' }}
       kotlin-mpp-package-version: ${{ inputs.kotlin-mpp-package-version || '0.0.2' }}
-      react-native-package-version: ${{ inputs.react-native-package-version }}
+      react-native-package-version: ${{ inputs.react-native-package-version || '0.0.2' }}
       publish: ${{ inputs.publish }}
     steps:
       - run: echo "set output variables"
--- a/.github/workflows/publish-all-platforms.yml
+++ b/.github/workflows/publish-all-platforms.yml
@@ -175,11 +175,11 @@ jobs:
       BREEZ_MVN_USERNAME: ${{ secrets.BREEZ_MVN_USERNAME }}
       BREEZ_MVN_PASSWORD: ${{ secrets.BREEZ_MVN_PASSWORD }}
 
+  # react native version x.y.z will at runtime require
+  # ios and android packages x.y.z being published already.
   publish-react-native:
     needs:
       - setup
-      - publish-maven
-      # todo: - publish-cocoapods
     if: ${{ needs.setup.outputs.react-native == 'true' }}
     uses: ./.github/workflows/publish-react-native.yml
     with:
--- a/.github/workflows/publish-all-platforms.yml
+++ b/.github/workflows/publish-all-platforms.yml
@@ -26,6 +26,10 @@ on:
         description: 'version for the kotlin multiplatform package (MAJOR.MINOR.BUILD)'
         required: false
         type: string
+      react-native-package-version:
+        description: 'version for the react native package (MAJOR.MINOR.BUILD)'
+        required: false
+        type: string
       publish:
         description: 'boolean indicating whether packages should be published. true to publish. false to build only. Default = false.'
         required: false
@@ -54,12 +58,14 @@ jobs:
       golang: ${{ !!inputs.golang-package-version }}
       maven: ${{ !!inputs.maven-package-version }}
       kotlin-mpp: ${{ !!inputs.kotlin-mpp-package-version }}
+      react-native: ${{ !!inputs.react-native-package-version }}
       ref: ${{ inputs.ref || github.sha }}
       csharp-package-version: ${{ inputs.csharp-package-version || '0.0.2' }}
       csharp-ref: ${{ inputs.csharp-ref || inputs.ref || github.sha }}
       golang-package-version: ${{ inputs.golang-package-version || '0.0.2' }}
       maven-package-version: ${{ inputs.maven-package-version || '0.0.2' }}
       kotlin-mpp-package-version: ${{ inputs.kotlin-mpp-package-version || '0.0.2' }}
+      react-native-package-version: ${{ inputs.react-native-package-version }}
       publish: ${{ inputs.publish }}
     steps:
       - run: echo "set output variables"
@@ -168,3 +174,17 @@ jobs:
     secrets:
       BREEZ_MVN_USERNAME: ${{ secrets.BREEZ_MVN_USERNAME }}
       BREEZ_MVN_PASSWORD: ${{ secrets.BREEZ_MVN_PASSWORD }}
+
+  publish-react-native:
+    needs:
+      - setup
+      - publish-maven
+      # todo: - publish-cocoapods
+    if: ${{ needs.setup.outputs.react-native == 'true' }}
+    uses: ./.github/workflows/publish-react-native.yml
+    with:
+      ref: ${{ needs.setup.outputs.ref }}
+      package-version: ${{ needs.setup.outputs.react-native-package-version }}
+      publish: ${{ needs.setup.outputs.publish == 'true' }}
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-react-native.yml
+++ b/.github/workflows/publish-react-native.yml
@@ -1,0 +1,52 @@
+name: Publish React Native Plugin
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: 'commit/tag/branch reference'
+        required: true
+        type: string
+      package-version:
+        description: 'version for the npm package (MAJOR.MINOR.BUILD)'
+        required: true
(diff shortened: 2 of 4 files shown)


.github/workflows/publish-all-platforms.yml
```diff
@@ -168,3 +174,17 @@ jobs:
...
+      - publish-maven
+      # todo: - publish-cocoapods
+    if: ${{ needs.setup.outputs.react-native == 'true' }}
+    uses: ./.github/workflows/publish-react-native.yml
+    with:
+      ref: ${{ needs.setup.outputs.ref }}
+      package-version: ${{ needs.setup.outputs.react-native-package-version }}
+      publish: ${{ needs.setup.outputs.publish == 'true' }}
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
cnixbtc, line 190: This still needs to be setup. cc @roeierez

libs/sdk-react-native/android/build.gradle
```diff
@@ -1,3 +1,12 @@
+import groovy.json.JsonSlurper
+
+def getVersionFromNpmPackage() {
+    def inputFile = new File("$rootDir/../node_modules/@breeztech/react-native-breez-sdk/package.json")
+    def packageJson = new JsonSlurper().parseText(inputFile.text)
+
+    return packageJson["version"]
+}
+
```
cnixbtc, line 9: I could imagine that at some point we'd like to have more flexibility and let the RN plugin version differ from the Android/iOS versions that it depends on. But for now I think it is fine to have them in sync.

.github/workflows/publish-react-native.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Make sure we publish the version as specified
+        working-directory: libs/sdk-react-native
+        run: npm --no-git-tag-version --allow-same-version version ${{ inputs.package-version }}
+
+      - name: Install dependencies
+        working-directory: libs/sdk-react-native
+        run: yarn
+
+      - name: Publish package to npm
```
JssDWt, line 47: Does the npm package need any compiled binaries included?

.github/workflows/publish-all-platforms.yml
```diff
@@ -168,3 +174,17 @@ jobs:
     secrets:
       BREEZ_MVN_USERNAME: ${{ secrets.BREEZ_MVN_USERNAME }}
       BREEZ_MVN_PASSWORD: ${{ secrets.BREEZ_MVN_PASSWORD }}
+
+  publish-react-native:
+    needs:
+      - setup
+      - publish-maven
```
JssDWt, line 181: The dependency on `publish-maven` needs some additional work in the Setup step.

That would need in the setup stage for example:
```
maven: ${{ !!inputs.maven-package-version || !!inputs.react-native-package-version }}
```

Same goes for dependencies of `publish-maven`, so android and the specific language bindings you need.

One thing what will happen then is; the maven package will actually be published, because `publish`=true. But you probably don't want that to be published if you're only building react-native. Maybe in the setup stage we can make specific booleans for every platform. Like `publish-react-native`, `publish-kotlin-mpp`, `publish-maven`, etc. And pass those specific booleans in the correcponding `publish` parameters of each job. (if that makes sense)

JssDWt (changes requested):

.github/workflows/publish-all-platforms.yml
cnixbtc: Makes sense, thanks for the explanation. 🙏 

Thinking of this again, the `publish-react-native` step doesn't strictly _need_ the `publish-maven` step to run beforehand. The `publish-react-native` step will success and the NPM package will be published but at runtime it will require the respective Android and iOS packages to be available. At the moment those are assumed to be released in lockstep: RN plugin x.y.z requires Android and iOS package x.y.z.

That's why I added the dependency at build time just to make it more explicit that publishing the RN package alone doesn't make sense. But I see now that this is not so easy to model here.

Would it make sense to add `maven: ${{ !!inputs.maven-package-version || !!inputs.react-native-package-version }}` like you suggested (also for all dependencies of that step) and then accept that all those will be build and published every time we publish react native?

What do you think?

.github/workflows/publish-react-native.yml
cnixbtc: Nope, not anymore. 🙌

.github/workflows/publish-all-platforms.yml
cnixbtc: Ok reconsidering this again. I removed the build time dependencies, we cannot (yet) enforce it for ios anyway so it's not very useful to _just_ do it for maven I guess. I think for the time being the flexibility to publish RN alone is better. We just published 0.2.0 and probably want to publish RN 0.2.0. For now whoever publishes just needs to know that this runtime dependency exists. 

I can also write that down in somewhere in a releasing doc or so. Thoughts? (also cc @roeierez to keep you in the loop when you do the publishing)

JssDWt: Interesting question. I guess it makes sense to not add the build dependency in the pipeline then if it's not strictly needed in order to publish a new version. It may lead to a small time of inconsistency if we publish the new react native version before publishing the new maven version, but I think that's ok?

.github/workflows/publish-all-platforms.yml
```diff
@@ -54,12 +58,14 @@ jobs:
...
       maven: ${{ !!inputs.maven-package-version }}
       kotlin-mpp: ${{ !!inputs.kotlin-mpp-package-version }}
+      react-native: ${{ !!inputs.react-native-package-version }}
       ref: ${{ inputs.ref || github.sha }}
       csharp-package-version: ${{ inputs.csharp-package-version || '0.0.2' }}
       csharp-ref: ${{ inputs.csharp-ref || inputs.ref || github.sha }}
       golang-package-version: ${{ inputs.golang-package-version || '0.0.2' }}
       maven-package-version: ${{ inputs.maven-package-version || '0.0.2' }}
       kotlin-mpp-package-version: ${{ inputs.kotlin-mpp-package-version || '0.0.2' }}
+      react-native-package-version: ${{ inputs.react-native-package-version }}
```
JssDWt, line 68: I think it's good to give this a default version, just for testing purposes.
If you run this workflow through the 'push' trigger for example, there are no inputs, so having a default helps with some edge cases.

```suggestion
      react-native-package-version: ${{ inputs.react-native-package-version || '0.0.2' }}
```

libs/sdk-react-native/android/build.gradle
JssDWt: Does this mean that in order to set the react native version, you have to do a commit with the new version? Or does this take the version from here https://github.com/breez/breez-sdk/pull/407/files#diff-f44ac18980dab182be3760a350b5fc0a5218d4e93aee37d12d56d3bd8bdabf70R41 ?

.github/workflows/publish-react-native.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+          scope: '@breeztech'
+
+      - name: Make sure we publish the version as specified
+        working-directory: libs/sdk-react-native
+        run: npm --no-git-tag-version --allow-same-version version ${{ inputs.package-version }}
```
JssDWt, line 41: Do set a default version here, in case you want to test the build through a push command (in that case there are no input parameters.
```suggestion
        run: npm --no-git-tag-version --allow-same-version version ${{ inputs.package-version || '0.0.2' }}
```

JssDWt (commented):
I think it looks good. 

Did you have a chance to test? You can test the build, by temporarily changing the workflow to trigger on 'push'.

libs/sdk-react-native/android/build.gradle
cnixbtc: The correct way would be to set the version in the `package.json` and commit that.

But if you don't do that, the line you linked will overwrite whatever is specified in that file. 

TL;DR: it's going to use the version specified in the workflow run.

.github/workflows/publish-all-platforms.yml
cnixbtc: Agreed. I think that is ok. We just need to be careful when publishing, not to publish RN when there's not a corresponding Android/iOS package.

cnixbtc:
> Did you have a chance to test? You can test the build, by temporarily changing the workflow to trigger on 'push'.

I tested by forking the repo, changing the `package.json` so that it would publish to `@cnixbtc/react-native-breez-sdk`, and running the workflow. It successfully succeeded. ✅

JssDWt (approved):

PR #411, opened 2023-08-28
ci: Add android bindings to publish-golang

dangeross:
This PR updates the publish-golang CI to depend on the build-android job and download the android binding artifacts into the breez-sdk-go repository.

See PR breez/breez-sdk-go#11 which adds the android lib structure to breez-sdk-go.

### DIFF
--- a/.github/workflows/publish-all-platforms.yml
+++ b/.github/workflows/publish-all-platforms.yml
@@ -45,7 +45,7 @@ jobs:
       windows: ${{ !!inputs.csharp-package-version || !!inputs.golang-package-version }}
       darwin: ${{ !!inputs.csharp-package-version || !!inputs.golang-package-version }}
       linux: ${{ !!inputs.csharp-package-version || !!inputs.golang-package-version }}
-      android: ${{ !!inputs.kotlin-mpp-package-version || !!inputs.maven-package-version }}
+      android: ${{ !!inputs.kotlin-mpp-package-version || !!inputs.maven-package-version || !!inputs.golang-package-version }}
       ios: ${{ !!inputs.kotlin-mpp-package-version || !!inputs.maven-package-version }}
       kotlin: ${{ !!inputs.kotlin-mpp-package-version || !!inputs.maven-package-version }}
       swift: false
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-golang.yml
```diff
@@ -67,6 +87,10 @@ jobs:
           git config --global user.email github-actions@github.com
           git config --global user.name github-actions
           git add breez_sdk/breez_sdk.go
+          git add breez_sdk/lib/android-386/libbreez_sdk_bindings.so
+          git add breez_sdk/lib/android-aarch/libbreez_sdk_bindings.so
+          git add breez_sdk/lib/android-aarch64/libbreez_sdk_bindings.so
+          git add breez_sdk/lib/android-amd64/libbreez_sdk_bindings.so
```
JssDWt, line 93: It could be that android needs `libc++_shared.so` to be available. It is also included in the downloaded archives next to `libbreez_sdk_bindings.so`.

JssDWt (commented):
Looking good!

JssDWt (approved):

.github/workflows/publish-golang.yml
dangeross: Good catch, seems like the any needed. I get linking errors otherwise

.github/workflows/publish-all-platforms.yml
```diff
@@ -125,6 +125,7 @@ jobs:
   publish-golang:
     needs: 
       - setup
+      - build-android
```
roeierez, line 128: In order to build android don't we need also to add also the golang condition in the setup.android output variable?

dangeross: Like this?
```
      android: ${{ !!inputs.kotlin-mpp-package-version || !!inputs.maven-package-version || !!inputs.golang-package-version }}
```

roeierez: Ah.. I see now you don't need it because you copy the binaries directly.
In the flutter publish I am copying the jniLibs which is built on android workflow that includes the c++ shared.

dangeross: I think you're right @roeierez. As publish-golang downloads the artifacts uploaded by the build job of build-android.  Can you double check @JssDWt?

JssDWt: Yes! Correct, I missed that. The artifacts are only available for the workflow you're running, so if golang needs android, it needs android to build.

PR #413, opened 2023-08-30
publish flutter CI

roeierez:


.github/workflows/publish-flutter.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ inputs.ref || github.sha }}
+          path: breez-sdk
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: android-jniLibs
+          path: flutter/sdk-flutter/android/src/main
+
```
JssDWt, line 45: Is this missing ios artifacts?

.github/workflows/publish-flutter.yml
```diff
@@ -0,0 +1,62 @@
...
+        working-directory: flutter
+        run: |
+          git config --global user.email github-actions@github.com
+          git config --global user.name github-
+          rm -f ../breez-sdk/libs/sdk-flutter/ios/breez_sdk.podspec.dev
+          mv ../breez-sdk/libs/sdk-flutter/ios/breez_sdk.podspec.production ../breez-sdk/libs/sdk-flutter/ios/breez_sdk.podspec
+          cp -r ../breez-sdk/libs/sdk-flutter/ios .
+          cp -r ../breez-sdk/libs/sdk-flutter/android .
+          cp -r ../breez-sdk/libs/sdk-flutter/lib .
+          cp ../breez-sdk/libs/sdk-flutter/pubspec.yaml .
```
JssDWt, line 56: The pubspec.yaml probably needs a version bump?

JssDWt (commented):
What is the purpose of the `breez-sdk-flutter` repo? Is it a clone of libs/sdk-flutter with binaries included? If that's the only reason we need it, we could also make CI put the binaries in the right place inside the breez-sdk repo, and then publish to pub.dev.

.github/workflows/publish-flutter.yml
roeierez: No, iOS is not pushed to this repo and the dependency is reflected via podfile.

roeierez:
> What is the purpose of the `breez-sdk-flutter` repo? Is it a clone of libs/sdk-flutter with binaries included? If that's the only reason we need it, we could also make CI put the binaries in the right place inside the breez-sdk repo, and then publish to pub.dev.

breez-sdk-flutter is the flutter plugin published to git.Currently we don't have a published version to pub.dev.
The next step is indeed to publish to pub.dev and then if we figure out that we don't need the git repo for the plugin basis then I agree we can change the CI to use the folder inside breez-sdk directly.

.github/workflows/publish-flutter.yml
roeierez: Good point!

roeierez: Fixed

.github/workflows/publish-flutter.yml
```diff
@@ -0,0 +1,73 @@
...
+          cp ../breez-sdk/libs/sdk-flutter/pubspec.lock .          
+           
+      - name: Tag the Flutter package        
+        working-directory: flutter
```
JssDWt, line 63:
```suggestion
        working-directory: flutter
        if: ${{ inputs.publish }}
```

JssDWt (commented):
Last nit then, I think it's good!

JssDWt (approved):

PR #415, opened 2023-08-31
Update deplay before triggering jitpack build

cnixbtc:
Sometimes it seems like MVN needs more time to process the artifacts before Jitpack can download them. Let's try with a 1min delay.


JssDWt:
Can we do something like polling until the artifacts are available?

dangeross (approved):
Agree with @JssDWt, maybe there is a way to poll that Maven is ready, or, recover and repeat a trigger to Jitpack if the build fails.

cnixbtc:
Good points, I will check!

roeierez:
@cnixbtc What is the status here?

cnixbtc:
Thanks for the ping @roeierez and sorry for the delay.

I've updated the Jitpack trigger to poll our maven repo for the availability of the package before triggering Jitpack. Thanks for the suggestion! Let me know what you guys think.

roeierez (approved):
LGTM

PR #419, opened 2023-09-01
Fix setting of invoice description to include hash

dangeross:
The invoice description hash is lost when building a new invoice to include route hints


roeierez (approved):
LGTM!

ok300 (approved):

PR #422, opened 2023-09-03
add Windows x86 target to C# nuget

JssDWt:
I think this is our first 32-bit version of the SDK, and there's no easy way to test whether it works in CI. So maybe a bit of caution.


roeierez (approved):

dangeross:
Two of the four android bindings are 32-bit also, but I've not tested them explicitly

PR #423, opened 2023-09-04
introduce init_logging function.

roeierez:
Decouple the global logging system from the connect method.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1064,20 +1064,26 @@ impl BreezServices {
     /// An error is thrown if the log file cannot be created in the working directory.
     ///
     /// An error is thrown if a global logger is already configured.
-    pub fn init_logging(log_dir: &str, app_logger: Option<Box<dyn log::Log>>) -> SdkResult<()> {
-        let target_log_file = Box::new(
-            OpenOptions::new()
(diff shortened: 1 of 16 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1064,20 +1064,26 @@ impl BreezServices {
...
-                .open(format!("{log_dir}/sdk.log"))
-                .map_err(|_| SdkError::InitFailed {
-                    err: "Can't create log file".into(),
-                })?,
-        );
-        let logger = env_logger::Builder::new()
-            .target(env_logger::Target::Pipe(target_log_file))
-            .parse_filters(
-                r#"
+    pub fn init_logging(
```
ok300, line 1067: NIT: The description for this has to be updated, but otherwise looks good.

ok300 (approved):

libs/sdk-core/src/breez_services.rs
ok300: My mistake, above comment is fine.

I meant the one in uniffi has to be updated: https://github.com/breez/breez-sdk/blob/c875ce74a123da5caf4c58a5383978ac919f30a5/libs/sdk-bindings/src/uniffi_binding.rs#L61-L62

PR #429, opened 2023-09-06
Generate React Native code

dangeross:
This PR adds a makefile task to generate the RN kotlin, swift & typescript code based on changes to the uniffi udl file

Depends on:
1. breez/breez-sdk-rn-generator#4
2. breez/breez-sdk-rn-generator#5
3. breez/breez-sdk-rn-generator#1
4. Tagging a release of breez-sdk-rn-generator


roeierez (approved):
LGTM!

JssDWt:
If we can generate the react native code, can we (at some point) remove the react native code from source control? And generate the code in CI when publishing a package?

dangeross:
> If we can generate the react native code, can we (at some point) remove the react native code from source control? And generate the code in CI when publishing a package?

It's only really the bridging code that is generated not the whole package structure, it's a lot like libs/sdk-flutter in that respect. An option could be to use the breez-sdk-react-native repo for package structure and use the CI to generate the code into there (kind of like the breez-sdk-go repo). It might create pre-release testing challenges though.

JssDWt:
> It's only really the bridging code that is generated not the whole package structure, it's a lot like libs/sdk-flutter in that respect. An option could be to use the breez-sdk-react-native repo for package structure and use the CI to generate the code into there (kind of like the breez-sdk-go repo). It might create pre-release testing challenges though.

For the prerelease testing challenges we could do a few things:
- archive the entire generated package and test manually. Run the CI again with publish=true if it works.
- When creating the package, commit but don't push. Then run an automated test, if it succeeds, push.

PR #432, opened 2023-09-07
Fix flutter release

JssDWt:
This PR builds sdk-core jniLibs for the flutter release instead of the sdk-bindings jniLibs in CI.
Also, it now archives the package. I tested the archived package and it works on an Android emulator.


dangeross (approved):
LGTM

PR #435, opened 2023-09-07
ci: remove rust cache for platform builds

JssDWt:
When compiling for specific targets, the rust cache is almost always a miss.
This causes the rust cache to only cost more time for a release, because it takes quite a long time to upload the cache.
Plus, these caches take space, which is stolen from the main CI caches.

So remove the rust caches for platform-specific builds.


ok300 (approved):

JssDWt:
Closing this, because the cache is used for subsequent builds.

PR #436, opened 2023-09-07
Return invoice on successful lnurl-withdraw

### LINKED ISSUE #427: Feature request: return payment hash on a successful LNURL withdrawal
It would be useful if `LnUrlCallbackStatus::Ok` included the payment hash of the invoice sent to the LNURL-w service. Otherwise, unless I'm missing something, it's hard for a consumer of the SDK to know exactly which payment is associated with this withdrawal.

ok300:
Include the `LNInvoice` as inner data in the `Ok` result of `BreezServices::lnurl_withdraw`.

Subtasks:
- [x] uniffi
- [x] flutter bridge
- [x] React Native bindings
- [x] Store optional field in `list_payments`
- [x] Add c-breez PR: https://github.com/breez/c-breez/pull/630

Fixes #427


ok300:
@roeierez it's ready for review.

I will add the RN bindings as last step, once we agree on the general structure.

dangeross:
> I will add the RN bindings as last step, once we agree on the general structure.

It could be the first use of the RN codegen 😎

ok300:
The PR simply adds the invoice in the withdraw `Ok` response.

However, if the goal is to be able to link a payment with a LNURL-withdraw request, an alternative way to do this is to extend `LnPaymentDetails`, which is a field of `Payment`

https://github.com/breez/breez-sdk/blob/41b3a01321b79ffc576755fa5ec24c80b959d5a0/libs/sdk-core/src/models.rs#L661-L678

with an `lnurl_withdraw_endpoint: Optional<String>`, where we can store the LNURL-w endpoint associated with this `Payment`.

That may be preferable to the approach in this PR, because in the PR the LNURL handling is inconsistent:
- `lnurl_pay` result has no reference to the invoice, but `LnPaymentDetails` has a few pay-related optional fields: https://github.com/breez/breez-sdk/blob/41b3a01321b79ffc576755fa5ec24c80b959d5a0/libs/sdk-core/src/models.rs#L669-L677
- `lnurl_withdraw` has the reverse: invoice in the service call result, but no reference in the `Payment`

What do you think? Is the PR approach ok? Or better switch to the above? Or do both?

danielgranhao:
If possible, as a user, I would like to have both :) that way, I can avoid always searching for the correct payment after making an LNURL withdrawal.

Otherwise, the withdrawal flow is: 
1.  Start withdrawal
2. Search for the created payment associated with the withdrawal
3. Wait for `BreezEvent::InvoicePaid` that includes the hash of the withdrawal payment

If we have what the current version of this PR proposes, the middle step can be discarded.

ok300:
@dangeross I took your RN generator for a test-drive and ran it on this PR.

Can you please doublecheck if all looks good? fe8884de178d4135b64f13f1a80c8224c079ccb5

dangeross (approved):
Looks good in general. There seems to be lots of linting changes with the kotlin files though. Can you give me your ktlint and kotlin versions?
```
ktlint --version
kotlin -version
```

ok300:
> ktlint --version

1.0.0

> kotlin -version

Kotlin version 1.9.10-release-459 (JRE 20.0.2)

dangeross:
> > ktlint --version
>
> 1.0.0

Ok, seems to be a version difference. I've upgraded to 1.0.0

roeierez:
@ok300 the PR looks good to me. As for your comment above do you want to add the step of associated the withdraw endpoint with the payment in the storage or not?  Another option is in a separate PR.

roeierez (approved):
LGTM! We need a corresponding c-breez PR as I believe this breaks it.

ok300:
> If possible, as a user, I would like to have both :)

Then I'll add both in this PR.

libs/sdk-core/src/models.rs
```diff
@@ -1114,6 +1117,18 @@ pub enum LnUrlCallbackStatus {
     },
 }
 
+/// [LnUrlCallbackStatus] specific to LNURL-withdraw, where the success case contains the invoice.
+#[derive(Serialize)]
+pub enum LnUrlWithdrawCallbackStatus {
```
roeierez, line 1122: Perhaps if we are introducing new types here we can use better naming?
`LnUrlWithdrawCallbackStatus`  => `LnUrlWithdrawResult`
`LnUrlWithdrawOkData`  => `LnurlWithdrawSuccess`
`LnUrlErrorData`  => `LnurlError`

roeierez (changes requested):
Looks good! Dropped a NIT comment.

erdemyerebasmaz:
@ok300 @roeierez 
Shouldn't we apply these naming changes to LnUrlCallbackStatus used on lnurl_auth as well?

ok300:
@erdemyerebasmaz how about `LnUrlCallbackStatus` => `LnUrlCallbackResult` ?

It's used not only in LNURL-auth, but also as intermediary step in LNURL-withdraw.

erdemyerebasmaz:
> @erdemyerebasmaz how about `LnUrlCallbackStatus` => `LnUrlCallbackResult` ?
> 
> It's used not only in LNURL-auth, but also as intermediary step in LNURL-withdraw.

@ok300 I see, isn't it supposed to be `LnUrlResult` according to:

> Perhaps if we are introducing new types here we can use better naming? 
`LnUrlWithdrawCallbackStatus` => `LnUrlWithdrawResult` 
`LnUrlWithdrawOkData` => `LnurlWithdrawSuccess` 
`LnUrlErrorData` => `LnurlError`

erdemyerebasmaz:
> Well, there are 2 naming conventions which don't overlap:
> 
> * one is for the LNURL-withdraw specific case. Calling the struct `LnUrlWithdrawResult` makes sense.
> * one is for a generic case that affects several callback types. `LnUrlResult` gives the wrong impression that its what a LNURL-workflow is supposed to end in, but it's only the case for auth. It's an intermediary result in withdraw. IMO `LnUrlCallbackResult` better captures the nature if it.

Thanks for the clarification, second item was what made it confusing for me. If that's the case, I see no issue with keeping `LnUrlCallbackStatus` as it is.

ok300:
Then @roeierez  this PR is ready for final review.

I also linked the c-breez PR by @erdemyerebasmaz in the description above.

dangeross (approved):
Looks good!

roeierez (approved):
LGTM

PR #441, opened 2023-09-07
Fix disconnect() mapping in bindings.rs

### LINKED ISSUE #439: Disconnect function not working
**Environment**

- Flutter app with breez-sdk-flutter
- Android emulator

**What**
After a successful `initialize` and `connect`, calling `disconnect` throws the following error:

> Background processing is not running

ok300:
Fixes #439

### DIFF
--- a/libs/sdk-core/src/binding.rs
+++ b/libs/sdk-core/src/binding.rs
@@ -17,7 +17,7 @@ use anyhow::{anyhow, Result};
 use flutter_rust_bridge::StreamSink;
 use log::{Level, LevelFilter, Metadata, Record};
 use once_cell::sync::{Lazy, OnceCell};
-use tokio::sync::mpsc;
+use tokio::sync::Mutex;
 
 use crate::breez_services::{self, BreezEvent, BreezServices, EventListener};
 use crate::chain::RecommendedFees;
@@ -38,8 +38,12 @@ use crate::{
     StaticBackupResponse,
 };
 
-static BREEZ_SERVICES_INSTANCE: OnceCell<Arc<BreezServices>> = OnceCell::new();
-static BREEZ_SERVICES_SHUTDOWN: OnceCell<mpsc::Sender<()>> = OnceCell::new();
+/*
+The format Lazy<Mutex<Option<...>>> for the following variables allows them to be instance-global,
+meaning they can be set only once per instance, but calling disconnect() will unset them.
+ */
+static BREEZ_SERVICES_INSTANCE: Lazy<Mutex<Option<Arc<BreezServices>>>> =
+    Lazy::new(|| Mutex::new(None));
 static NOTIFICATION_STREAM: OnceCell<StreamSink<BreezEvent>> = OnceCell::new();
 static RT: Lazy<tokio::runtime::Runtime> = Lazy::new(|| tokio::runtime::Runtime::new().unwrap());
 static LOG_INIT: OnceCell<bool> = OnceCell::new();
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/binding.rs
```diff
@@ -83,13 +81,7 @@ pub fn node_info() -> Result<NodeState> {
...
 /// Cleanup node resources and stop the signer.
 pub fn disconnect() -> Result<()> {
-    block_on(async {
-        let shutdown_handler = BREEZ_SERVICES_SHUTDOWN.get();
-        match shutdown_handler {
-            None => Err(anyhow!("Background processing is not running")),
-            Some(s) => s.send(()).await.map_err(anyhow::Error::msg),
-        }
-    })
+    block_on(async { get_breez_services()?.disconnect().await })
```
JssDWt, line 84: I think the `BREEZ_SERVICES_INSTANCE` has to be reset in order to be able to connect again later.

roeierez: I agree. OneCell only let you do a one time initialization so we might need to use a different mechanism here to hold the single breez services instance.

ok300: The simplest thing I found that looks promising is
```rust
static BREEZ_SERVICES_INSTANCE: Lazy<Arc<Mutex<Option<Arc<BreezServices>>>>> =
    Lazy::new(|| Arc::new(Mutex::new(None)));
```

but its horrible.

We want it mutable, but also possibly empty. Wrapped in an `Arc` for sure. And holding a pointer (`Arc`) to the `BreezServices`, not the instance itself.

I'll keep looking, but if someone has a better idea, let me know.

roeierez: Why not this?

`static BREEZ_SERVICES_INSTANCE: Mutex<Option<Arc<BreezServices>>> = Mutex::new(None);`

ok300: Good idea, tried it: 99f4aae90e87b93bb6917020122d8973dc07a98e

Had to wrap it in a `Lazy<>` though.

If this looks good, I'll do it for the other globals too.

libs/sdk-core/src/binding.rs
```diff
@@ -415,12 +421,13 @@ impl log::Log for BindingLogger {
...
 }
 
-fn get_breez_services() -> Result<&'static BreezServices> {
-    let n = BREEZ_SERVICES_INSTANCE.get();
-    match n {
-        Some(a) => Ok(a),
-        None => Err(anyhow!("Node service was not initialized")),
-    }
+fn get_breez_services() -> Result<Arc<BreezServices>> {
+    block_on(async move {
```
roeierez, line 425: Isn't using block_on is an issue here given the fact that all other methods that use `get_breez_services` are already being invoked inside a blocking code?

libs/sdk-core/src/binding.rs
```diff
@@ -48,16 +48,19 @@ static LOG_INIT: OnceCell<bool> = OnceCell::new();
...
-            .map_err(|_| SdkError::InitFailed {
-                err: "static node services already set".into(),
-            })?;
-
-        Ok(())
+        let mut locked = BREEZ_SERVICES_INSTANCE.lock().await;
+        match *locked {
+            None => {
+                let breez_services =
+                    BreezServices::connect(config, seed, Box::new(BindingEventListener {})).await?;
```
roeierez, line 55: The connect is being called after acquiring the lock but disconnect is not. Better to be consistent here.

roeierez (changes requested):

libs/sdk-core/src/binding.rs
ok300: Done

libs/sdk-core/src/binding.rs
ok300: Done. I changed `disconnect` to also first acquire the lock.

roeierez:
> I changed the remaining global variables to the new `Lazy<Mutex<Option<...>>>` type. They are also all unset on `disconnect()`.
> 
> The PR is ready for final review.
> 
> I generally think the use of `disconnect()` and allowing a new `connect()` on the same instance is a fragile setup. Main reason is that we cannot control how the app calls (or doesn't call) `init_logging` before `connect()`. Since `init_logging()` sets a global logger, it cannot be changed when `connect()` is called again, presumably with a different set of credentials. For example, it would not be possible to keep per-wallet logs in separate directories.
> 
> @roeierez what are your thoughts on this? What would speak against deprecating `disconnect()` and instead asking that, on each new `connect()` a different instance is used?

We would gladly use the same pattern that we use uniffi which means allowing multiple instances but unfortunately dart doesn't have yet uniffi support and the fluter rust bridge that we are using doesn't support object references (at least as for the last time I checked).

libs/sdk-core/src/binding.rs
```diff
@@ -84,22 +93,29 @@ pub fn node_info() -> Result<NodeState> {
...
+        let mut locked_sdk_instance = BREEZ_SERVICES_INSTANCE.lock().await;
+        *locked_sdk_instance = None;
+
+        let mut locked_notifications_stream = NOTIFICATION_STREAM.lock().await;
+        *locked_notifications_stream = None;
+
+        let mut locked_log_init = LOG_INIT.lock().await;
+        *locked_log_init = None;
+
+        get_breez_services().await?.disconnect().await?;
```
roeierez, line 105: Does this work? I am pretty sure there is a deadlock here as the get_breez_services() aquiers the lock again...
Also why do we need to involve the NOTIFICATION_STREAM and the LOG_INIT ? can't these global handlers stay with OneCell?

roeierez (changes requested):

libs/sdk-core/src/binding.rs
ok300: Good point. I shifted the `disconnect` above the locking, so they don't interfere.

> Also why do we need to involve the NOTIFICATION_STREAM and the LOG_INIT 

True. i thought they are set again after `disconnect`, but they cannot because `init_logging` sets the global logger, so it cannot be called again. I'll revert them.

ok300: Done in 54b3b4810c5ffc04ef152f949b36c31f30b6670f

roeierez (approved):

PR #442, opened 2023-09-07
Bump mockito dependency

### LINKED ISSUE #147: Update mockito to v1.0.0
Here's the migration guide:
  - [Migrating to the new API](https://github.com/lipanski/mockito/releases/tag/1.0.0)

ok300:
Fixes #147


libs/sdk-core/src/input_parser.rs
```diff
@@ -585,17 +585,24 @@ impl From<Uri<'_>> for BitcoinAddressData {
...
+    use mockito::{Mock, Server, ServerGuard};
+    use once_cell::sync::Lazy;
 
     use crate::input_parser::*;
     use crate::models::Network;
 
+    /// Mock server used in tests
+    pub(crate) static MOCK_HTTP_SERVER: Lazy<Mutex<ServerGuard>> =
+        Lazy::new(|| Mutex::new(Server::new()));
+
```
JssDWt, line 605: I don't think this needs to be shared between tests. `let mut server = mockito::Server::new();` in the tests is sufficient and avoids shared state.

tools/sdk-cli/src/main.rs
```diff
@@ -5,32 +5,38 @@ mod commands;
...
 
     let persistence = CliPersistence { data_dir };
     let history_file = &persistence.history_file();
 
-    let rl = &mut Editor::<()>::new().unwrap();
+    let rl = &mut Editor::new()?;
+    rl.set_helper(Some(CliHelper {
+        hinter: HistoryHinter {},
+        colored_prompt: format!("\x1b[1;32m{}\x1b[0m", "sdk> "),
+    }));
```
JssDWt, line 39: What does this do?

JssDWt (commented):
Looks good. The gl-client upgrade may not be needed anymore?

libs/sdk-core/src/input_parser.rs
ok300: That's what I started with, but the problem is mockito chooses a random port on start. We later need to know this port to replace "localhost" in test values with `localhost:port`. Not the best approach, I know, but I found no other way.

So this `static MOCK_HTTP_SERVER` is shared among tests to go around this issue with random ports.

tools/sdk-cli/src/main.rs
ok300: Well, it colors the prompt. It's a sort of "being color green" / "end color" type of char sequence for terminals.

Yea its kind of weird. I can add a comment to clarify or just remove it, let me know.

ok300: It's one thing the new version of `rustyline` supports, so I figured we could use it.

libs/sdk-core/src/input_parser.rs
JssDWt: Ah it actually runs a real server? I didn't realize that.

tools/sdk-cli/src/main.rs
JssDWt: OK, it seems the color is added again in the highlighter? https://github.com/breez/breez-sdk/pull/442/files#diff-7e62d977711c2b08170e3927acd4e04382b392b03dc4326a5e6bdaa3c646e6daR77

ok300: That's separate, it highlights the autocompleted text in a different shade of white.

Based on the CLI history, e.g. if you called `node_info` before, as you start typing `node` it will autofill `_info` with the different highlight. Pressing right arrow auto-completes the command.

PR #443, opened 2023-09-08
Bump dependencies

ok300:
Update dependency versions where possible.


ok300:
The last commit replaces wildcard versions in our `Cargo.toml` with specific ones, which fixes #447.

roeierez (approved):
LTGM!

PR #444, opened 2023-09-10
Anchor signer

roeierez:
This PR upgrades the greenlight dependency to the one that supports zero fee htlc: https://github.com/Blockstream/greenlight/pull/237

Since there was a breaking change there that changes the `bip32_ext_key` value we also needed to fallback to the old key in some cases. We will need to stay with this code for some reasonable time.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -374,6 +374,7 @@ impl BreezServices {
         &self,
         req_data: ReceivePaymentRequest,
     ) -> SdkResult<ReceivePaymentResponse> {
+        info!("receive payment started1");
         self.payment_receiver.receive_payment(req_data).await
     }
 
@@ -1117,7 +1118,7 @@ impl BreezServices {
             .target(env_logger::Target::Pipe(target_log_file))
             .parse_filters(
                 r#"
-                info,
+                debug,
                 gl_client=warn,
                 h2=warn,
                 hyper=warn,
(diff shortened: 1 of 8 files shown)


ok300:
Starting when (date, commit, etc) did GL change the `bip32_ext_key` implementation?

I'm planning to restore a backup from before that date, using this code (when PR is ready), to test if the fallback works.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1117,7 +1117,7 @@ impl BreezServices {
             .target(env_logger::Target::Pipe(target_log_file))
             .parse_filters(
                 r#"
-                info,
+                debug,
```
JssDWt, line 1120: don't forget to undo this

JssDWt (approved):
It looks good. As long as the upgraded client is the only client. Should there be any legacy clients too for the same node, this is a problem, because then we're encrypting the backup with an unknown key to that client.

roeierez:
> It looks good (assuming it's tested). As long as the upgraded client is the only client. Should there be any legacy clients too for the same node, this is a problem, because then we're encrypting the backup with an unknown key to that client.

Right but currently it won't run side by side even because of the signer incompetability.

dangeross (approved):
LGTM! 

Were you able to test if an anchor channel was created?

roeierez:
> LGTM!
> 
> Were you able to test if an anchor channel was created?

Yes, was created.

PR #445, opened 2023-09-10
Baseline RN code generation

dangeross:
First code generation from #429


roeierez (approved):
So happy to see this! 🥇

dangeross:
@roeierez I have successfully tested the generated code in the Satimoto app. It took a bit of code rework to handle the updated type structure and the inclusion of type variants. Specific parts affected are lightning payment (including payment events) handling, LNURL result types and the interfaces to setLogStream (renamed from startLogStream) and connect (adds event handler as third param)

I will rebase, regenerate the baseline and merge

Issue #437, opened 2023-09-07
Add SCB to the SDK


kingonly:
* Fetch is on channel creation and store it locally
* Expose API to retrieve it
* Document how to use it from a cln node

ok300:
Docs are ready at https://sdk-doc.breez.technology/guide/static_channel_backup.html

PR #448, opened 2023-09-11
Expose static backup data

roeierez:
This PR exposed the static backup file from the node.
The goal is to persist the up to date static backup file locally for a case where the node is not available to serve that file.
Still missing the React Native (Hope to use the generator for that).

fixed https://github.com/breez/breez-sdk/issues/437

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -312,6 +312,14 @@ dictionary ReceivePaymentResponse {
     u64? opening_fee_msat;
 };
 
+dictionary StaticBackupRequest {
+    string working_dir;
+};
+
+dictionary StaticBackupResponse {
+    sequence<string>? backup;
+};
+
 dictionary ReceiveOnchainRequest {
     OpeningFeeParams? opening_fee_params=null;
 };
@@ -614,4 +622,7 @@ namespace breez_sdk {
  sequence<u8> mnemonic_to_seed(string phrase);
 
  Config default_config(EnvironmentType env_type, string api_key, NodeConfig node_config);
+
+ [Throws=SdkError]
+ StaticBackupResponse static_backup(StaticBackupRequest request);
 };
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -717,7 +717,17 @@ impl BreezServices {
 
         // update node state and channels state
         self.persister.set_node_state(&new_data.node_state)?;
+
+        let channels_before_update = self.persister.list_channels()?;
         self.persister.update_channels(&new_data.channels)?;
+        let channels_after_update = self.persister.list_channels()?;
+
+        // Fetch the static backup if needed and persist it
+        if channels_before_update.len() != channels_after_update.len() {
+            info!("fetching static backup file from node");
+            let backup = self.node_api.static_backup().await?;
+            self.persister.set_static_backup(backup)?;
+        }
 
         //fetch closed_channel and convert them to Payment items.
         let closed_channel_payments_res: Result<Vec<Payment>> = self
@@ -840,6 +850,16 @@ impl BreezServices {
             EnvironmentType::Production => Config::production(api_key, node_config),
             EnvironmentType::Staging => Config::staging(api_key, node_config),
         }
+    }
+
+    /// Get the static backup data from the peristent storage.
+    /// This data enables the user to recover the node in an external core ligntning node.
+    /// See here for instructions on how to recover using this data: https://docs.corelightning.org/docs/backup-and-recovery#backing-up-using-static-channel-backup
+    pub fn static_backup(request: StaticBackupRequest) -> SdkResult<StaticBackupResponse> {
+        let storage = SqliteStorage::new(request.working_dir);
+        Ok(StaticBackupResponse {
+            backup: storage.get_static_backup()?,
+        })
     }
 
     /// Generates an url that can be used by a third part provider to buy Bitcoin with fiat currency.
--- a/libs/sdk-react-native/ios/BreezSDKMapper.swift
+++ b/libs/sdk-react-native/ios/BreezSDKMapper.swift
@@ -23,7 +23,8 @@ class BreezSDKMapper {
         var list = [AesSuccessActionDataDecrypted]()
         for value in arr {
             if let val = value as? [String: Any?] {
-                list.append(try asAesSuccessActionDataDecrypted(data: val))
+                var aesSuccessActionDataDecrypted = try asAesSuccessActionDataDecrypted(data: val)
+                list.append(aesSuccessActionDataDecrypted)
             } else {
                 throw SdkError.Generic(message: "Invalid element type AesSuccessActionDataDecrypted")
             }
@@ -52,7 +53,8 @@ class BreezSDKMapper {
         var list = [BackupFailedData]()
         for value in arr {
             if let val = value as? [String: Any?] {
-                list.append(try asBackupFailedData(data: val))
+                var backupFailedData = try asBackupFailedData(data: val)
+                list.append(backupFailedData)
             } else {
                 throw SdkError.Generic(message: "Invalid element type BackupFailedData")
             }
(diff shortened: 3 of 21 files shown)


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -613,4 +621,7 @@ namespace breez_sdk {
  sequence<u8> mnemonic_to_seed(string phrase);
 
  Config default_config(EnvironmentType env_type, string api_key, NodeConfig node_config);
+
+ [Throws=SdkError]
+ StaticBackupResponse static_backup(StaticBackupRequest request);
```
dangeross, line 626: Is there a reason static_backup is exposed as a "static" method of the SDK instead of a service instance method?
These are my two concerns:

1. I think our intention with handling multiple instance states is for the connect method to create/determine a subdirectory off of the base working directory derived from the node seed. With this the client does not need to manage which working_dir is used for which seed. A side effect of this is the client using the SDK will not know the working_dir of the node (only the base working directory). If this is a instance method, then the working_dir is not needed.
2. Although this does give access to the backup while the node is not connected, it could lead to the client using the SDK to download an out-of-date backup. If this is a instance method and the sync had occurred, then at least we know we are returning the latest static backup.

roeierez: The requirements for the whole feature is to be able to access the static backup data when the sdk is not able to connect to the node (sort of doom day when greenlight is down for example).
To meet this requirement we can't count on connect to succeed and this is the reason I chose a static API.
As for multiple nodes I think this can evolve to that. The working_dir (base directory) would still be needed but when we introduce multiple nodes under this need we might need to add another field to the StaticBackupRequest such as the seed or the node id.
As for the latest file, we do fetch it on every new channel (see the sync logic) so even if called statically it should be up to date.
LMK if you see any better option here.

dangeross: > The requirements for the whole feature is to be able to access the static backup data when the sdk is not able to connect to the node (sort of doom day when greenlight is down for example).
> To meet this requirement we can't count on connect to succeed and this is the reason I chose a static API.
> As for multiple nodes I think this can evolve to that. The working_dir (base directory) would still be needed but when we introduce multiple nodes under this need we might need to add another field to the StaticBackupRequest such as the seed or the node id.

👍 
> As for the latest file, we do fetch it on every new channel (see the sync logic) so even if called statically it should be up to date.

I was just thinking of a scenario where multiple clients/app connect to the same node. If Client A opens a new channel, the static backup of Client A is synced. Any other clients have stale static backups until they connect and sync. These stale backups could potentially be used to recover from. I guess there is not much you can do to prevent this, if its the dooms day scenario.

libs/sdk-core/src/breez_services.rs
```diff
@@ -704,6 +704,15 @@ impl BreezServices {
         self.persister.set_node_state(&new_data.node_state)?;
         self.persister.update_channels(&new_data.channels)?;
 
+        // Fetch the static backup if needed and persist it
+        let static_backup = self.persister.get_static_backup()?;
+        let backed_up_channels = static_backup.unwrap_or(vec![]);
+        if backed_up_channels.len() != new_data.channels.len() {
```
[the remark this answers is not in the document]
roeierez: > I'm not too sure about this check. If one channel opens and one channel closes (on another client), you will never fetch the new backup. Is there a way to get the state of the backup (aka, which channels were backed up) more reliably? Maybe we can check our persisted channels and check whether any state has changed between the current sync?

My assumption was that the static backup data should have one entry per channel for all channels including closed ones (At least this is also the behavior I saw when testing). This code compares that array length to the list of all channels (closed and opened) after sync so it should be up to date.
I now checked in LND and saw that the closed channels are _**removed**_ from the static backup file and if this is the case here then yes, we should change the fetch trigger. Good point.



> These are the fields of the scb in cln: Note the `unused` field. Looks like the backup also changes if the channel was used?

According to the docs here we should fetch this data whenever a new channel is opened as this is the point where the file is changed: https://docs.corelightning.org/docs/backup-and-recovery#static-channel-backup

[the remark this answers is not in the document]
roeierez: @JssDWt  I changed the check to compare the persistent channels length to the new fetched channels length.

libs/sdk-bindings/src/breez_sdk.udl
roeierez: I agree there is no 100% solution here especially if we have multiple clients. In this case there are highly chances that at list one of them will have the most up to static backup data.
Even with one instance there could be a race where the sync failed or some network error happened.
I think we can say this thing for sure. As long as the sync API call succeed then the static data should be up to date.

libs/sdk-core/src/breez_services.rs
```diff
@@ -702,7 +702,17 @@ impl BreezServices {
 
         // update node state and channels state
         self.persister.set_node_state(&new_data.node_state)?;
+
+        let channels_before_update = self.persister.list_channels()?;
         self.persister.update_channels(&new_data.channels)?;
+        let channels_after_update = self.persister.list_channels()?;
+
+        // Fetch the static backup if needed and persist it
+        if channels_before_update.len() != channels_after_update.len() {
```
dangeross, line 711: If a channel is closed the static backup is not updated here, right?

Maybe comparing a hash or a HashSet of before/after channels list might work.

roeierez: That's right if a channel is closed the static backup is not fetched . This is the expected behavior as we should only fetch it when a new channel is opened, see docs here: https://docs.corelightning.org/docs/backup-and-recovery#static-channel-backup

dangeross: Ok, I thought the static backup only contains active channels

dangeross (approved):
LGTM!

dangeross:
You can run the RN codegen on this now :wink:

JssDWt (approved):
LGTM

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -23,7 +23,7 @@ class BreezSDKMapper {
         var list = [AesSuccessActionDataDecrypted]()
         for value in arr {
             if let val = value as? [String: Any?] {
-                try list.append(asAesSuccessActionDataDecrypted(data: val))
+                list.append(try asAesSuccessActionDataDecrypted(data: val))
```
dangeross, line 26: Seems like the linter didn't run? But in any case, I'll correct the generator to position the `try` correctly

dangeross: I've added breez/breez-sdk-rn-generator#6 that should resolve the try hoisting issue with swift linting

dangeross (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -842,6 +852,16 @@ impl BreezServices {
         }
     }
 
+    /// Get the static backup data from the peristent storage.
+    /// This data enables the user to recover the node in an external core ligntning node.
+    /// See here for instructions on how to recover using this data: https://docs.corelightning.org/docs/backup-and-recovery#backing-up-using-static-channel-backup
+    pub fn static_backup(request: StaticBackupRequest) -> SdkResult<StaticBackupResponse> {
+        let storage = SqliteStorage::new(request.working_dir);
```
erdemyerebasmaz, line 859: I'm a bit late to the conversation but, isn't this `working_dir` almost always the same with [SDK Config's working_dir](https://github.com/breez/breez-sdk/blob/a441154a106eb7c4dd70a49453bd73c98f94edaa/libs/sdk-core/src/models.rs#L468)?

erdemyerebasmaz (commented):
@roeierez @JssDWt @dangeross 
I have a question about `working_dir` of `StaticBackupRequest`

libs/sdk-core/src/breez_services.rs
ok300: Yes, but from what I understand, the SDK config is not available here to provide the `working_dir` because this is a static method, not a service method called on an instance of `BreezServices`.

See here for more background: https://github.com/breez/breez-sdk/pull/448#discussion_r1322070122

PR #449, opened 2023-09-11
Use the channel close tx blocktime as fallback close ts

ok300:
For older DBs, where the channel `closed_at` was not part of the schema, we were defaulting to the epoch time.

This PR changes that, such that now the channel close time is taken from the closing tx block time.

Tested with a pre-December-2022 wallet.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -730,16 +730,18 @@ impl BreezServices {
         }
 
         //fetch closed_channel and convert them to Payment items.
-        let closed_channel_payments_res: Result<Vec<Payment>> = self
-            .persister
-            .list_channels()?
-            .into_iter()
-            .filter(|c| c.state == ChannelState::Closed || c.state == ChannelState::PendingClose)
-            .map(closed_channel_to_transaction)
-            .collect();
+        let mut closed_channel_payments: Vec<Payment> = vec![];
+        for closed_channel in
+            self.persister.list_channels()?.into_iter().filter(|c| {
+                c.state == ChannelState::Closed || c.state == ChannelState::PendingClose
+            })
+        {
+            let closed_channel_tx = self.closed_channel_to_transaction(closed_channel).await?;
+            closed_channel_payments.push(closed_channel_tx);
+        }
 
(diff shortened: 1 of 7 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1154,6 +1156,57 @@ impl BreezServices {
...
+        let channel_closed_at = channel.closed_at.unwrap_or(
+            {
+                // Lookup tx based on channel funding tx ID
+                let outspends = self
+                    .chain_service
+                    .transaction_outspends(channel.funding_txid.clone())
+                    .await?;
+
+                // If outputs are spent, take the block time as the channel closing timestamp
+                match outspends.iter().find(|&out| out.spent) {
```
roeierez, line 1177: The funding tx may have other output. We need to find the one that funded the channel and here we only take the first one.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
roeierez: Maybe we can select the output that matches the channel capacity?

ok300: Finding the output based on `funding_outnum`, as discussed.

libs/sdk-core/src/chain.rs
```diff
@@ -11,6 +11,15 @@ pub trait ChainService: Send + Sync {
     /// See <https://mempool.space/docs/api/rest#get-address-transactions>
     async fn address_transactions(&self, address: String) -> Result<Vec<OnchainTx>>;
     async fn current_tip(&self) -> Result<u32>;
+    /// Gets a transaction.
+    ///
+    /// See <https://mempool.space/docs/api/rest#get-transaction>
+    async fn transaction(&self, txid: String) -> Result<OnchainTx>;
```
roeierez, line 17: This seems unused?

roeierez (changes requested):
LGTM! just one issue with unnecessary mothod.

libs/sdk-core/src/chain.rs
ok300: Done, removed

PR #452, opened 2023-09-12
Check if send_payment invoice is already paid

### LINKED ISSUE #414: Inconsistent Invoice Payment Confirmation in sendPayment Method
In the `sdk-flutter` project, there seems to be an issue with the `sendPayment` method in the `BreezBridge`. 

**Expected Behavior:**
When attempting to pay an invoice for the first time, it should succeed, and subsequent attempts to pay the same invoice should result in a clear "Invoice Already Paid" exception.

**Actual Behavior:**
Currently, when paying an invoice for the second time or any subsequent attempts, the system still provides a success message. This is not the expected behavior and can be confusing to users.

**Steps to Reproduce:**
1. Create an invoice.
2. Attempt to pay the invoice for the first time (expect success).
3. Attempt to pay the same invoice again (expect an "Invoice Already Paid" exception).

**Additional Information:**
This bug affects the user experience, as users may believe they successfully paid the invoice multiple times when it should be prevented after the first payment. This issue should be addressed to ensure accurate feedback to users and proper handling of payment attempts.

dangeross:
Lookup the invoice payment by hash and if the payment is already completed, throw an error.

Fixes #414

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -228,14 +228,20 @@ impl BreezServices {
     ///
     /// * `bolt11` - The bolt11 invoice
     /// * `amount_sats` - The amount to pay in satoshis
-    pub async fn send_payment(&self, bolt11: String, amount_sats: Option<u64>) -> Result<Payment> {
+    pub async fn send_payment(
+        &self,
+        bolt11: String,
+        amount_sats: Option<u64>,
+    ) -> SdkResult<Payment> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(bolt11.as_str())?;
         match self
             .persister
             .get_completed_payment_by_hash(&parsed_invoice.payment_hash)?
         {
-            Some(_) => Err(anyhow!("Invoice already paid")),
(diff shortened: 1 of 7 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -230,16 +231,24 @@ impl BreezServices {
...
-            parsed_invoice.payee_pubkey.clone(),
-            Some(parsed_invoice),
-            payment_res,
-        )
-        .await
+        match self
+            .persister
+            .get_completed_payment_by_hash(&parsed_invoice.payment_hash)?
+        {
+            Some(_) => Err(anyhow!("Invoice already paid")),
```
roeierez, line 238: Perhaps we can introduce a new SdkError here?

roeierez (changes requested):
Other than a small enhancement looks good.

libs/sdk-core/src/breez_services.rs
dangeross: Done. Added SdkError and updated test

roeierez (approved):
LGTM!

PR #460, opened 2023-09-14
Add paging support for payment list

gcomte:


libs/sdk-core/src/breez_services.rs
```diff
@@ -451,9 +451,10 @@ impl BreezServices {
         filter: PaymentTypeFilter,
         from_timestamp: Option<i64>,
         to_timestamp: Option<i64>,
+        amount_of_payments: Option<u32>,
```
roeierez, line 454: Perhaps we can pass here two parameters? offset, limit
This will make it more generic and add full paging support.

libs/sdk-core/src/binding.rs
```diff
@@ -207,10 +207,11 @@ pub fn list_payments(
     filter: PaymentTypeFilter,
     from_timestamp: Option<i64>,
     to_timestamp: Option<i64>,
+    amount_of_payments: Option<u32>,
```
roeierez, line 210: This file should be generated using `make flutter_rust_bridge` in sdk-flutter directory

roeierez (changes requested):
Thank you very much! Paging support is indeed a good improvement.
Dropped a few comments.
There are also instructions to generate the ReactNative bindings in the Readme so please follow that as part of this PR. It is a simple step but needed.

libs/sdk-core/src/breez_services.rs
roeierez, line 454: In order to get the support not only in rust we need to change the udl file and `BlockingBreezServices` in the sdk-binding crate.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
gcomte: I added the field in [this commit](https://github.com/breez/breez-sdk/compare/4abb0b97f4e35971e4a0aa82c9711051ab3ceb53..1e3f7c6ef0b278a21206b1fab9d075525713cdf9) but then the parameter list got very long, so I decided to restructure it [here](https://github.com/breez/breez-sdk/pull/460/commits/6254b0a62b17fc7598cf13973da15f021ff9be33).

roeierez: This is the right decision! Please be aware that this was also changed in main branch to a struct.

PR #464, opened 2023-09-15
Add convenience tasks to check fmt and clippy

dangeross:


ok300:
`cargo fmt --all`  and `cargo clippy --all` will also cover tests.

dangeross:
`cargo clippy --all` does seem to cover tests for me

ok300:
Good catch, that should be `cargo clippy --tests`

ok300:
You can have a look how we do it in the CI:
* https://github.com/breez/breez-sdk/blob/main/.github/workflows/main.yml#L15-L25
* https://github.com/breez/breez-sdk/blob/main/.github/workflows/main.yml#L166-L173

dangeross:
> You can have a look how we do it in the CI:

This is exactly where I copied them from 😄 Just so I can to a quick pre-commit check

roeierez (approved):

PR #466, opened 2023-09-17
Send onchain the largest amount possible

roeierez:
This PR is to supports the use case of sending the maximum amount to an on chain address (send_onchain).
Since this operation is a reverse swap where the user node sends over lightning we need to calculate the shortest stable route to the swapper node and find the maximum amount to forward for each channel while making sure we have enough fees to pay over lightning.
There are two parts to achieve the above:

## Calculating the max amount to send to the swapper node
For this purpose one API method was added: `max_reverse_swap_amount` which returns the maximum amount to send to the reverse swapper. This is possible due to the following two assumption that the sdk has:
1. The reverse swapper node can provide us with the routing hints to its private node.
2. The path to the reverse swapper consists mostly 3 hops and has enough liquidity to route all funds.

## Sending the amount while draining the channels
For this purpose we needed to implement another way of sending (send_pay) in the NodeAPI trait where we can control the internal htlc splitting.
The `send_pay` (sdk method) behaves as follows:
1. For a given invoice finds the path to the destination node (payee)  that is no longer than 3 hops (including the routing hints).
2. Calculates the route (including fees and cltv delta expiries).
3. Using the greenlight `send_pay` to send each htlc (one per channel).

- [x] Implement `max_reverse_swap_payment`
- [x] Implement `send_onchain` using a predefined route, emptying all channels.

### DIFF
--- a/libs/sdk-core/src/binding.rs
+++ b/libs/sdk-core/src/binding.rs
@@ -34,13 +34,13 @@ use crate::models::{Config, LogEntry, NodeState, Payment, SwapInfo};
 use crate::{
     BackupStatus, BuyBitcoinRequest, BuyBitcoinResponse, CheckMessageRequest, CheckMessageResponse,
     EnvironmentType, ListPaymentsRequest, LnUrlCallbackStatus, LnUrlPayRequest,
-    LnUrlWithdrawRequest, LnUrlWithdrawResult, NodeConfig, OpenChannelFeeRequest,
-    OpenChannelFeeResponse, PrepareRefundRequest, PrepareRefundResponse, PrepareSweepRequest,
-    PrepareSweepResponse, ReceiveOnchainRequest, ReceivePaymentRequest, ReceivePaymentResponse,
-    RefundRequest, RefundResponse, ReverseSwapFeesRequest, ReverseSwapInfo, ReverseSwapPairInfo,
-    SendOnchainRequest, SendOnchainResponse, SendPaymentRequest, SendPaymentResponse,
-    SendSpontaneousPaymentRequest, SignMessageRequest, SignMessageResponse, StaticBackupRequest,
-    StaticBackupResponse, SweepRequest, SweepResponse,
+    LnUrlWithdrawRequest, LnUrlWithdrawResult, MaxReverseSwapAmountResponse, NodeConfig,
+    OpenChannelFeeRequest, OpenChannelFeeResponse, PrepareRefundRequest, PrepareRefundResponse,
+    PrepareSweepRequest, PrepareSweepResponse, ReceiveOnchainRequest, ReceivePaymentRequest,
+    ReceivePaymentResponse, RefundRequest, RefundResponse, ReverseSwapFeesRequest, ReverseSwapInfo,
+    ReverseSwapPairInfo, SendOnchainRequest, SendOnchainResponse, SendPaymentRequest,
+    SendPaymentResponse, SendSpontaneousPaymentRequest, SignMessageRequest, SignMessageResponse,
+    StaticBackupRequest, StaticBackupResponse, SweepRequest, SweepResponse,
 };
 
 /*
@@ -299,6 +299,11 @@ pub fn list_fiat_currencies() -> Result<Vec<FiatCurrency>> {
 }
 
 /*  On-Chain Swap API's */
+
+pub fn max_reverse_swap_amount() -> Result<MaxReverseSwapAmountResponse> {
+    block_on(async { get_breez_services().await?.max_reverse_swap_amount().await })
+        .map_err(anyhow::Error::new::<SdkError>)
+}
 
 /// See [BreezServices::send_onchain]
 pub fn send_onchain(req: SendOnchainRequest) -> Result<SendOnchainResponse> {
--- a/libs/sdk-core/src/binding.rs
+++ b/libs/sdk-core/src/binding.rs
@@ -300,6 +300,7 @@ pub fn list_fiat_currencies() -> Result<Vec<FiatCurrency>> {
 
 /*  On-Chain Swap API's */
 
+/// See [BreezServices::max_reverse_swap_amount]
 pub fn max_reverse_swap_amount() -> Result<MaxReverseSwapAmountResponse> {
     block_on(async { get_breez_services().await?.max_reverse_swap_amount().await })
         .map_err(anyhow::Error::new::<SdkError>)
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -701,17 +701,12 @@ impl BreezServices {
     pub async fn max_reverse_swap_amount(&self) -> SdkResult<MaxReverseSwapAmountResponse> {
         // fetch the last hop hints from the swapper
         let last_hop = self.btc_send_swapper.last_hop_for_payment().await?;
-        let lsp = self.lsp_info().await?;
-        let lsp_pubkey = hex::decode(lsp.pubkey).map_err(|e| SdkError::Generic {
-            err: format!("Failed to decode lsp pubkey: {e}"),
-        })?;
         // calculate the largest payment we can send over this route using maximum 3 hops
         // as follows:
         // User Node -> LSP Node -> Routing Node -> Swapper Node
         let max_to_pay = self
             .node_api
             .max_sendable_amount(
-                lsp_pubkey,
                 Some(
                     hex::decode(&last_hop.src_node_id).map_err(|e| SdkError::Generic {
                         err: format!("Failed to decode hex node_id: {e}"),
@@ -738,15 +733,7 @@ impl BreezServices {
             Use the in_progress_reverse_swaps method to get an overview of currently ongoing reverse swaps".into(), 
         });
 
-        let lsp_info = self.lsp_info().await?;
-        let lsp_pubkey = hex::decode(lsp_info.pubkey).map_err(|e| SendOnchainError::Generic {
-            err: format!("Failed to decode lsp pubkey: {e}"),
-        })?;
-
-        let full_rsi = self
-            .btc_send_swapper
-            .create_reverse_swap(req, lsp_pubkey)
-            .await?;
+        let full_rsi = self.btc_send_swapper.create_reverse_swap(req).await?;
         let rsi = self
             .btc_send_swapper
             .convert_reverse_swap_info(full_rsi)
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -707,7 +707,7 @@ impl BreezServices {
         // User Node -> LSP Node -> Routing Node -> Swapper Node
         let (max_to_pay, _) = self
             .node_api
-            .max_amount_to_send(
+            .max_sendable_amount(
                 Some(
                     hex::decode(&last_hop.src_node_id)
                         .map_err(|e| anyhow!("Failed to decode hex node_id: {e}"))?,
@@ -718,9 +718,9 @@ impl BreezServices {
             .await?;
 
         // Sum the max amount per channel and return the result
-        let total_msat = max_to_pay.into_iter().map(|m| m.amount_msat).sum();
-
-        Ok(MaxReverseSwapAmountResponse { total_msat })
+        let total_msat: u64 = max_to_pay.into_iter().map(|m| m.amount_msat).sum();
+        let total_sat = total_msat / 1000;
+        Ok(MaxReverseSwapAmountResponse { total_sat })
     }
 
     /// Creates a reverse swap and attempts to pay the HODL invoice
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -702,8 +702,9 @@ impl BreezServices {
         // fetch the last hop hints from the swapper
         let last_hop = self.btc_send_swapper.last_hop_for_payment().await?;
         let lsp = self.lsp_info().await?;
-        let lsp_pubkey =
-            hex::decode(lsp.pubkey).map_err(|e| anyhow!("Failed to decode lsp pubkey: {e}"))?;
+        let lsp_pubkey = hex::decode(lsp.pubkey).map_err(|e| SdkError::Generic {
+            err: format!("Failed to decode lsp pubkey: {e}"),
+        })?;
         // calculate the largest payment we can send over this route using maximum 3 hops
         // as follows:
         // User Node -> LSP Node -> Routing Node -> Swapper Node
@@ -712,8 +713,9 @@ impl BreezServices {
             .max_sendable_amount(
                 lsp_pubkey,
                 Some(
-                    hex::decode(&last_hop.src_node_id)
-                        .map_err(|e| anyhow!("Failed to decode hex node_id: {e}"))?,
+                    hex::decode(&last_hop.src_node_id).map_err(|e| SdkError::Generic {
+                        err: format!("Failed to decode hex node_id: {e}"),
+                    })?,
                 ),
                 swap_out::reverseswap::MAX_PAYMENT_PATH_HOPS,
                 Some(&last_hop),
@@ -737,8 +739,9 @@ impl BreezServices {
         });
 
         let lsp_info = self.lsp_info().await?;
-        let lsp_pubkey = hex::decode(lsp_info.pubkey)
-            .map_err(|e| anyhow!("Failed to decode lsp pubkey: {e}"))?;
+        let lsp_pubkey = hex::decode(lsp_info.pubkey).map_err(|e| SendOnchainError::Generic {
+            err: format!("Failed to decode lsp pubkey: {e}"),
+        })?;
 
         let full_rsi = self
             .btc_send_swapper
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -696,7 +696,7 @@ impl BreezServices {
     /// Returns the max amount that can be sent on-chain using the send_onchain method.
     /// The returned amount is the sum of the max amount that can be sent on each channel
     /// minus the expected fees.
-    /// This is possible since the route to the swapper node is known in advance ans is expected
+    /// This is possible since the route to the swapper node is known in advance and is expected
     /// to consist of maximum 3 hops.
     pub async fn max_reverse_swap_amount(&self) -> SdkResult<MaxReverseSwapAmountResponse> {
         // fetch the last hop hints from the swapper
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -701,13 +701,16 @@ impl BreezServices {
     pub async fn max_reverse_swap_amount(&self) -> SdkResult<MaxReverseSwapAmountResponse> {
         // fetch the last hop hints from the swapper
         let last_hop = self.btc_send_swapper.last_hop_for_payment().await?;
-
+        let lsp = self.lsp_info().await?;
+        let lsp_pubkey =
+            hex::decode(lsp.pubkey).map_err(|e| anyhow!("Failed to decode lsp pubkey: {e}"))?;
         // calculate the largest payment we can send over this route using maximum 3 hops
         // as follows:
         // User Node -> LSP Node -> Routing Node -> Swapper Node
-        let (max_to_pay, _) = self
+        let max_to_pay = self
             .node_api
             .max_sendable_amount(
+                lsp_pubkey,
                 Some(
                     hex::decode(&last_hop.src_node_id)
                         .map_err(|e| anyhow!("Failed to decode hex node_id: {e}"))?,
@@ -733,7 +736,14 @@ impl BreezServices {
             Use the in_progress_reverse_swaps method to get an overview of currently ongoing reverse swaps".into(), 
         });
 
-        let full_rsi = self.btc_send_swapper.create_reverse_swap(req).await?;
+        let lsp_info = self.lsp_info().await?;
+        let lsp_pubkey = hex::decode(lsp_info.pubkey)
+            .map_err(|e| anyhow!("Failed to decode lsp pubkey: {e}"))?;
+
+        let full_rsi = self
+            .btc_send_swapper
+            .create_reverse_swap(req, lsp_pubkey)
+            .await?;
         let rsi = self
             .btc_send_swapper
             .convert_reverse_swap_info(full_rsi)
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -693,6 +693,36 @@ impl BreezServices {
         Ok(res)
     }
 
+    /// Returns the max amount that can be sent on-chain using the send_onchain method.
+    /// The returned amount is the sum of the max amount that can be sent on each channel
+    /// minus the expected fees.
+    /// This is possible since the route to the swapper node is known in advance ans is expected
+    /// to consist of maximum 3 hops.
+    pub async fn max_reverse_swap_amount(&self) -> SdkResult<MaxReverseSwapAmountResponse> {
+        // fetch the last hop hints from the swapper
+        let last_hop = self.btc_send_swapper.last_hop_for_payment().await?;
+
+        // calculate the largest payment we can send over this route using maximum 3 hops
+        // as follows:
+        // User Node -> LSP Node -> Routing Node -> Swapper Node
+        let (max_to_pay, _) = self
+            .node_api
+            .max_amount_to_send(
+                Some(
+                    hex::decode(&last_hop.src_node_id)
+                        .map_err(|e| anyhow!("Failed to decode hex node_id: {e}"))?,
+                ),
+                swap_out::reverseswap::MAX_PAYMENT_PATH_HOPS,
+                Some(&last_hop),
+            )
+            .await?;
+
+        // Sum the max amount per channel and return the result
+        let total_msat = max_to_pay.into_iter().map(|m| m.amount_msat).sum();
+
+        Ok(MaxReverseSwapAmountResponse { total_msat })
+    }
+
     /// Creates a reverse swap and attempts to pay the HODL invoice
     pub async fn send_onchain(
         &self,
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -456,14 +456,14 @@ impl Greenlight {
         via_peer_channels: Vec<ListpeersPeersChannels>,
         payee_node_id: Option<Vec<u8>>,
         max_hops: u32,
-        last_hop_hints: Option<&RouteHintHop>,
+        last_hop_hint: Option<&RouteHintHop>,
     ) -> NodeResult<Vec<MaxChannelAmount>> {
         let mut client = self.get_node_client().await?;
 
         // Consider the hints as part of the route. If there is a routing hint we will
         // attempt to calculate the path until the last hop in the hint and then add
         // the last hop to the path.
-        let (last_node, max_hops) = match last_hop_hints {
+        let (last_node, max_hops) = match last_hop_hint {
             Some(hop) => (hex::decode(&hop.src_node_id)?, max_hops - 1),
             None => match payee_node_id.clone() {
                 Some(node_id) => (node_id, max_hops),
@@ -544,7 +544,7 @@ impl Greenlight {
                 .await?;
 
             // Add the last hop hints (if any) to the route
-            if let Some(hint) = last_hop_hints {
+            if let Some(hint) = last_hop_hint {
                 payment_path.edges.extend(vec![PaymentPathEdge {
                     base_fee_msat: hint.fees_base_msat as u64,
                     fee_per_millionth: hint.fees_proportional_millionths as u64,
@@ -1198,7 +1198,7 @@ impl NodeAPI for Greenlight {
         &self,
         payee_node_id: Option<Vec<u8>>,
         max_hops: u32,
-        last_hop_hints: Option<&RouteHintHop>,
+        last_hop_hint: Option<&RouteHintHop>,
     ) -> NodeResult<Vec<MaxChannelAmount>> {
         let mut client = self.get_node_client().await?;
 
@@ -1215,7 +1215,7 @@ impl NodeAPI for Greenlight {
                     peer.channels,
                     payee_node_id.clone(),
                     max_hops,
-                    last_hop_hints,
+                    last_hop_hint,
                 )
                 .await?;
             max_channel_amounts.extend_from_slice(max_amounts_for_peer.as_slice());
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -1103,7 +1103,7 @@ impl NodeAPI for Greenlight {
                 Some(node_id) => (node_id, max_hops),
                 None => {
                     return Err(NodeError::RouteNotFound(anyhow!(
-                        "no payee node id or last hop hints provided, cannot calculate max amount"
+                        "No payee node id or last hop hints provided, cannot calculate max amount"
                     )));
                 }
             },
@@ -1119,7 +1119,7 @@ impl NodeAPI for Greenlight {
                 fromid: Some(via_peer_id.clone()),
                 fuzzpercent: Some(0),
                 exclude: vec![],
-                // we reduce the first hope that we calculate manually
+                // we deduct the first hop that we calculate manually
                 maxhops: Some(max_hops - 1),
             })
             .await?
@@ -1175,7 +1175,7 @@ impl NodeAPI for Greenlight {
             let chan_id = c
                 .clone()
                 .channel_id
-                .ok_or(NodeError::Generic(anyhow!("empty channel id")))?;
+                .ok_or(NodeError::Generic(anyhow!("Empty channel id")))?;
 
             // First hop is forwarding so no fees and delays.
             let first_edge = PaymentPathEdge {
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -32,6 +32,7 @@ use gl_client::tls::TlsConfig;
 use gl_client::{node, utils};
 use lightning::util::message_signing::verify;
 use lightning_invoice::{RawInvoice, SignedRawInvoice};
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use strum_macros::{Display, EnumString};
 use tokio::sync::{mpsc, Mutex};
@@ -753,6 +754,8 @@ impl NodeAPI for Greenlight {
         let mut amount_sent_msat = 0;
         // The total amount received by the recipient
         let mut amount_received_msat = 0;
+        // Generate a random group_id for the payment
+        let group_id = rand::random::<u64>();
 
         // The algorithm goes over each channel and drains it until the received amount
         // equals to the amount to pay defined in the bolt11 invoice.
@@ -805,7 +808,7 @@ impl NodeAPI for Greenlight {
                 payment_hash: hex::decode(invoice.payment_hash.clone())?,
                 partid: Some(1),
                 timeout: Some(self.sdk_config.payment_timeout_sec),
-                groupid: None,
+                groupid: Some(group_id),
             })
             .await?
             .into_inner();
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -435,8 +435,8 @@ impl Greenlight {
 
             info!("found channel in route: {:?}", first_channel);
             hops.push(PaymentPathEdge {
-                base_fee_msat: first_channel.base_fee_millisatoshi,
-                fee_per_millionth: first_channel.fee_per_millionth,
+                base_fee_msat: first_channel.base_fee_millisatoshi as u64,
+                fee_per_millionth: first_channel.fee_per_millionth as u64,
                 node_id: hop.id.clone(),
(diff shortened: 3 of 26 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -1108,6 +1122,13 @@ pub(crate) fn parse_short_channel_id(id_str: &str) -> Result<u64> {
     Ok((block_num & 0xFFFFFF) << 40 | (tx_num & 0xFFFFFF) << 16 | (tx_out & 0xFFFF))
 }
 
+pub(crate) fn format_short_channel_id(id: u64) -> String {
+    let block_num = (id >> 40) as u32;
+    let tx_num = ((id >> 16) & 0xFFFFFF) as u32;
+    let tx_out = id as u16;
```
dangeross, line 1128: Should this be?
```
let tx_out = (id & 0xFFFF) as u16;
```

libs/sdk-core/src/node_api.rs
```diff
@@ -76,6 +77,19 @@ pub trait NodeAPI: Send + Sync {
         node_id: String,
         amount_msat: u64,
     ) -> NodeResult<PaymentResponse>;
+
+    /// Attempts to find a payment path "manually" and send the htlcs in a way that will drain
+    /// Large channels first.
+    /// This is useful function to send the largest anount possible to a node.
+    async fn send_pay(&self, bolt11: String, max_hops: u32) -> NodeResult<PaymentResponse>;
+
+    /// Calcualtes the maximum amount that can be sent to a node.
```
dangeross, line 86: nit: **Calculates** the maximum

libs/sdk-core/src/swap_out/boltzswap.rs
```diff
@@ -14,15 +14,74 @@ use reqwest::{Body, Client};
...
 use crate::swap_out::reverseswap::CreateReverseSwapResponse;
-use crate::ReverseSwapServiceAPI;
+use crate::{ReverseSwapServiceAPI, RouteHint, RouteHintHop};
 
 use super::error::{ReverseSwapError, ReverseSwapResult};
 
 const BOLTZ_API_URL: &str = "https://api.boltz.exchange/";
 const GET_PAIRS_ENDPOINT: &str = concatcp!(BOLTZ_API_URL, "getpairs");
 const GET_SWAP_STATUS_ENDPOINT: &str = concatcp!(BOLTZ_API_URL, "swapstatus");
+const GET_ROUTE_HINTS_ENDPIONT: &str = concatcp!(BOLTZ_API_URL, "routinghints");
```
dangeross, line 24: typo: GET_ROUTE_HINTS_END**POINT**

tools/sdk-cli/src/commands.rs
```diff
@@ -80,6 +86,8 @@ pub(crate) enum Commands {
         sat_per_vbyte: u32,
     },
 
+    MaxOnchainPayment {},
```
dangeross, line 89: Any reason this is named differently in the cli? I could read this as max send or receive onchain payment

dangeross (changes requested):
Mostly nits about spelling in comments, otherwise looks great

libs/sdk-core/src/breez_services.rs
```diff
@@ -672,6 +672,36 @@ impl BreezServices {
         Ok(res)
     }
 
+    /// Returns the max amount that can be sent on-chain using the send_onchain method.
+    /// The returned amount is the sum of the max amount that can be sent on each channel
+    /// minus the expected fees.
+    /// This is possible since the route to the swapper node is known in advance ans is expected
```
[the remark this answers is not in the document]

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -409,6 +409,37 @@ impl Greenlight {
             .collect();
         Ok(utxos)
     }
+
+    async fn build_payment_path(&self, route: Vec<pb::cln::GetrouteRoute>) -> Result<PaymentPath> {
```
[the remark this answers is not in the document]
roeierez: Done

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -409,6 +409,37 @@ impl Greenlight {
...
+                    destination: None,
+                })
+                .await?
+                .into_inner()
+                .channels;
+
+            let first_channel = hopchannels
+                .first()
+                .ok_or("no route found")
+                .map_err(|e| anyhow!(e))?;
```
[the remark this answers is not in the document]
roeierez: Done

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +582,117 @@ impl NodeAPI for Greenlight {
...
+        let last_hop = invoice.routing_hints.first().and_then(|rh| rh.hops.first());
+        let mut client: node::ClnClient = self.get_node_client().await?;
+
+        // We first calculate for each channel the max amount to pay (at the receiver)
+        let (mut max_amount_per_channel, payment_path) = self
+            .max_amount_to_send(Some(hex::decode(invoice.payee_pubkey)?), max_hops, last_hop)
+            .await?;
+        info!("send_pay: route: {:?}", payment_path);
+
+        // Calculat the total amount to pay
```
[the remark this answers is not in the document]
roeierez: Done

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +582,117 @@ impl NodeAPI for Greenlight {
...
+        let total: u64 = max_amount_per_channel.iter().map(|m| m.amount_msat).sum();
+
+        // Sort the channels by max amount descending so we can build the route in a way that it
+        // drains the largest channels first
+        max_amount_per_channel.sort_by(|m1, m2| m2.amount_msat.cmp(&m1.amount_msat));
+
+        let amount_to_pay = match invoice.amount_msat {
+            Some(amount) => Ok(amount),
+            None => Err(NodeError::Generic(anyhow!("invoice has no amount"))),
```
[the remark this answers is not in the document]
roeierez: Right, Done

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1365,7 +1603,44 @@ impl From<cln::ListpeersPeersChannels> for Channel {
...
+    let hops_arr = route.edges.as_slice();
+
+    for i in 1..hops_arr.len() + 1 {
+        let reverse_index = hops_arr.len() - i;
+        let hop = hops_arr[reverse_index].clone();
+        (to_forward, cltv_delay) = match reverse_index == hops_arr.len() - 1 {
+            // last hop should not take any fees and should use the final_cltv_delta.
+            true => (to_forward, final_cltv_delta),
+
+            // all other hops are forwarding and should take fees and increaase the cltv delta.
```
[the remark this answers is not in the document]
roeierez: Done

libs/sdk-core/src/models.rs
roeierez: Done

libs/sdk-core/src/node_api.rs
roeierez: Done

libs/sdk-core/src/swap_out/boltzswap.rs
roeierez: Done

tools/sdk-cli/src/commands.rs
roeierez: Leftover, Fixed.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -925,6 +1070,102 @@ impl NodeAPI for Greenlight {
         }
     }
 
+    async fn max_amount_to_send(
```
JssDWt, line 1073: This method assumes all channels are with the same peer. You'll want to either get a path for each peer, or filter the used channels to that single peer. Because a different peer will have a different route and different fees.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -925,6 +1070,102 @@ impl NodeAPI for Greenlight {
...
+                channel_delay: hint.cltv_expiry_delta,
+            }])
+        }
+
+        info!("max_amount_to_send: route_hops = {:?}", payment_path.edges);
+
+        // We fetch the opened channels so can calculate max amount to send for each channel
+        let (_, mut opened_channels, _, _) =
+            Self::fetch_channels_and_balance(client.clone()).await?;
+        opened_channels.sort_by(|c1, c2| {
```
JssDWt, line 1142: I think this sort is no longer necessary, because it's sorted here: https://github.com/breez/breez-sdk/pull/466/files#diff-7466a53546c6cded460ef5502a9168b13cc930d94fdf7d0ea8166bb10c75fe7eR604

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +585,117 @@ impl NodeAPI for Greenlight {
...
+                    payment_hash: hex::decode(invoice.payment_hash.clone())?,
+                    label: None,
+                    amount_msat: Some(gl_client::pb::cln::Amount {
+                        msat: amount_to_pay,
+                    }),
+                    bolt11: Some(bolt11.clone()),
+                    payment_secret: Some(invoice.payment_secret.clone()),
+                    partid: Some(part_id),
+                    localinvreqid: None,
+                    groupid: None,
```
JssDWt, line 668: Let's set the groupid here too, because the parts form one logical payment together.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +585,117 @@ impl NodeAPI for Greenlight {
...
+                invoice.min_final_cltv_expiry_delta,
+            );
+            info!(
+                "send_pay route to pay: {:?}, received_amount = {}",
+                route, to_pay
+            );
+
+            // We send the part using the node API
+            client
+                .send_pay(pb::cln::SendpayRequest {
```
JssDWt, line 657: We may want to check the outcome of this methodcall has a status that is `pending`. Otherwise we would have already completed the payment. That shouldn't happen, but doesn't hurt either.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +585,117 @@ impl NodeAPI for Greenlight {
...
+                    route,
+                    payment_hash: hex::decode(invoice.payment_hash.clone())?,
+                    label: None,
+                    amount_msat: Some(gl_client::pb::cln::Amount {
+                        msat: amount_to_pay,
```
JssDWt, line 662:
```suggestion
                        msat: to_pay,
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +585,117 @@ impl NodeAPI for Greenlight {
...
+        // The algorithm goes over each channel and drains it until the recived amount
+        // equals to the amount to pay defined in the bolt11 invoice.
+        for max in max_amount_per_channel {
+            // calculating the incoming amount for the remaining amount to pay.
+            let left_to_pay = amount_to_pay - amount_received;
+            // Whether we draining the whole channel balance or only what is left to pay
+            let to_pay = match left_to_pay > max.amount_msat {
+                true => max.amount_msat,
+                false => left_to_pay,
+            };
```
JssDWt, line 641:
```suggestion
            let to_pay = std::cmp::min(left_to_pay, max.amount_msat);
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +585,117 @@ impl NodeAPI for Greenlight {
...
+            );
+            info!(
+                "send_pay route to pay: {:?}, received_amount = {}",
+                route, to_pay
+            );
+
+            // We send the part using the node API
+            client
+                .send_pay(pb::cln::SendpayRequest {
+                    route,
```
JssDWt, line 658: This route only works for a single peer, see my comment in `max_amount_to_send`

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +585,117 @@ impl NodeAPI for Greenlight {
...
+                bolt11: Some(bolt11.clone()),
+            })
+            .await?;
+
+        // We need to allocate a part id for each part that we are sending.
+        let mut part_id = 1;
+        // The total amount we sent
+        let mut amount_sent = 0;
+        // The total amount received
+        let mut amount_received = 0;
```
JssDWt, line 630: Make it a little clearer what these variables mean.

```suggestion
        // The total amount we sent. i.e. what the recipient received + fees
        let mut total_amount_sent = 0;
        // The total amount received by the recipient
        let mut total_amount_received = 0;
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1365,7 +1606,44 @@ impl From<cln::ListpeersPeersChannels> for Channel {
...
+    route: PaymentPath,
+    to_pay_msat: u64,
+    final_cltv_delta: u64,
+) -> (Vec<pb::cln::SendpayRoute>, u64) {
+    let mut sendpay_route = vec![];
+    let mut to_forward = to_pay_msat;
+    let mut cltv_delay = 0;
+    let hops_arr = route.edges.as_slice();
+
+    for i in 1..hops_arr.len() + 1 {
```
JssDWt, line 1619: A reverse loop here would be a little easier to follow (since `i` is not used)

JssDWt (changes requested):
I've gone over a large chunk, will continue later.

libs/sdk-core/src/greenlight/node_api.rs
ok300: IMO the variable names should include the unit to avoid confusion.
(e.g. `total_amount_sent` -> `total_amount_sent_msat`)

libs/sdk-core/src/models.rs
```diff
@@ -1215,6 +1236,58 @@ impl FromStr for BuyBitcoinProvider {
...
+    pub short_channel_id: String,
+    pub channel_delay: u64,
+    pub base_fee_msat: u32,
+    pub fee_per_millionth: u32,
+}
+
+impl PaymentPathEdge {
+    pub(crate) fn amount_to_forward(&self, in_amount_msat: u64) -> u64 {
+        let forward = (in_amount_msat - self.base_fee_msat as u64) as f64
+            / (1f64 + self.fee_per_millionth as f64 / 1000000f64);
```
ok300, line 1274: `1000000f64` would be better readable as `1_000_000_f64`

ok300 (commented):
Found a few NITs, otherwise LGTM.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: @JssDWt we need to use amount_to_pay since this is how send_pay works. It requires the full amount for every part.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -925,6 +1070,102 @@ impl NodeAPI for Greenlight {
...
+                    )));
+                }
+            },
+        };
+
+        // fetch a route from greenlight
+        let route_response = client
+            .get_route(GetrouteRequest {
+                id: last_node.clone(),
+                amount_msat: Some(pb::cln::Amount { msat: 1_000 }),
```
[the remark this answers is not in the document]
roeierez: I am not sure that it will work as the whole point of constructing a route manually is to bypass the node pathfinding that might return no route for amounts near local balance.
The assumption of this whole feature is that there exist one route with 3 hops that has enough liquidity to the destination so there shouldn't be a difference if we change the amount.
3 hops means: 
`gl -> LSP -> Routing Node -> Swapper`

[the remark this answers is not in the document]
JssDWt: The base fee is a large factor in small payments, so it could play a role if there is an alternative route with lower base fees for example. I understand the pathfinding won't work too well when you get close to the balance. Maybe try with amount * 90% - 1000sat? Or the 1 sat if that ends up to be too low?

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: Ah makes sense, the part amount is in the route.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Good point! we probably need to filter only these channels that reaches the first hop of the route

libs/sdk-core/src/models.rs
```diff
@@ -682,6 +685,17 @@ pub struct ReverseSwapFeesRequest {
...
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MaxReverseSwapAmountResponse {
+    pub total_msat: u64,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MaxChannelAmount {
+    pub channel_id: String,
+    pub amount_msat: u64,
```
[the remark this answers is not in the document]
roeierez: True, channel balances can be combined.

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
roeierez: Fixed

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -925,6 +1070,102 @@ impl NodeAPI for Greenlight {
...
+    async fn max_amount_to_send(
+        &self,
+        payee_node_id: Option<Vec<u8>>,
+        max_hops: u32,
+        last_hop_hints: Option<&RouteHintHop>,
+    ) -> NodeResult<(Vec<MaxChannelAmount>, PaymentPath)> {
+        let mut client = self.get_node_client().await?;
+
+        // Consider the hints as part of the route. If there is a routing hint we will
+        // attmept to calculate the path untill the last hop in the hint and then add
```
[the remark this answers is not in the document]
roeierez: Fixed

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -925,6 +1070,102 @@ impl NodeAPI for Greenlight {
...
+                }
+            },
+        };
+
+        // fetch a route from greenlight
+        let route_response = client
+            .get_route(GetrouteRequest {
+                id: last_node.clone(),
+                amount_msat: Some(pb::cln::Amount { msat: 1_000 }),
+                riskfactor: 0,
```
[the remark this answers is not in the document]
roeierez: The reverse swap route (gl -> LSP -> RoutingNode -> boltz) actually is the cheapest and takes zero fee.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Changed to only consider channels for the first hop.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Yes, that's the idea we only do that via the LSP which is a "stable" route (such that we can route all. htlcs via this route and it has enough liquidity for all of them).

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Fixed

libs/sdk-core/src/models.rs
```diff
@@ -682,6 +685,17 @@ pub struct ReverseSwapFeesRequest {
     pub send_amount_sat: Option<u64>,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MaxReverseSwapAmountResponse {
+    pub total_msat: u64,
```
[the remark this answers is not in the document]
roeierez: Makes sense. Changed.

libs/sdk-core/src/models.rs
roeierez: Fixed

libs/sdk-core/src/models.rs
```diff
@@ -1215,6 +1236,58 @@ impl FromStr for BuyBitcoinProvider {
...
+        let amount_to_forward = forward.floor() as u64;
+        info!("amount_to_forward: in_amount_msat = {in_amount_msat},base_fee_msat={}, fee_per_millionth={}  amount_to_forward: {}", self.base_fee_msat, self.fee_per_millionth, amount_to_forward);
+        amount_to_forward
+    }
+
+    // forward_amount_msat * (1000000 + self.fee_per_millionth)/1000000 + self.base_fee_msat = x
+    pub(crate) fn amount_from_forward(&self, forward_amount_msat: u64) -> u64 {
+        let in_amount_msat = self.base_fee_msat as f64
+            + forward_amount_msat as f64 * (1000000f64 + self.fee_per_millionth as f64)
+                / 1000000f64;
```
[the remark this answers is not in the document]
roeierez: Fixed

libs/sdk-core/src/swap_out/reverseswap.rs
```diff
@@ -95,6 +96,32 @@ impl BTCSendSwap {
...
+            .ok_or_else(|| {
+                ReverseSwapError::Generic(anyhow!(
+                    "No route hints found for reverse routing node {reverse_routing_node:?}"
+                ))
+            })?
+            .hops
+            .first()
+            .ok_or_else(|| {
+                ReverseSwapError::Generic(anyhow!(
+                    "No route hints found for reverse routing node {reverse_routing_node:?}"
```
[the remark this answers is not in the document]
roeierez: Fixed

libs/sdk-core/src/greenlight/node_api.rs
roeierez: We need to iterate over the reversed route and to be able also to update the previous iterated hop (as forwarding hop) with fees and cltv delta.
Any idea how to simplify it?

roeierez: OK, I fixed to use enumerate() which provides also the index.

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
JssDWt: Should this be the case regardless of the LSP used?

ok300 (approved):
LGTM, all the points I raised were already addressed.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +582,117 @@ impl NodeAPI for Greenlight {
...
+                total
+            )));
+        }
+
+        // This is needed in greenlight for the signer to recognize this invoice.
+        client
+            .pre_approve_invoice(pb::cln::PreapproveinvoiceRequest {
+                bolt11: Some(bolt11.clone()),
+            })
+            .await?;
```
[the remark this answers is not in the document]
dangeross: Think this got missed

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -551,6 +582,117 @@ impl NodeAPI for Greenlight {
...
+
+        // We need to allocate a part id for each part that we are sending.
+        let mut part_id = 1;
+        // The total amount we sent
+        let mut amount_sent = 0;
+        // The total amount received
+        let mut amount_received = 0;
+
+        // The algorithm goes over each channel and drains it until the recived amount
```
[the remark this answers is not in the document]
dangeross: Also missed

dangeross (changes requested):
Looking good, i think a few comments are hidden/missed that haven't been addressed. There are a couple of open questions from @JssDWt regarding routes

libs/sdk-core/src/models.rs
```diff
@@ -1215,6 +1236,57 @@ impl FromStr for BuyBitcoinProvider {
...
+    pub base_fee_msat: u32,
+    pub fee_per_millionth: u32,
+}
+
+impl PaymentPathEdge {
+    pub(crate) fn amount_to_forward(&self, in_amount_msat: u64) -> u64 {
+        let forward = (in_amount_msat - self.base_fee_msat as u64) as f64
+            / (1f64 + self.fee_per_millionth as f64 / 1_000_000f64);
+
+        let amount_to_forward = forward.floor() as u64;
```
JssDWt, line 1276: I think this should be

```suggestion
               let amount_to_forward = ((in_amount_msat - self.base_fee_msat as u64) * 1000000
            + 1000000
            + self.fee_per_millionth as u64
            - 1)
            / (1000000 + self.fee_per_millionth as u64);
```
Source: https://github.com/lightning/bolts/blob/master/04-onion-routing.md?plain=1#L313

Similarly for amount_from_forward, but the other way around.
It changes the rounding a bit.

libs/sdk-core/src/node_api.rs
```diff
@@ -76,6 +77,19 @@ pub trait NodeAPI: Send + Sync {
         node_id: String,
         amount_msat: u64,
     ) -> NodeResult<PaymentResponse>;
+
+    /// Attempts to find a payment path "manually" and send the htlcs in a way that will drain
+    /// Large channels first.
+    /// This is useful function to send the largest anount possible to a node.
```
JssDWt, line 83:
```suggestion
    /// This is useful function to send the largest amount possible to a node.
```

JssDWt (commented):
A few more comments.
It's looking really good. Great work.

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
roeierez: The swapper node should be connected to the LSP and the sdk should have a stable route to it via the LSP, that's why we ask the swapper API for the routing hints before calculating the path.
So the lsp must be used ATM.

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
roeierez: As long as we get the route, Why should we prefer this route or another in the context of this function?
In the outer scope the way we use it we ensure we get the route we want because we are combining with the swapper routing hints.

libs/sdk-core/src/models.rs
roeierez: Good catch! I wish I saw that before I had to come up with this calculation myself.
Fixed

libs/sdk-core/src/node_api.rs
roeierez: Fixed

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
roeierez: I see in the docs now that we can pass 0 so it ignores the amount and just lookup for a path. I think it makes more sense so will try that.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -925,6 +1067,124 @@ impl NodeAPI for Greenlight {
...
+                fee_per_millionth: hint.fees_proportional_millionths,
+                node_id: payee_node_id.unwrap_or_default(),
+                short_channel_id: format_short_channel_id(hint.short_channel_id),
+                channel_delay: hint.cltv_expiry_delta,
+            }])
+        }
+
+        info!("max_sendable_amount: route_hops = {:?}", payment_path.edges);
+
+        // We first want to find the channels with the first hop of the route
```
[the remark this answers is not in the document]
roeierez: Makes sense. I see how I can change it to use the lsp peer.

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
roeierez: I pushed using zero amount. This should also bypass the min_htlc limits.

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
roeierez: I changed this method to get the `via_peer_id`.
We also use it with the lsp peer id so now it is explicitly mentioned that the routing for `send_onchain` is via the lsp.
I also changed a bit the way max_sendable_amount returns the result as I found an issue with multiple channels.
Now every amount per channel has its associated path with its unique first hop: https://github.com/breez/breez-sdk/pull/466/commits/eb36970d20490777599658956a27ffa1d73f3d30

libs/sdk-core/src/models.rs
```diff
@@ -682,6 +685,18 @@ pub struct ReverseSwapFeesRequest {
...
+    pub total_sat: u64,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MaxChannelAmount {
+    pub channel_id: String,
+    pub amount_msat: u64,
+    pub path: PaymentPath,
+}
+
```
erdemyerebasmaz, line 699: Can we add documentation for these newly introduced structs and their fields for the end-users?
They don't need to be in-depth, a brief explanation is enough.

dangeross (changes requested):
Mostly comments around errors, otherwise looks ready I think

dangeross:
Also there are still conflicts

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -544,6 +587,124 @@ impl NodeAPI for Greenlight {
...
+            )
+            .await?;
+        info!("send_pay: routes: {:?}", max_amount_per_channel);
+
+        // Calculate the total amount to pay
+        let total_msat: u64 = max_amount_per_channel.iter().map(|m| m.amount_msat).sum();
+
+        // Sort the channels by max amount descending so we can build the route in a way that it
+        // drains the largest channels first
+        max_amount_per_channel.sort_by(|m1, m2| m2.amount_msat.cmp(&m1.amount_msat));
```
ok300, line 616:
```suggestion
        max_amount_per_channel.sort_by_key(|m| Reverse(m.amount_msat));
```

ok300: Nit: this makes it more readable.

libs/sdk-core/src/models.rs
```diff
@@ -1215,6 +1237,66 @@ impl FromStr for BuyBitcoinProvider {
...
+impl PaymentPathEdge {
+    pub(crate) fn amount_to_forward(&self, in_amount_msat: u64) -> u64 {
+        // return incomingAmt - divideCeil(
+        //     feeRateParts*(incomingAmt-c.FeeBaseMSat),
+        //     feeRateParts+c.FeeProportionalMillionths,
+        // )
+        let amount_to_forward = Self::devide_ceil(
+            1_000_000 * (in_amount_msat - self.base_fee_msat),
+            1000000 + self.fee_per_millionth,
```
ok300, line 1280:
```suggestion
            1_000_000 + self.fee_per_millionth,
```

ok300 (approved):

libs/sdk-core/src/models.rs
roeierez: Done

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Done

libs/sdk-core/src/models.rs
roeierez: Done

dangeross (approved):

roeierez:
I made one more commit to allow also draining channels with peers other than the lsp, as long as we have a route from these peers.
Also when calculating the channels max amounts we ignore now specific peers that have no route instead of returning error for that endpoint.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -406,6 +413,161 @@ impl Greenlight {
...
+
+        // opened_channels.into_iter().filter(|c: &cln::ListpeersPeersChannels|c)
+        opened_channels.sort_by(|c1, c2| {
+            c2.spendable_msat
+                .clone()
+                .unwrap_or_default()
+                .msat
+                .cmp(&c1.spendable_msat.clone().unwrap_or_default().msat)
+        });
+
```
JssDWt, line 524: This can be removed. (re https://github.com/breez/breez-sdk/pull/466#discussion_r1383148636)
```suggestion
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -924,6 +1194,35 @@ impl NodeAPI for Greenlight {
         }
     }
 
+    async fn max_sendable_amount(
+        &self,
+        payee_node_id: Option<Vec<u8>>,
+        max_hops: u32,
+        last_hop_hints: Option<&RouteHintHop>,
```
JssDWt, line 1201: nit: singular for now.
```suggestion
        last_hop_hint: Option<&RouteHintHop>,
```

libs/sdk-core/src/models.rs
```diff
@@ -1215,6 +1241,62 @@ impl FromStr for BuyBitcoinProvider {
...
+
+    pub(crate) fn amount_from_forward(&self, forward_amount_msat: u64) -> u64 {
+        let in_amount_msat = self.base_fee_msat
+            + forward_amount_msat * (1_000_000 + self.fee_per_millionth) / 1_000_000;
+
+        print!("amount_from_forward: in_amount_msat = {in_amount_msat},base_fee_msat={}, fee_per_millionth={}  amount_to_forward: {}", self.base_fee_msat, self.fee_per_millionth, forward_amount_msat);
+        in_amount_msat
+    }
+
+    fn devide_ceil(dividend: u64, factor: u64) -> u64 {
```
JssDWt, line 1295: nit
```suggestion
    fn divide_ceil(dividend: u64, factor: u64) -> u64 {
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -548,6 +710,114 @@ impl NodeAPI for Greenlight {
...
+                invoice.min_final_cltv_expiry_delta,
+            );
+            info!(
+                "send_pay route to pay: {:?}, received_amount = {}",
+                route, to_pay_msat
+            );
+
+            // We send the part using the node API
+            client
+                .send_pay(SendpayRequest {
```
JssDWt, line 779: re https://github.com/breez/breez-sdk/pull/466#discussion_r1383164929

JssDWt (commented):
Looking really good, we're getting close!
The algorithm for finding routes per peer looks good.
_Maybe_ we want to try for each hop hint in case the first one doesn't give any routes, but not sure whether that's necessary. Can there ever be more hop hints in an invoice?

libs/sdk-core/src/greenlight/node_api.rs
roeierez: If there is no practical implications I prefer not to complicate?

JssDWt: Yeah makes sense.

libs/sdk-core/src/binding.rs
```diff
@@ -300,6 +300,11 @@ pub fn list_fiat_currencies() -> Result<Vec<FiatCurrency>> {
 
 /*  On-Chain Swap API's */
 
+pub fn max_reverse_swap_amount() -> Result<MaxReverseSwapAmountResponse> {
```
erdemyerebasmaz, line 303: `/// See [BreezServices::max_reverse_swap_amount]`

libs/sdk-core/src/models.rs
```diff
@@ -682,6 +685,22 @@ pub struct ReverseSwapFeesRequest {
...
+    pub total_sat: u64,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MaxChannelAmount {
+    /// The channel id.
+    pub channel_id: String,
+    /// The max amount can be sent from this channel.
+    pub amount_msat: u64,
+    /// The payment path to be used for thte maximum amount.
```
erdemyerebasmaz, line 700: Typo
```suggestion
    /// The payment path to be used for the maximum amount.
```

libs/sdk-core/src/greenlight/node_api.rs
roeierez: In the new code we are not using `fetch_channels_and_balance` anymore so the sorting is needed

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -548,6 +710,114 @@ impl NodeAPI for Greenlight {
...
+            amount_sent_msat += sent_msat;
+            amount_received_msat += to_pay_msat;
+            if amount_received_msat == amount_to_pay_msat {
+                break;
+            }
+        }
+
+        // Now we wait for the first part to be completed as a way to wait for the payment
+        // to complete.
+        let response = client
```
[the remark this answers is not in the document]
roeierez: Noted. TBH I prefer to start simple with this first implementation.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: fixed

libs/sdk-core/src/greenlight/node_api.rs
roeierez: See my comment there, it is still needed.

libs/sdk-core/src/models.rs
roeierez: fixed

libs/sdk-core/src/greenlight/node_api.rs
roeierez: See comment there, I prefer to keep it simple for this first version.

libs/sdk-core/src/binding.rs
roeierez: fixed

libs/sdk-core/src/models.rs
roeierez: fixed

JssDWt (approved):
LGTM!

erdemyerebasmaz (approved):

Issue #428, opened 2023-09-06
Logger per sdk instance(node)


roeierez:
Currently the SDK use rust log crate which uses a single logger. While this works fine when there is only one sdk node in the process,  it doesn't scale well to multiple nodes.
The idea is to pass the `connect` method, which is the entry point to connect to the node, a logger and an optional log directory.
From that point all log lines that are related to that node would be streamed to the given logger and if a logging directory was passed also to a file in that directory.
This will also need to propagate to the binding languages: dart, udl file, React Native.
It will require different code pattern when writing to the log so  we will also need to convert existing log statement to the new pattern.

PR #467, opened 2023-09-19
Logger per sdk instance(node)

Torakushi:
Fixes https://github.com/breez/breez-sdk/issues/428

the second commit is way bigger than the other, but it's simply adding the logger to the signatures and replacing the macros

- [ ] The PR in breez-rn-generator needs to be merged before: https://github.com/breez/breez-sdk-rn-generator/pull/8

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -1,10 +1,10 @@
-dictionary RouteHintHop {    
+dictionary RouteHintHop {
      string src_node_id;
      u64 short_channel_id;
      u32 fees_base_msat;
-     u32 fees_proportional_millionths;   
-     u64 cltv_expiry_delta;    
-     u64? htlc_minimum_msat;    
+     u32 fees_proportional_millionths;
+     u64 cltv_expiry_delta;
+     u64? htlc_minimum_msat;
      u64? htlc_maximum_msat;
  };
 
@@ -97,7 +97,7 @@ dictionary NodeState {
     u64 max_single_payment_amount_msat;
     u64 max_chan_reserve_msats;
     sequence<string> connected_peers;
-    u64 inbound_liquidity_msats;    
+    u64 inbound_liquidity_msats;
 };
 
 dictionary SignMessageRequest {
@@ -138,8 +138,8 @@ enum PaymentType {
 
 dictionary Payment {
     string id;
-    PaymentType payment_type;    
-    i64 payment_time;    
+    PaymentType payment_type;
+    i64 payment_time;
     u64 amount_msat;
     u64 fee_msat;
     PaymentStatus status;
@@ -271,7 +271,7 @@ dictionary PaymentFailedData {
 };
 
 dictionary BackupFailedData {
-    string error;    
+    string error;
 };
 
 [Enum]
@@ -282,7 +282,7 @@ interface BreezEvent {
     PaymentSucceed(Payment details);
     PaymentFailed(PaymentFailedData details);
     BackupStarted();
-    BackupSucceeded(); 
+    BackupSucceeded();
     BackupFailed(BackupFailedData details);
 };
 
@@ -295,7 +295,7 @@ callback interface LogStream {
     void log(LogEntry l);
 };
 
-callback interface EventListener {    
+callback interface EventListener {
     void on_event(BreezEvent e);
 };
 
@@ -539,7 +539,28 @@ dictionary SweepResponse {
     sequence<u8> txid;
 };
 
-interface BlockingBreezServices {   
+callback interface Logger {
+    void log(LogMessage log_message);
+};
+
+enum LogLevel {
+    "Error",
+    "Warn",
+    "Info",
+    "Debug",
+    "Trace",
+};
+
+// Define the LogMessage struct
+dictionary LogMessage {
+    LogLevel level;
+    string message;
+    string module_path;
+    string file;
+    u32 line;
+};
+
+interface BlockingBreezServices {
 
    [Throws=SdkError]
    void disconnect();
@@ -647,10 +668,10 @@ interface BlockingBreezServices {
    BuyBitcoinResponse buy_bitcoin(BuyBitcoinRequest req);
 };
 
-namespace breez_sdk {  
+namespace breez_sdk {
 
  [Throws=SdkError]
- BlockingBreezServices connect(Config config, sequence<u8> seed, EventListener listener);
+ BlockingBreezServices connect(Config config, sequence<u8> seed, EventListener listener, Logger node_logger, string? log_file_path);
 
  [Throws=SdkError]
  void set_log_stream(LogStream log_stream);
--- a/libs/sdk-core/src/binding.rs
+++ b/libs/sdk-core/src/binding.rs
@@ -31,7 +31,7 @@ use crate::models::{Config, LogEntry, NodeState, Payment, SwapInfo};
 use crate::{
     BackupStatus, BuyBitcoinRequest, BuyBitcoinResponse, CheckMessageRequest, CheckMessageResponse,
     EnvironmentType, ListPaymentsRequest, LnUrlCallbackStatus, LnUrlWithdrawRequest,
-    LnUrlWithdrawResult, NodeConfig, OpenChannelFeeRequest, OpenChannelFeeResponse,
+    LnUrlWithdrawResult, NodeConfig, NopLogger, OpenChannelFeeRequest, OpenChannelFeeResponse,
     ReceiveOnchainRequest, ReceivePaymentRequest, ReceivePaymentResponse, ReverseSwapFeesRequest,
     ReverseSwapInfo, ReverseSwapPairInfo, SignMessageRequest, SignMessageResponse,
     StaticBackupRequest, StaticBackupResponse, SweepRequest, SweepResponse,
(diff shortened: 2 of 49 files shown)


Torakushi:
I have two concerns about log injection:

1) In OpenFeeParamsMenu, we can log a warning when we fail to parse 'valid_unit'. IMO, we should not inject a logger into this structure as it is a simple "tuple struct", and it makes no sense to inject a logger into a "params" structure (https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/models.rs#L909). What's more, we'll have to modify the udl file a LOT just to log ' warn !("Failed to parse valid_until for OpeningFeeParams : {ofp:?}")'

2) Another case is that there are functions called by BreezServices that make HTTP requests and save the result. However, they use a generic static "Parse" function which is used throughout the SDK and not just for HTTP requests. In relation to my first point, it makes sense to log the HTTP response using the node's logger. But this implies "heavy changes" (logger choice logic, etc.). I've made a test commit, to show you what it looks like and if it works for you! (If not, we can simply use the global logger).

Looking forward to hearing your thoughts

Torakushi:
I have two concerns about log injection:

In OpenFeeParamsMenu, we can log a warning when we fail to parse 'valid_unit'. IMO, we should not inject a logger into this structure as it is a simple "tuple struct", and it makes no sense to inject a logger into a "params" structure (https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/models.rs#L909). What's more, we'll have to modify the udl file a LOT just to log ' warn !("Failed to parse valid_until for OpeningFeeParams : {ofp:?}")'

Another case is that there are functions called by BreezServices that make HTTP requests and save the result. However, they use a generic static "Parse" function which is used throughout the SDK and not just for HTTP requests. In relation to my first point, it makes sense to log the HTTP response using the node's logger. But this implies "heavy changes" (logger choice logic, etc.). I've made a test commit, to show you what it looks like and if it works for you! (If not, we can simply use the global logger).

Looking forward to hearing your thoughts

Torakushi:
I have two concerns about log injection:

In OpenFeeParamsMenu, we can log a warning when we fail to parse 'valid_unit'. IMO, we should not inject a logger into this structure as it is a simple "tuple struct", and it makes no sense to inject a logger into a "params" structure (https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/models.rs#L909). What's more, we'll have to modify the udl file a LOT just to log ' warn !("Failed to parse valid_until for OpeningFeeParams : {ofp:?}")'

Another case is that there are functions called by BreezServices that make HTTP requests and save the result. However, they use a generic static "Parse" function which is used throughout the SDK and not just for HTTP requests. In relation to my first point, it makes sense to log the HTTP response using the node's logger. But this implies "heavy changes" (logger choice logic, etc.). I've made a test commit, to show you what it looks like and if it works for you! (If not, we can simply use the global logger).

Looking forward to hearing your thoughts

JssDWt:
This looks promising! The PR sure became big with all the loglines changed, sorry about that.

I enabled the CI on your PR. You'll get some review from clippy. You can run `cargo fmt` in the `libs` folder to format your code, and `cargo clippy && cargo clippy --tests` to get some additional code formatting checks from clippy.


> In OpenFeeParamsMenu, we can log a warning when we fail to parse 'valid_unit'. IMO, we should not inject a logger into this structure as it is a simple "tuple struct", and it makes no sense to inject a logger into a "params" structure (https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/models.rs#L909). What's more, we'll have to modify the udl file a LOT just to log ' warn !("Failed to parse valid_until for OpeningFeeParams : {ofp:?}")'

We definitely don't want to clutter he udl file with loggers, other than the static methods we expose I think. Also not injecting a logger there seems reasonable. Maybe that specific check should be more strict and validation of the menu should just fail with an error, so the caller can log the error. (not part of this PR). I guess with injected loggers we have to be a bit more mindful about where to place logs.


> Another case is that there are functions called by BreezServices that make HTTP requests and save the result. However, they use a generic static "Parse" function which is used throughout the SDK and not just for HTTP requests. In relation to my first point, it makes sense to log the HTTP response using the node's logger. But this implies "heavy changes" (logger choice logic, etc.). I've made a test commit, to show you what it looks like and if it works for you! (If not, we can simply use the global logger).

This makes perfect sense, to pass the logger around for these cases. Probably it shouldn't even be optional. What do you mean with "logger choice logic"?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -628,10 +649,10 @@ interface BlockingBreezServices {
    BuyBitcoinResponse buy_bitcoin(BuyBitcoinRequest req);
 };
 
-namespace breez_sdk {  
+namespace breez_sdk {
 
  [Throws=SdkError]
- BlockingBreezServices connect(Config config, sequence<u8> seed, EventListener listener);
+ BlockingBreezServices connect(Config config, sequence<u8> seed, EventListener listener, Logger node_logger, string? log_file_path);
```
JssDWt, line 655: I think it's time to wrap this into a `ConnectRequest`. 
```
dictionary ConnectRequest {
    Config config;
    sequence<u8> seed;
    EventListener listener;
    Logger? logger=null;
    string? log_file_path=null;
};
```

The logger can probably be optional. I think by default we can initialize our own logger that logs to stdout. (will that work? I don't know). If the client sets a logger, we use that instead.

libs/sdk-core/src/binding.rs
```diff
@@ -143,6 +149,14 @@ pub fn breez_events_stream(s: StreamSink<BreezEvent>) -> Result<()> {
     Ok(())
 }
 
+/// If used, this must be called before `connect`. It can only be called once.
+pub fn breez_node_log_stream(s: StreamSink<LogMessage>) -> Result<()> {
```
JssDWt, line 153: Yeah there's two log streams here now. I guess that's inevitable. Probably a good solution.

JssDWt (commented):
Went over the first half of the files. Not too many comments!

libs/sdk-bindings/src/breez_sdk.udl
Torakushi: Interesting! I do agree (and it will help for the multi-instances static_backup static call as well)!

I will give it a try!

Torakushi:
@JssDWt 
> This makes perfect sense, to pass the logger around for these cases. Probably it shouldn't even be optional. What do you > mean with "logger choice logic"?

Sorry, I was unclear! 

I talked about this [macro I created]https://github.com/breez/breez-sdk/pull/467/commits/07581ab04e371071c2863ef5fb71809c895f2781#diff-1e112ace150c8d7fa35768660b6784d2777195efd868d2b19bddb9e3cbb6fa21R51) that will choose between the local logger and the global logger

```
#[macro_export]
macro_rules! log_or_debug_generic {
    ($logger:expr, $lvl:expr, $message:expr, $($args:tt)*) => {
        if let Some(logger) = $logger {
            $crate::log!(logger, $lvl, $message, $($args)*);
        } else {
            match $lvl {
                $crate::LogLevel::Error => error!($message, $($args)*),
                ............
```

As it's optional, If no logger is provided, I will use the global logger. IMO, it's a good way to as as we use 'Parse<>' function everywhere and not only for the node (it will be quite heavy to add a logger for every 'Parse<>' in the code (especially as we are only logging for a very specific case of this function ). 
In addition, not making the logger optional will force us to modify the udl file because we are exposing functions that use 'parse' directly: 'parse_input',..

libs/sdk-bindings/src/breez_sdk.udl
Torakushi: I quickly tried to implement the "connectRequest" yesterday and it's not trivial. The problem is that you can't simply put a trait in an udl structure (because you have to implement Equalable, Hashable,...). So we should rather work with references, but after trying, I don't get the impression that it works... (I've followed their documentation, but no success... It was a quick try but it still takes a bit of work to try it out maybe)

In any case, if we do that, we'll have to change a lot of things, (add Arc everywhere, even for the evenHandler and then we'll have pretty things like:
"self.event_listener.as_ref().as_ref().unwrap().on_event(e.clone())")

Given the changes this would involve, I wonder if this is still within the scope of the PR? or should it be a separated PR?

If you think so, I'll look into it seriously!

roeierez: > I quickly tried to implement the "connectRequest" yesterday and it's not trivial. The problem is that you can't simply put a trait in an udl structure (because you have to implement Equalable, Hashable,...). So we should rather work with references, but after trying, I don't get the impression that it works... (I've followed their documentation, but no success... It was a quick try but it still takes a bit of work to try it out maybe)
> 
> In any case, if we do that, we'll have to change a lot of things, (add Arc everywhere, even for the evenHandler and then we'll have pretty things like: "self.event_listener.as_ref().as_ref().unwrap().on_event(e.clone())")
> 
> Given the changes this would involve, I wonder if this is still within the scope of the PR? or should it be a separated PR?
> 
> If you think so, I'll look into it seriously!

@Torakushi given this analysis it is definitely not part of this PR which is already large enough.

JssDWt:
In general this PR looks really good I think. You nailed it. 

I want to rediscuss the benefits of the bigger picture of this structure. The benefit is we have _our_ node logs structured on a per-node basis, so we can split these up to the appropriate loggers. The problem is still that our dependencies don't use this structure. So any logging from a dependency will not go to the appropriate logger. The most important dependencies being gl-client and vls currently. 

That's fine, and definitely not part of the scope of this PR. But I think we do need a plan, or at least an outlook on a path to get these logs into the same logger eventually before merging this.

Issue #336, opened 2023-07-18
Update SDK service methods to return `SdkResult`


ok300:
Change existing service from returning `anyhow::Result` to return `SdkResult`.

This migration is completed when the trait

https://github.com/breez/breez-sdk/blob/b8f1c490d99df82f2fd3a5b4ef7fcf9d2908d488/libs/sdk-core/src/error.rs#L41-L48

is not needed anymore.

A good way to find which method to migrate next is therefore to comment out that trait implementation, then run `cargo clippy` in `/libs/sdk-core` to see which service methods cause errors.

When all service methods are migrated, the trait above should be deleted.

Partially addressed in #298

roeierez:
@dangeross can we close this?

dangeross:
I just checked and all exposed methods except `prepare_refund` return SdkResult or a Result with a method specific error. I will make a PR to change `prepare_refund`.

Issue #458, opened 2023-09-14
Occasional "payment not found" error on `sendPayment`


ok300:
Customer report:

> I encountered an exception while sending a payment with the `sendPayment()` method, as follows: `FfiException(RESULT_ERROR, payment not found, null).`
This is just a 2000 sats invoice generated from the Alby wallet that I'm trying to pay. The thing is, the payment did become successful, and the wallet balance did change. However, with the `await _breezBridge.sendPayment(bolt11: bolt11);` method, the exception was thrown and caught. I tried to pay another invoice to see if this particular exception is thrown again or not. Hopefully, this exception is not thrown, and the payment becomes successful, so it was just thrown once.

The payment in question:
```
{
  "bolt11": lnbc20u1pjs9l4dpp533ew4wgd8n68j30u06hdgq6fj8gfmkxqxja80n33yht5mach2x4sdq5g9kxy7fqd9h8vmmfvdjscqzzsxqyz5vqsp5ptqjepu9mthd8x0uvlx68480vhjjnrmlvrv09hn0pgwrl7dtfsgq9qyyssqdk6037rh5fq7n78cmsgy78zsnwaknqsc3heceqyz5uw6lfmehkppfp6u7k6xv0q6qz42qv3kf2mwrgvkcpmw7peacwt9vx0xlpldfhgqjal99s,
  "payeePubkey": 030a58b8653d32b99200a2334cfe913e51dc7d155aa0116c176657a4f1722677a3,
  "paymentHash": 8c72eab90d3cf47945fc7eaed4034991d09dd8c034ba77ce3125d74df71751ab,
  "description": Alby invoice,
  "descriptionHash": null,
  "amountMsat": 2000000,
  "timestamp": 1694695085,
  "expiry": 86400,
  "paymentSecret": [10, 193, 44, 135, 133, 218, 238, 211, 153, 252, 103, 205, 163, 212, 239, 101, 229, 41, 143, 127, 96, 216, 242, 222, 111, 10, 28, 63, 249, 171, 76, 16],
}
```

This is the first and only time it happened (e.g. not reproducible).

ok300:
Maybe related to the recent #452?

roeierez:
@dangeross I don't think it is related because the payment was succesfull.
@ok300 I suspect the lookup in the db for the payment after the sync failed, maybe because that payment wasn't retrieved from greenlight from some reason in the sync.

dangeross:
[...]

PR #469, opened 2023-09-20
Error codes

### LINKED ISSUE #459: Feature request: increase specificity of errors
To be possible to create translatable, user-oriented error messages, the SDK would need to return more specific errors.

Example: when paying a Bolt11 invoice using `send_payment()`, several things can go wrong such as:
* invoice was already paid
* invoice was issued by ourselves
* invoice expired
* invoice from different network (e.g. testnet)
* not a single route found
* all routes tried
* timeout
* etc

If it could make sense to provide information to the user about a specific failure reason, then there needs to be a specific error variant to represent it, as describing the error in a string will only be useful to developers integrating the SDK.

Of the previous error examples, some could be argued if it makes sense to inform the user or not (not a single route found vs all routes tried vs timeout), but others, I think it definitely makes sense to inform the user about (e.g. invoice already paid, self-payment, invoice expired).

I expanded on the invoice paying example, but it most likely applies in other places. In general, any error that the user could be interested in or could react to, there should be a error variant for it, and a different error message returned by the SDK isn't enough.

With more specific errors, we can even give our users some advice; for example, check your internet connection, check that the recipient is connected to LN and online, LSP is down, etc...

dangeross:
Structures the code in such a way as an error enum can be constructed per SDK interface method that needs one.

Accompanying PRs:
- [x] breez/breez-sdk-rn-generator#9
- [x] ~update c-breez error handling~

```
Err(SendPaymentError::AlreadyPaid)
```

Fixes #459

### DIFF
--- a/libs/sdk-core/src/backup.rs
+++ b/libs/sdk-core/src/backup.rs
@@ -1,5 +1,6 @@
 use crate::{
     breez_services::BackupFailedData,
+    error::SdkResult,
     persist::db::{HookEvent, SqliteStorage},
     BreezEvent, Config,
 };
@@ -63,8 +64,8 @@ pub struct BackupState {
 /// BackupTransport is the interface for syncing the sdk state between multiple apps.
 #[tonic::async_trait]
 pub trait BackupTransport: Send + Sync {
-    async fn pull(&self) -> Result<Option<BackupState>>;
-    async fn push(&self, version: Option<u64>, data: Vec<u8>) -> Result<u64>;
+    async fn pull(&self) -> SdkResult<Option<BackupState>>;
+    async fn push(&self, version: Option<u64>, data: Vec<u8>) -> SdkResult<u64>;
 }
 
 pub(crate) struct BackupWatcher {
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -4,26 +4,30 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
-use anyhow::{anyhow, ensure, Result};
+use anyhow::{anyhow, Result};
 use bip39::*;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::hashes::{sha256, Hash};
 use bitcoin::util::bip32::ChildNumber;
 use chrono::Local;
+use futures::TryFutureExt;
 use log::{LevelFilter, Metadata, Record};
 use serde_json::json;
 use tokio::sync::{mpsc, watch, Mutex};
 use tokio::time::sleep;
 use tonic::codegen::InterceptedService;
+use tonic::metadata::errors::InvalidMetadataValue;
 use tonic::metadata::{Ascii, MetadataValue};
 use tonic::service::Interceptor;
-use tonic::transport::{Channel, Uri};
+use tonic::transport::{Channel, Endpoint, Uri};
 use tonic::{Request, Status};
 
 use crate::backup::{BackupRequest, BackupTransport, BackupWatcher};
-use crate::boltzswap::BoltzApi;
 use crate::chain::{ChainService, MempoolSpace, Outspend, RecommendedFees};
-use crate::error::{SdkError, SdkResult};
+use crate::error::{
+    LnUrlAuthError, LnUrlPayError, LnUrlWithdrawError, ReceiveOnchainError, ReceivePaymentError,
+    SdkError, SdkResult, SendOnchainError, SendPaymentError,
+};
 use crate::fiat::{FiatCurrency, Rate};
 use crate::greenlight::{GLBackupTransport, Greenlight};
 use crate::grpc::channel_opener_client::ChannelOpenerClient;
@@ -43,14 +47,16 @@ use crate::lnurl::withdraw::validate_lnurl_withdraw;
 use crate::lsp::LspInformation;
 use crate::models::{
     parse_short_channel_id, ChannelState, ClosedChannelPaymentDetails, Config, EnvironmentType,
-    FiatAPI, LnUrlCallbackStatus, LspAPI, NodeAPI, NodeState, Payment, PaymentDetails, PaymentType,
+    FiatAPI, LnUrlCallbackStatus, LspAPI, NodeState, Payment, PaymentDetails, PaymentType,
     ReverseSwapPairInfo, ReverseSwapServiceAPI, SwapInfo, SwapperAPI,
     INVOICE_PAYMENT_FEE_EXPIRY_SECONDS,
 };
 use crate::moonpay::MoonPayApi;
+use crate::node_api::NodeAPI;
 use crate::persist::db::SqliteStorage;
-use crate::reverseswap::BTCSendSwap;
-use crate::swap::BTCReceiveSwap;
+use crate::swap_in::swap::BTCReceiveSwap;
+use crate::swap_out::boltzswap::BoltzApi;
+use crate::swap_out::reverseswap::BTCSendSwap;
 use crate::BuyBitcoinProvider::Moonpay;
 use crate::*;
 
@@ -192,12 +198,10 @@ impl BreezServices {
     ///
     /// It should be called only once when the app is started, regardless whether the app is sent to
     /// background and back.
-    async fn start(self: &Arc<BreezServices>) -> SdkResult<()> {
+    async fn start(self: &Arc<BreezServices>) -> Result<()> {
         let mut started = self.started.lock().await;
         if *started {
-            return Err(SdkError::InitFailed {
-                err: "BreezServices already started".into(),
-            });
+            return Err(anyhow!("BreezServices already started"));
         }
         let start = Instant::now();
         self.start_background_tasks().await?;
@@ -208,12 +212,18 @@ impl BreezServices {
     }
 
     /// Trigger the stopping of BreezServices background threads for this instance.
-    pub async fn disconnect(&self) -> Result<()> {
+    pub async fn disconnect(&self) -> SdkResult<()> {
         let mut started = self.started.lock().await;
         if !*started {
-            return Err(anyhow::Error::msg("BreezServices is not running"));
+            return Err(SdkError::Generic {
+                err: "BreezServices is not running".into(),
+            });
         }
-        self.shutdown_sender.send(()).map_err(anyhow::Error::msg)?;
+        self.shutdown_sender
+            .send(())
+            .map_err(|e| SdkError::Generic {
+                err: format!("Shutdown failed: {e}"),
+            })?;
         *started = false;
         Ok(())
     }
@@ -222,7 +232,10 @@ impl BreezServices {
     ///
     /// Calling `send_payment` ensures that the payment is not already completed, if so it will result in an error.
     /// If the invoice doesn't specify an amount, the amount is taken from the `amount_msat` arg.
-    pub async fn send_payment(&self, req: SendPaymentRequest) -> SdkResult<SendPaymentResponse> {
+    pub async fn send_payment(
+        &self,
+        req: SendPaymentRequest,
+    ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
         let invoice_amount_msat = parsed_invoice.amount_msat.unwrap_or_default();
@@ -230,15 +243,15 @@ impl BreezServices {
 
         // Ensure amount is provided for zero invoice
         if provided_amount_msat == 0 && invoice_amount_msat == 0 {
-            return Err(SdkError::SendPaymentFailed {
-                err: "amount must be provided when paying a zero invoice".into(),
+            return Err(SendPaymentError::InvalidAmount {
+                err: "Amount must be provided when paying a zero invoice".into(),
             });
         }
 
         // Ensure amount is not provided for invoice that contains amount
         if provided_amount_msat > 0 && invoice_amount_msat > 0 {
-            return Err(SdkError::SendPaymentFailed {
-                err: "amount should not be provided when paying a non zero invoice".into(),
+            return Err(SendPaymentError::InvalidAmount {
+                err: "Amount should not be provided when paying a non zero invoice".into(),
             });
         }
 
@@ -246,13 +259,12 @@ impl BreezServices {
             .persister
             .get_completed_payment_by_hash(&parsed_invoice.payment_hash)?
         {
-            Some(_) => Err(SdkError::SendPaymentFailed {
-                err: "Invoice already paid".into(),
-            }),
+            Some(_) => Err(SendPaymentError::AlreadyPaid),
             None => {
                 let payment_res = self
                     .node_api
                     .send_payment(req.bolt11.clone(), req.amount_msat)
+                    .map_err(Into::into)
                     .await;
                 let payment = self
                     .on_payment_completed(
@@ -270,11 +282,12 @@ impl BreezServices {
     pub async fn send_spontaneous_payment(
         &self,
         req: SendSpontaneousPaymentRequest,
-    ) -> SdkResult<SendPaymentResponse> {
+    ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
         let payment_res = self
             .node_api
             .send_spontaneous_payment(req.node_id.clone(), req.amount_msat)
+            .map_err(Into::into)
             .await;
         let payment = self
             .on_payment_completed(req.node_id, None, payment_res)
@@ -290,7 +303,7 @@ impl BreezServices {
     /// is made.
     ///
     /// This method will return an [anyhow::Error] when any validation check fails.
-    pub async fn lnurl_pay(&self, req: LnUrlPayRequest) -> Result<LnUrlPayResult> {
+    pub async fn lnurl_pay(&self, req: LnUrlPayRequest) -> Result<LnUrlPayResult, LnUrlPayError> {
         match validate_lnurl_pay(req.amount_msat, req.comment, req.data.clone()).await? {
             ValidatedCallbackResponse::EndpointError { data: e } => {
                 Ok(LnUrlPayResult::EndpointError { data: e })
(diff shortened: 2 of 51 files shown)


roeierez (approved):
Really like it! LGTM

dangeross:
FYI, it seems when using the interface as an error the err field is not mapped to the bindings Error(message), but instead extends the ExtendedError(code, err) : Error(). So the Error message is not set. This is the case at least for swift and kotlin. I'm looking into it

https://github.com/mozilla/uniffi-rs/issues/1434

libs/sdk-react-native/example/App.js
```diff
@@ -145,6 +145,7 @@ const App = () => {
                 addLine("backupStatus", JSON.stringify(await backupStatus()))
             } catch (e) {
                 addLine("error", e.toString())
+                console.log(`Error: ${JSON.stringify(e)}`)
```
JssDWt, line 148: How would you handle this error here in real life?
I guess you don't really know whether it's a SendPaymentFailed error or something else here?
What would the logic to do something based on the error type look like?

libs/sdk-core/src/error.rs
```diff
@@ -30,8 +47,11 @@ pub enum SdkError {
     #[error("Failed to receive payment: {err}")]
     ReceivePaymentFailed { err: String },
 
-    #[error("Failed to send payment: {err}")]
-    SendPaymentFailed { err: String },
+    #[error("Failed to send payment: {code} {err}")]
+    SendPaymentFailed {
+        code: SendPaymentFailedCode,
+        err: String,
+    },
```
ok300, line 54: Why do we need the error message in addition to the error code?

Now we use the error message because the error kinds are very broad, so the message helps clarify the sub-type of error:

```rust
Err(SdkError::SendPaymentFailed {
    err: "Invoice already paid".into(),
}),
```

But if we add error codes, the message becomes superfluous, no?

```rust
Err(SdkError::SendPaymentFailed {
    code: SendPaymentFailedCode::AlreadyPaid,
    err: "Invoice already paid".into(),
}),
```

If the message is always meant to stay the same, and we want to include it for logging or display purposes, then perhaps a constructor for every error kind can help:

```rust
impl SdkError {
    pub fn send_payment_failed() -> Self {
        SdkError::SendPaymentFailed {
            code: SendPaymentFailedCode::AlreadyPaid,
            err: "Invoice already paid".into(),
        }
    }
}
```

so we can just call these methods when we have to instantiate an error type:

```rust
Err(SdkError::send_payment_failed())
```

libs/sdk-react-native/example/App.js
dangeross: It's a bit more difficult in RN as you only have the option to pass a string code and message. The error e would look something like:
```
{
    code: "SendPaymentFailed.AlreadyPaid",
    message: "Invoice already paid",
    ...
}
```
You would probably map the error code to translation key to get the correct language for an error message, or do some feedback/logic based on the code key.

libs/sdk-core/src/error.rs
dangeross: The message can be used to add extra context for debugging etc.

```
SdkError::PaymentError {
    code: PaymentErrorCode::DecodingFailed,
    err: format!("Failed to decode hex payment hash: {e}"),
}

SdkError::PaymentError {
    code: PaymentErrorCode::DecodingFailed,
    err: format!("Failed to decode hex payee pubkey: {e}"),
}
```

Good idea with the constructor for repeated errors 👍

libs/sdk-react-native/example/App.js
dangeross: It's not possible to add extra fields into the promise reject from native android/ios to typescript 😒 

Here's the interface:
```
reject(code, message, e)
```

And the result:
```
{
  "code": "ServiceError.Generic",
  "message": "No LSP found for id 03cea51f-b654-4fb0-8e82-eca137f236a0",
  "domain": "BreezSDK.SdkError",
  "userInfo": {},
  "nativeStackIOS": [
    "0   BreezSDKExample                     0x000000011003faf0 RCTJSErrorFromCodeMessageAndNSError + 112",
    "1   BreezSDKExample                     0x000000010ffd1b1a __41-[RCTModuleMethod processMethodSignature]_block_invoke_2.129 + 170",
    "2   BreezSDKExample                     0x000000010fa99b17 $sSo8NSStringCSgACSo7NSErrorCSgIeyByyy_SSSgAGs5Error_pSgIegggg_TR + 359",
    "3   BreezSDKExample                     0x000000010faabde6 $s8BreezSDK07RNBreezB0C9rejectErr3err0D0ys5Error_p_ySSSg_AHsAG_pSgtctF + 4710",
    "4   BreezSDKExample                     0x000000010faa5dee $s8BreezSDK07RNBreezB0C7lspInfo_6rejectyyypSgc_ySSSg_AGs5Error_pSgtctF + 638",
    "5   BreezSDKExample                     0x000000010faa6032 $s8BreezSDK07RNBreezB0C7lspInfo_6rejectyyypSgc_ySSSg_AGs5Error_pSgtctFTo + 162",
    "6   CoreFoundation                      0x00007ff80042f2fc __invoking___ + 140",
    "7   CoreFoundation                      0x00007ff80042c529 -[NSInvocation invoke] + 305",
    "8   CoreFoundation                      0x00007ff80042c7bc -[NSInvocation invokeWithTarget:] + 70",
    "9   BreezSDKExample                     0x000000010ffd3536 -[RCTModuleMethod invokeWithBridge:module:arguments:] + 2534",
    "10  BreezSDKExample                     0x000000010ffd79da _ZN8facebook5reactL11invokeInnerEP9RCTBridgeP13RCTModuleDatajRKN5folly7dynamicEiN12_GLOBAL__N_117SchedulingContextE + 1402",
    "11  BreezSDKExample                     0x000000010ffd728c _ZZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEiENK3$_0clEv + 156",
    "12  BreezSDKExample                     0x000000010ffd71e9 ___ZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEi_block_invoke + 25",
    "13  libdispatch.dylib                   0x0000000115fa8d18 _dispatch_call_block_and_release + 12",
    "14  libdispatch.dylib                   0x0000000115fa9f5b _dispatch_client_callout + 8",
    "15  libdispatch.dylib                   0x0000000115fb1d41 _dispatch_lane_serial_drain + 1064",
    "16  libdispatch.dylib                   0x0000000115fb2b1b _dispatch_lane_invoke + 441",
    "17  libdispatch.dylib                   0x0000000115fc0f13 _dispatch_workloop_worker_thread + 959",
    "18  libsystem_pthread.dylib             0x00007ff836172ce3 _pthread_wqthread + 326",
    "19  libsystem_pthread.dylib             0x00007ff836171c67 start_wqthread + 15"
  ]
}
```

JssDWt: OK, I guess nothing can be done about that. This will work too!

libs/sdk-core/src/error.rs
```diff
@@ -1,60 +1,311 @@
...
+
+impl fmt::Display for ValueErrorCode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
 /// Type of error returned by the SDK
 #[derive(Debug, Error)]
 pub enum SdkError {
```
ok300, line 111: NIT: Enum values in alphabetic order

libs/sdk-core/src/error.rs
```diff
@@ -1,60 +1,311 @@
...
+    PaymentError { code: PaymentErrorCode, err: String },
+
+    #[error("Onchain error [{code}]: {err}")]
+    OnchainError { code: OnchainErrorCode, err: String },
+
+    #[error("Value error [{code}]: {err}")]
+    ValueError { code: ValueErrorCode, err: String },
+}
+
+impl SdkError {
```
ok300, line 128: NIT: constructor methods in alphabetic order, to make it easier to search or add new ones.

libs/sdk-core/src/error.rs
```diff
@@ -1,60 +1,311 @@
...
     fn from(value: anyhow::Error) -> Self {
-        Self::Generic {
-            err: value.to_string(),
-        }
+        Self::generic(value.to_string())
     }
 }
+
+#[macro_export]
+macro_rules! ensure {
```
ok300, line 305: Good idea!

NIT: call it `ensure_sdk!` to avoid confusion with `anyhow::ensure!`

libs/sdk-core/src/reverseswap.rs
```diff
@@ -26,6 +27,16 @@ use tokio::time::{sleep, Duration};
 pub const ESTIMATED_CLAIM_TX_VSIZE: u64 = 138;
 pub const ESTIMATED_LOCKUP_TX_VSIZE: u64 = 153;
 
+impl SdkError {
```
ok300, line 30: NIT: might be better to keep all `impl SdkError`s in the same place, for a better overview?

libs/sdk-core/src/swap.rs
```diff
@@ -22,6 +23,16 @@ use ripemd::{Digest, Ripemd160};
 use crate::breez_services::{BreezEvent, BreezServer, PaymentReceiver, Receiver};
 use crate::models::{Swap, SwapInfo, SwapStatus, SwapperAPI};
 
+impl SdkError {
```
ok300, line 26: See above

libs/sdk-core/src/error.rs
ok300, line 128: NIT: Can the methods that take `err: String` as argument be changed to take instead `err: &str` and do the conversion to `String` internally?

This would make calling them easier, because we would avoid calling `to_string()` every time we give in an error message.

```rust
ensure_sdk!(condition, SdkError::generic("Danger error"));
```

If we happen to have a `String`, we can then give it as a reference

```rust
let error_msg: String = "Some message".
ensure_sdk!(condition, SdkError::generic(&error_msg));
```

ok300 (commented):
Added a few suggestions, but otherwise LGTM!

libs/sdk-core/src/error.rs
dangeross: Done [4edbbbe](https://github.com/breez/breez-sdk/pull/469/commits/4edbbbee20fdca35aebe326f3048148e0f9c7b48)

libs/sdk-core/src/error.rs
dangeross: Done [4edbbbe](https://github.com/breez/breez-sdk/pull/469/commits/4edbbbee20fdca35aebe326f3048148e0f9c7b48)

libs/sdk-core/src/error.rs
dangeross: Done [4edbbbe](https://github.com/breez/breez-sdk/pull/469/commits/4edbbbee20fdca35aebe326f3048148e0f9c7b48)

libs/sdk-core/src/reverseswap.rs
dangeross: Done [4edbbbe](https://github.com/breez/breez-sdk/pull/469/commits/4edbbbee20fdca35aebe326f3048148e0f9c7b48)

libs/sdk-core/src/error.rs
dangeross: Done [4edbbbe](https://github.com/breez/breez-sdk/pull/469/commits/4edbbbee20fdca35aebe326f3048148e0f9c7b48). In most cases Strings are now referenced like below. Let me know if there is a more rusty way to do it.
```
SdkError::lsp_error(LspErrorCode::ConnectionFailed, &e.to_string())
SdkError::lsp_not_found(&format!("Unknown LSP: {lsp_id}"))
```

ok300: There might be fancier ways, but this is very good!

ok300 (approved):

JssDWt (commented):
I went over the LSP and payment error codes. Will check the others later.

I think as a rule of thumb, the error code should mean something in the domain. So too small error codes should not propagate to the client. Most error codes defined here definitely match that definition!

Payment error codes should eventually be more specific. We should eventually extract the error reason from the greenlight response to give the user specific information why the payment failed.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1172,9 +1199,7 @@ impl BreezServices {
                 .create(true)
                 .append(true)
                 .open(format!("{log_dir}/sdk.log"))
-                .map_err(|_| SdkError::InitFailed {
-                    err: "Can't create log file".into(),
-                })?,
+                .map_err(|_| SdkError::init_failed("Can't create log file"))?,
```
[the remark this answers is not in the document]
dangeross: What do you mean by too small error codes? Too generic or too low level?

libs/sdk-core/src/error.rs
```diff
@@ -1,60 +1,342 @@
...
+}
+
+/// The payment error codes
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+pub enum PaymentErrorCode {
+    AlreadyPaid,
+    ChannelNotFound,
+    InvalidAmount,
+    OpeningFeeParamsNotFound,
+    PaymentFailed,
```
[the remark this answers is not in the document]
dangeross: I wanted to find an intuitive way to get the CLN payment error from the RPC response, but tonic doesn't play well with extended response error codes. I could parse the error code from the error message, but it seems hacky.
```"Error calling method Pay: RpcError { code: Some(207), message: \"Invoice expired\" }"```

File: libs/sdk-core/src/breez_services.rs [hunk 10]
```diff
@@ -802,9 +831,10 @@ impl BreezServices {
                         .await?;
                     Ok(p)
                 }
-                None => Err(SdkError::SendPaymentFailed {
-                    err: "Payment not found".into(),
-                }),
+                None => Err(SdkError::payment_error(
+                    PaymentErrorCode::PaymentNotFound,
```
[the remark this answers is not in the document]
dangeross: I think it's a specific edge case that could be handled better by a client with an error code. There is even an open issue about it #458. Better to be more informative then less.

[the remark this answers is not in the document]
JssDWt: I agree that it's good to be informative, but this error specifically is about an SDK bug. It doesn't really have anything to do with paying, or how to respond to a payment.

libs/sdk-core/src/breez_services.rs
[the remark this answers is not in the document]
dangeross: Agreed. I've removed some error codes. I think there is a balance we have to find, to be informative to both users and clients, to allow actions to be taken based on the error code.

libs/sdk-core/src/lsp.rs
```diff
@@ -85,14 +85,15 @@ impl LspInformation {
                     }
                 }
                 Err(e) => {
-                    return Err(SdkError::CalculateOpenChannelFeesFailed { err: e.to_string() })
+                    return Err(SdkError::lsp_error(
+                        LspErrorCode::CalculateOpenChannelFeesFailed,
```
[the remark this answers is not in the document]
dangeross: Done [9a0cbb6](https://github.com/breez/breez-sdk/pull/469/commits/9a0cbb6dcabb959799e3854c071d021a0bd04b42)

libs/sdk-core/src/error.rs
```diff
@@ -1,60 +1,342 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+/// The payment error codes
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+pub enum PaymentErrorCode {
+    AlreadyPaid,
+    ChannelNotFound,
```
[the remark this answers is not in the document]
dangeross: Done [9a0cbb6](https://github.com/breez/breez-sdk/pull/469/commits/9a0cbb6dcabb959799e3854c071d021a0bd04b42)

libs/sdk-core/src/error.rs
```diff
@@ -1,60 +1,342 @@
...
+    InvalidAmount,
+    OpeningFeeParamsNotFound,
```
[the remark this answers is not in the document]
dangeross: Done [9a0cbb6](https://github.com/breez/breez-sdk/pull/469/commits/9a0cbb6dcabb959799e3854c071d021a0bd04b42)

libs/sdk-core/src/breez_services.rs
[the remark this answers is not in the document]
dangeross: Done [9a0cbb6](https://github.com/breez/breez-sdk/pull/469/commits/9a0cbb6dcabb959799e3854c071d021a0bd04b42)

libs/sdk-core/src/error.rs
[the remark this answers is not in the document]
JssDWt: It's the cln grpc plugin that literally returns this message, so the tonic error code is actually unknown. So I guess eventually we'll have to either parse the message, or change the returned codes in the cln-grpc plugin.
https://github.com/ElementsProject/lightning/blob/9074341ef6a7afb3f94be68fad22b948f7e42d97/cln-grpc/src/server.rs#L748

libs/sdk-core/src/greenlight/error.rs
dangeross, line 1: @ok300 is the right place to add this enum and function to parse CLN RPC errors?

libs/sdk-core/src/error.rs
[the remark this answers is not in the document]
dangeross: I've added parsing of this error message, so several cln grpc errors now map to specific error codes like InvoiceExpired, RouteNotFound or PaymentTimeout

dangeross:
Fixes #336

libs/sdk-core/src/greenlight/error.rs
ok300: Yes this is good :+1:

libs/sdk-core/src/backup.rs
```diff
@@ -63,8 +64,8 @@ pub struct BackupState {
 /// BackupTransport is the interface for syncing the sdk state between multiple apps.
 #[tonic::async_trait]
 pub trait BackupTransport: Send + Sync {
-    async fn pull(&self) -> Result<Option<BackupState>>;
-    async fn push(&self, version: Option<u64>, data: Vec<u8>) -> Result<u64>;
+    async fn pull(&self) -> NodeResult<Option<BackupState>>;
+    async fn push(&self, version: Option<u64>, data: Vec<u8>) -> NodeResult<u64>;
```
roeierez, line 68: This trait should be detached from the node api. We do implement it using greenlight today but it can be easily extended to use external service (Gdrive, icloud, remote server, etc...).
My main comment here is that I think it shouldn't return NodeResult. It seems to me that backup module should have its own errors (or Generic).

roeierez (changes requested):

JssDWt (approved):
LGTM!

PR #470, opened 2023-09-21
Lsps0 implementation

JssDWt:
Adds a basic LSPS0 client implementation. With support for request/response messages to the LSPS0 server, and support for receiving LSPS0 notifications (this is a different kind of notification than the notification in our offline receive milestone).

The idea is:
- Every user node (greenlight node) initializes a single `lsps0::Transport`. You start the transport in a thread. That makes it start listening to LSPS0 messages from any remote node, and allows sending request/response messages to any remote node.
- You initialize a `lsps0::Client` to interact with a specific remote node. There can be many `lsps0::Client`s.

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -1,4 +1,5 @@
 use std::cmp::{max, min};
+use std::pin::Pin;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
@@ -16,7 +17,8 @@ use gl_client::pb::cln::listinvoices_invoices::ListinvoicesInvoicesStatus;
 use gl_client::pb::cln::listpays_pays::ListpaysPaysStatus;
 use gl_client::pb::cln::{
     self, CloseRequest, ListclosedchannelsClosedchannels, ListclosedchannelsRequest,
-    ListinvoicesInvoices, ListpaysPays, ListpeerchannelsRequest, StaticbackupRequest,
+    ListinvoicesInvoices, ListpaysPays, ListpeerchannelsRequest, SendcustommsgRequest,
+    StaticbackupRequest,
 };
 use gl_client::pb::cln::{AmountOrAny, InvoiceRequest};
 use gl_client::pb::{Amount, OffChainPayment, PayStatus, Peer, WithdrawResponse};
@@ -32,12 +34,14 @@ use serde::{Deserialize, Serialize};
 use strum_macros::{Display, EnumString};
 use tokio::sync::{mpsc, Mutex};
 use tokio::time::sleep;
+use tokio_stream::{Stream, StreamExt};
 use tonic::Streaming;
 
 use crate::invoice::parse_invoice;
 use crate::models::*;
 use crate::persist::db::SqliteStorage;
 use crate::{Channel, ChannelState, NodeConfig};
+use std::iter::Iterator;
 
 const MAX_PAYMENT_AMOUNT_MSAT: u64 = 4294967000;
 const MAX_INBOUND_LIQUIDITY_MSAT: u64 = 4000000000;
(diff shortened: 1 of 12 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -810,6 +811,69 @@ impl NodeAPI for Greenlight {
...
+        // Copy the greenlight stream to the mpsc::channel to map to our
+        // internal CustomMessage struct.
+        tokio::spawn(async move {
+            loop {
+                let msg = match stream.message().await {
+                    Ok(m) => match m {
+                        Some(m) => m,
+                        None => return,
+                    },
+                    Err(_) => return,
```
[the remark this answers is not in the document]
roeierez: I think the way we fetch messages in a reliable way from the custom message stream could be similar to the way we do it for the invoice stream (stream_incoming). We have a task that wraps the call to the stream_incoming and whenever there is an error it sleeps some delay and then create the stream again (track_invoices).

libs/sdk-core/src/lsps0/jsonrpc.rs
```diff
@@ -0,0 +1,99 @@
...
+use serde_json::Value;
+
+#[derive(Debug)]
+pub enum JsonRpcServerObject {
+    Notification(JsonRpcNotification),
+    Response(JsonRpcResponse),
+    Error(JsonRpcError),
+}
+
+impl<'de> Deserialize<'de> for JsonRpcServerObject {
```
roeierez, line 15: Do we need the Deserialize implementation? serde is pretty flexible in deserialization enums (https://serde.rs/enum-representations.html#untagged)

libs/sdk-core/src/lsps0/transport.rs
```diff
@@ -0,0 +1,677 @@
...
+                None => break Err(anyhow!("custom message stream ended.")),
+            };
+            if msg.message_type != LSPS0_MESSAGE_TYPE {
+                continue;
+            }
+
+            let v: JsonRpcServerObject = match serde_json::from_slice(&msg.payload) {
+                Ok(v) => v,
+                Err(e) => {
+                    debug!(
```
roeierez, line 89: perhaps `warn` or `error` ?

libs/sdk-core/src/lsps0/transport.rs
```diff
@@ -0,0 +1,677 @@
...
+            if jsonrpc_version != JSONRPC_VERSION {
+                debug!("error deserializing lsps0 payload {:?}: Invalid jsonrpc version. Expected {:?}.", &msg.payload, JSONRPC_VERSION);
+                continue;
+            }
+
+            match v {
+                JsonRpcServerObject::Notification(n) => {
+                    let id = get_notification_handler_id(&n.method, msg.peer_id.clone());
+                    if let Some(tx) = (*self.notification_handlers.read().await).get(&id) {
+                        match tx.send(n).await {
```
roeierez, line 112: The send is inside the lock, perhaps it is better to release the lock immediately after we finished reading from the notification_handlers.

libs/sdk-core/src/lsps0/transport.rs
```diff
@@ -0,0 +1,677 @@
...
+}
+
+fn get_request_handler_id(request_id: &str, node_id: Vec<u8>) -> String {
+    let mut id = hex::encode(node_id);
+    id.push('|');
+    id.push_str(request_id);
+    id
+}
+
+fn get_notification_handler_id(method: &str, node_id: Vec<u8>) -> String {
```
roeierez, line 250: Shouldn't we further scope the notification handler by a random request id? I might be missing something but it looks like that if the client invokes the same method in parallel  the later invocation will replace the the entry in the response_handlers of the former that will hang as a result?

roeierez (changes requested):
Overall architecture is right on spot IMO. Dropped some comments/improvments.

libs/sdk-core/src/lsps0/transport.rs
JssDWt: It works like that for request/response. Notifications don't have an id, just a method. Notifications are 'streaming' from the server side.

libs/sdk-core/src/lsps0/jsonrpc.rs
JssDWt: Works with untagged. That example in the link is literally jsonrpc ^^

libs/sdk-core/src/lsps0/transport.rs
JssDWt: Isn't the lock out of scope already here?

libs/sdk-core/src/lsps0/transport.rs
roeierez: Great! I guess I confused it.

libs/sdk-core/src/lsps0/transport.rs
roeierez: Yes you are right , the temporary variable is release.

libs/sdk-core/src/lsps0/transport.rs
```diff
@@ -0,0 +1,677 @@
...
+                Err(e) => {
+                    debug!(
+                        "error deserializing lsps0 payload {:?}: {}",
+                        &msg.payload, e
+                    );
+                    continue;
+                }
+            };
+
+            let jsonrpc_version = match &v {
```
roeierez, line 97: This match scope hints that perhaps we need an outer structure with the common fields?
For example:

```
struct JsonPayload {
  jsonrpc: string,

  #[serde(flatten)]
  object: JSONRpcServerObject,
}
```

The flatten means the fields are embedded.
Only if it makes sense, didn't think about it all the way through.

JssDWt: Makes sense, done

libs/sdk-core/src/lsps0/transport.rs
JssDWt: Changed most of the deserialization errors to warn.

libs/sdk-core/src/lsps0/jsonrpc.rs
JssDWt: Done

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
JssDWt: Done. I had in mind to put that retry logic in the caller. But it indeed makes more sense here.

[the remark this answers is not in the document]
roeierez: Looks good now. 
The only issue I see now is regarding the tokio task that is running here and is not controlled in a sense that there is no code that stops it.
Do we really need a task here? (and channel to communicate?).
Since the caller (transport) already running in a task and call this method from within the task it would look cleaner if this method returns a stream of custom message by transforming the gRpc stream without background processing.
This way the caller is responsible for running it in the background (or not) and we don't need to worry about managing this task.
What do you think?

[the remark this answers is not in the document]
JssDWt: For some reason I totally misplaced this comment, and thought it was about `transport.rs`. Anyway, good points.

I think it would be great to be able to map that stream into another stream while reading it. Like `stream.map(..)`, or some kind of Generator, yield return kind of pattern. I couldn't find any though.

It is sort-of stopped, in the sense that when the receiver drops, the method returns, so the task is completed. Also, when the sender drops (silently return), the receiver will get an empty message, therefore it knows the stream disconnected. There could be useful info in the error, it should probably be logged. 

I couldn't find any good way to map the one stream to the other without usnig the task, or blocking the caller, which I don't like. Also, you need to return the receiver, so you can't infinitely loop here without returning something?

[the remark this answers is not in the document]
roeierez: tonic streaming implements tokio::Stream https://github.com/hyperium/tonic/blob/master/tonic/src/codec/decode.rs#L384

So perhaps we can use tokio stream mapping to return the steam with our own model? https://tokio.rs/tokio/tutorial/streams ,(see adapters section)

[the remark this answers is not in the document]
JssDWt: Updated to map the stream with `tokio_stream`. Can you check again?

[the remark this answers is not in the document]
roeierez: That looks great!

roeierez (approved):
LGTM!

PR #475, opened 2023-09-21
Logging: allow modules to log on trace

ok300:
Our logging framework was only considering log levels at `debug` and up. This PR changes that to allow `trace` as well.

No modules currently log at `trace`, but with this PR, they could and their statements would land in the SDK log.

### DIFF
--- a/libs/sdk-bindings/src/uniffi_binding.rs
+++ b/libs/sdk-bindings/src/uniffi_binding.rs
@@ -42,7 +42,7 @@ impl BindingLogger {
 impl log::Log for BindingLogger {
     fn enabled(&self, m: &Metadata) -> bool {
         // ignore the internal uniffi log to prevent infinite loop.
-        return m.level() <= Level::Debug && *m.target() != *"breez_sdk_bindings::uniffi_binding";
+        return m.level() <= Level::Trace && *m.target() != *"breez_sdk_bindings::uniffi_binding";
     }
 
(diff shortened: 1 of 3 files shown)


JssDWt:
Does this exclude the super low-level frames that flood the logs when you enable trace?

ok300:
Yes, enabling `trace` at the root logger still respects the individual submodule logger levels, and none of them is at `trace` now:

https://github.com/breez/breez-sdk/blob/f76fcb1c20b0bfa42e2b0b5d8e9ea00156385c57/libs/sdk-core/src/breez_services.rs#L1160-L1176

JssDWt (approved):

roeierez:
@ok300 it would be useful to also include by default the signer logs and the greenlight node logs in a better verbosity. I would try to move them to trace and see if we don't get too much info. This is very usefull in support and investigation of issues.

roeierez (approved):

ok300:
@roeierez I tried the `trace` level on the VLS modules, but it gave about 5 MB of logs (>50k lines) for a single payment.

Maybe its better to keep it as it is, and only increase logging when needed?

roeierez:
> @roeierez I tried the `trace` level on the VLS modules, but it gave about 5 MB of logs (>50k lines) for a single payment.
> 
> Maybe its better to keep it as it is, and only increase logging when needed?

Yes, I agree

PR #477, opened 2023-09-21
Fix SDK panic caused by closing channels

ok300:
Calling `close_lsp_channels` would trigger a panic when trying to parse the mempool result, because the outspends do not always have `status` and `vin`.


roeierez (approved):
LGTM

PR #478, opened 2023-09-22
Expose txid in sweep bindings

### LINKED ISSUE #476: Feature request: return `txid` on on-chain sweep
It would be useful if the `txid` would be returned when sweeping the on-chain funds, so that it can be displayed to the user for tracking purposes.

dleutenegger:
Fixes #476

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -520,6 +520,10 @@ enum BuyBitcoinProvider {
     "Moonpay",
 };
 
+dictionary SweepResponse {
+    sequence<u8> txid;
+};
+
 interface BlockingBreezServices {   
 
    [Throws=SdkError]
@@ -565,7 +569,7 @@ interface BlockingBreezServices {
    sequence<Payment> list_payments(ListPaymentsRequest request);
 
    [Throws=SdkError]
-   void sweep(string to_address, u64 fee_rate_sats_per_vbyte);
+   SweepResponse sweep(string to_address, u64 fee_rate_sats_per_vbyte);
 
    [Throws=SdkError]
    sequence<Rate> fetch_fiat_rates();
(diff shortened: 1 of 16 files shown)


ok300:
Looks good!

The only question I have is around the React Native bindings, as I see a lot of formatting changes. Did you generate the bindings using the [RN generator](https://github.com/breez/breez-sdk-rn-generator)? 

And if yes, what are your `ktlint` and `kotlin` versions?

```
ktlint --version
kotlin -version
```

dleutenegger:
> Looks good!
> 
> The only question I have is around the React Native bindings, as I see a lot of formatting changes. Did you generate the bindings using the [RN generator](https://github.com/breez/breez-sdk-rn-generator)?
> 
> And if yes, what are your `ktlint` and `kotlin` versions?
> 
> ```
> ktlint --version
> kotlin -version
> ```

Must have missed the warnings. Committed the new files with the linting now.

```
kotlin => 1.9.0
ktlint => 1.0.0
swiftformat => 0.52.4
```

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -554,7 +558,7 @@ interface BlockingBreezServices {
    sequence<Payment> list_payments(ListPaymentsRequest request);
 
    [Throws=SdkError]
-   void sweep(string to_address, u64 fee_rate_sats_per_vbyte);
+   SweepResponse sweep(string to_address, u64 fee_rate_sats_per_vbyte);
```
roeierez, line 561: Perhaps also use SweepRequest as argument that wraps all these arguments. I know it is not part of this PR but since we are changing the interface here anyway it might worth doing it in one shot.

roeierez (changes requested):
Looks great. I added one request if not too hard.
Thank you for that!

ok300 (approved):
Looks good!

PR #479, opened 2023-09-23
workaround for x86_64-android

JssDWt:
A workaround that fixes https://github.com/breez/c-breez/issues/553
Recommended by @yukibtc here https://github.com/breez/c-breez/issues/553#issuecomment-1731477289

Flutter uses sdk-core for Android, that's why this build script is put in sdk-core. It may be necessary to put it in sdk-bindings as well, for bindings for other languages that work on android, not sure.

Did a testrun with a breez-sdk-flutter release from the CI: https://github.com/breez/breez-sdk/suites/16489120334/artifacts/942132999
The testrun worked.

PR #480, opened 2023-09-25
Use flutter sdk-bindings

JssDWt:
Use sdk-bindings instead of sdk-core for the flutter binaries, for consistency. This makes all bindings use the sdk-bindings binary.

Tried on my breez sdk example flutter project.


erdemyerebasmaz (approved):
Changes look good!

Why don't we target `x86_64-linker` branch and address https://github.com/breez/breez-sdk/pull/479#issuecomment-1733692875 in that PR as well?

roeierez (approved):
LGTM

erdemyerebasmaz:
Follow up question;

Do we need the `makefile` on sdk-core anymore?

@roeierez @JssDWt @ok300

PR #482, opened 2023-09-25
Simple lsps2 client

JssDWt:
A simple lsps2 client, just a wrapper for the rpc functions defined in the LSPS2 spec. Depends on https://github.com/breez/breez-sdk/pull/470


roeierez (approved):
LGTM

PR #486, opened 2023-09-26
fix flacky tests

Torakushi:
We added some flacky tests when testing ln_url callback.
The problem was that we ran the tests in parallel and possibly mocked the same URL twice (leading to interference)

### DIFF
--- a/libs/sdk-core/src/input_parser.rs
+++ b/libs/sdk-core/src/input_parser.rs
@@ -621,7 +621,10 @@ pub(crate) mod tests {
     use crate::input_parser::*;
     use crate::models::Network;
 
-    /// Mock server used in tests
+    /// Mock server used in tests. As the server is shared between tests,
+    /// we should not mock the same url twice with two different outputs,
+    /// one way to do so is to add a random string that will be a differentiator
+    /// in the URL.
     pub(crate) static MOCK_HTTP_SERVER: Lazy<Mutex<ServerGuard>> =
         Lazy::new(|| Mutex::new(Server::new()));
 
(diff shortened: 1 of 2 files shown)


ok300 (changes requested):
Looks good.

The only thing remaining is to fix the `cargo clippy` warnings.

Torakushi:
@ok300 Clippy wasn't happy with functions with too much arguments

I've created two structures that will be passed as argument to our different functions

ok300 (approved):

ok300 (changes requested):
I've ran this a few times and about once every 5-10 runs, I get

```
failures:

---- input_parser::tests::test_lnurl_withdraw_lud_17_error stdout ----
Error: Unrecognized input type
```

There must be something about the order of tests, or which tests get executed in parallel, that causes this.

ok300:
I mean, in principle, the tests are not flaky, because they run well one test at a time:

```
cargo test -- --test-threads=1
```

AFAIK the CI is also doing that, by reserving only 1 CPU to the workflow that calls `cargo test`.

So there is always the option of keeping it as it is, and running the tests sequentially.

Torakushi:
Yes, that's true, it's only when we run tests in parallel.

The problem with `cargo test -- --test-threads=1` is that if we have many more tests, it will take forever. It's okay for the CI, but for local development it can be annoying :/ (But it's not really annoying at the moment, so it depends on what we want!)

Torakushi:
So, I solved the concurrency problem in the tests. Essentially, when testing an error case, we don't need to provide a "real" url as we will not decode the lnurl to check if the url matches.
I added a comment for the HTTP_MOCK_SERVER to warn against this problem.

PS: Sharing a mock server might not be ideal because of this. Mocking the http server directly can be a solution (like Jesse suggested).

JssDWt:
> Sharing a mock server might not be ideal because of this. Mocking the http server directly can be a solution (like Jesse suggested).

Yeah I suggested mocking the http client actually, so you can inspect the request in the test, rather than having a shared server.

Torakushi:
Yes sorry for the typo!
`Mocking the http server directly --> Mocking the http client directly`

JssDWt (approved):

JssDWt:
@ok300 review reminder

ok300 (approved):

PR #488, opened 2023-09-27
Fix issue with remote to local sync, affecting backup

ok300:
When optimistic sync fails, we try a bidirectional sync, of which one step is to sync from remote to local.

I found that this particular tx failed because the remote schema did not have the new column, which caused the entire sync to abort by throwing the error.

The fix is to not throw the error if this discrepancy happens. The schema is apparently synchronized after this tx is executed, so on the next `backup` this tx won't throw an error.

I admit I don't fully understand yet the backup mechanism, so the explanation above may be incorrect. Therefore I'm marking this as draft for now.

If the explanation and solution makes sense, we should apply the same change to all `tx.commit(..)` that fall under the `if to_local` check.

### DIFF
--- a/libs/sdk-core/src/persist/sync.rs
+++ b/libs/sdk-core/src/persist/sync.rs
@@ -90,7 +90,7 @@ impl SqliteStorage {
     ) -> Result<()> {
         let sync_data_file = remote_storage.sync_db_path();
         match SqliteStorage::migrate_sync_db(sync_data_file.clone()) {
-            Ok(_) => {}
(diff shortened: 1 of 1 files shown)


roeierez:
The error is thrown since the schema is not synchornized which could only happen if one ran some migrations out of order.
This, as far as I know, will only happen to us testing migrations and should not happen to someone that ran the migrations in the right order.
I think if some inconsistency happens then the backup should fail. I don't see how it fixes itself as you mentioned in the description.

PR #489, opened 2023-09-28
Properly handle new column for sync table

ok300:
As discussed, moving the statement to the sync migrations function.

Even so, it fails with

```
Error: Failed to use the local DB for persistence: rusqlite_migrate error: RusqliteError { query: "ALTER TABLE payments_external_info ADD COLUMN lnurl_withdraw_endpoint TEXT;", err: SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("no such table: payments_external_info")) }
```

even after clearing the SQL files locally and trying again.

Could be an issue with my setup, so marking as draft until someone else can confirm it's working.

### DIFF
--- a/libs/sdk-core/src/persist/migrations.rs
+++ b/libs/sdk-core/src/persist/migrations.rs
@@ -460,6 +460,14 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
         channel_opening_fees TEXT NOT NULL
        ) STRICT;
        ",
+        "
+        CREATE TABLE IF NOT EXISTS payments_external_info (
+         payment_id TEXT NOT NULL PRIMARY KEY,
+         lnurl_success_action TEXT,
+         ln_address TEXT,
+         lnurl_metadata TEXT
+         ) STRICT;
+        ",
         "ALTER TABLE payments_external_info ADD COLUMN lnurl_withdraw_endpoint TEXT;",
     ]
 }
--- a/libs/sdk-core/src/persist/migrations.rs
+++ b/libs/sdk-core/src/persist/migrations.rs
@@ -468,6 +468,28 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
          lnurl_metadata TEXT
          ) STRICT;
         ",
-        "ALTER TABLE payments_external_info ADD COLUMN lnurl_withdraw_endpoint TEXT;",
+        "
+        ALTER TABLE payments_external_info RENAME TO payments_external_info_old;
+
+        CREATE TABLE payments_external_info (
+         payment_id TEXT NOT NULL PRIMARY KEY,
+         lnurl_success_action TEXT,
+         ln_address TEXT,
+         lnurl_metadata TEXT,
+         lnurl_withdraw_endpoint TEXT
+        ) STRICT;
+
+        INSERT INTO payments_external_info
+         (payment_id, lnurl_success_action, ln_address, lnurl_metadata, lnurl_withdraw_endpoint)
+         SELECT
+          payment_id,
+          lnurl_success_action,
+          ln_address,
+          lnurl_metadata,
+          NULL
+         FROM payments_external_info_old;
+
+        DROP TABLE payments_external_info_old;
+        ",
     ]
 }
--- a/libs/sdk-core/src/persist/migrations.rs
+++ b/libs/sdk-core/src/persist/migrations.rs
@@ -385,37 +385,10 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
         ) STRICT;
         ",
 
-        //sync & backup
+        //sync & backup (moved to sync migration function below)
         "
-        CREATE TABLE IF NOT EXISTS sync.sync_requests (
-         id INTEGER PRIMARY KEY AUTOINCREMENT,
-         changed_table TEXT NOT NULL
-        ) STRICT;
-
-        CREATE TRIGGER sync.sync_requests_swaps 
-         AFTER INSERT ON swaps         
-        BEGIN
-         INSERT INTO sync_requests(changed_table) VALUES('swaps');
-        END;
-
-        CREATE TRIGGER sync.sync_requests_swap_refunds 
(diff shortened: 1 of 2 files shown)


libs/sdk-core/src/persist/migrations.rs
```diff
@@ -460,5 +460,6 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
         channel_opening_fees TEXT NOT NULL
        ) STRICT;
        ",
+        "ALTER TABLE payments_external_info ADD COLUMN lnurl_withdraw_endpoint TEXT;",
```
roeierez, line 463: Since we already had a migration in main db this statement might have already been running so it will fail bow with column already exists.
In sqlite for these cases we need to create a temp table with the right scheme copy the data to the temp table, remove the originall table and as a final step rename the temp table to the original name

roeierez: We have an example in our code

roeierez (changes requested):

libs/sdk-core/src/persist/migrations.rs
ok300: Its not the case here, because the error I listed above is "no such table: payments_external_info"

ok300: I ran the CLI on this branch with a new wallet (fresh invite code, no local SQL files) and got the same error:

```
sdk> connect -i "<invite-code>"

Error: Failed to use the local DB for persistence:
  rusqlite_migrate error: RusqliteError {
    query: "ALTER TABLE payments_external_info ADD COLUMN lnurl_withdraw_endpoint TEXT;",
    err: SqliteFailure(
      Error {code: Unknown, extended_code: 1 },
      Some("no such table: payments_external_info")
    )
  }
```

ok300:
I could confirm the PR on 3 separate test-wallets.

All 3 had this issue before, but with the PR `backup` and `sync` worked fine. However for the PR to work, I had to first remove any local SQL files.

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -460,5 +460,99 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
...
+         AFTER INSERT ON reverse_swaps
+        BEGIN
+         INSERT INTO sync_requests(changed_table) VALUES('reverse_swaps');
+        END;
+
+        CREATE TRIGGER IF NOT EXISTS sync_requests_payments_external_info
+         AFTER INSERT ON payments_external_info
+        BEGIN
+         INSERT INTO sync_requests(changed_table) VALUES('payments_external_info');
+        END;
```
JssDWt, line 532: I think these triggers should not be here?

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -460,5 +460,99 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
...
+         SELECT
+          payment_id,
+          lnurl_success_action,
+          ln_address,
+          lnurl_metadata,
+          NULL
+         FROM payments_external_info_old;
+
+        DROP TABLE payments_external_info_old;
+        ",
```
JssDWt, line 556: There are triggers on this table. I think triggers reference the table id, not the table name. So renaming the table, then creating a new one with the same name will malfunction the triggers.

libs/sdk-core/src/persist/migrations.rs
JssDWt: The triggers are created in the main db migrations as well, but they are not optional. So I think that may either fail on initial db creation, or create duplication side effects.

libs/sdk-core/src/persist/migrations.rs
JssDWt: Thinking about it, recreating the trigger here will fail the main db migration, because that's also unconditionally creating the trigger.

roeierez: @JssDWt @ok300  Perhaps now that we have added the triggers here we can replace the main db triggers with empty statements?
And we can also re-add the trigger after the table deletion (because sqlite implicitly deletes all triggers associated with the table before the table is deleted)?

JssDWt: I think we should just let the main db migrations run first, and the sync migrations second. Then you also don't need to recreate the table here.

So do:
`CREATE TABLE IF NOT EXISTS ...` on the `payments_external_info` table.
then add the column.

JssDWt: I see the main table is actually created first, that was my confusion.

So just the CREATE TABLE IF NOT EXISTS ... on the payments_external_info table.
then add the column, that should be enough.

JssDWt: I was wrong about the order of migrations, I thought the sync migration ran first, and then the main migration, but it's the other way around. So just readding the trigger is enough. No need to change the main migrations.

roeierez: I am thinking even if the main db run first it is confusing to try and foresee all the combinations with the main db and sync db so we can take advantage of this refactoring to strictly isolate the migrations, each on its own file and schema and also remove any future confusion. That's why I suggested to nullify these statements.

ok300: I moved the `payments_external_info` trigger right after the re-creation of the table, so the trigger is created on the new table: ccb7c850f9f5c538fd08695a4fa9b4011f7acce0

ok300: @roeierez if I understood your point correctly to nullify main migration statments which are not needed anymore, this removes the triggers and the `sync_requests` table from the main migrations: 1eab2935fc3e28e5bd21958fca50ba2ff6d5fda7

They all of now exist in the sync migration below.

JssDWt (approved):
Tested on a node that had the lnurl_withdraw_endpoint column missing. Works!

roeierez (approved):

PR #491, opened 2023-09-28
update flutter readme

JssDWt:
Update flutter readme to include the proper binaries.

Also copy the *contents* of the jniLibs folder and not the jniLibs folder itself.
Problem was if you ran `make android` twice, there would be a `jniLibs` folder inside the `jniLibs` folder.


dangeross (approved):
Looks good!

PR #492, opened 2023-09-29
Fix rev swap edge case affecting offline nodes

ok300:
Before this PR, the following scenario could occur:
- reverse swap is started and goes into `InProgress` state
- lockup tx is confirmed, claim tx is broadcasted
  - before the next `sync`, the node goes offline (rev swap state is still `InProgress`)
- claim tx gets confirmed
- node comes back online and executes the loop below for monitored rev swaps (incl. the above `InProgress` one)
  - it tries to re-broadcast the claim tx, calling `create_claim_tx`
  - however `create_claim_tx` relies on utxos to calculate the total confirmed amount which should be sent in the claim tx
  - since the claim tx is already confirmed, thus spending the available utxo in the lockup address, this means no utxos in the lockup address are found
  - this leads to a panic, because with total amount of 0 we try to subtract mining fees, which gives a negative result, which is not allowed for a tx value

With this PR, we only attempt to create and build the claim tx if it's not already confirmed.

If it is confirmed, this step is skipped. On the next `sync`, the `InProgress` rev swaps will automatically transition to `CompletedConfirmed`.

### DIFF
--- a/libs/sdk-core/src/reverseswap.rs
+++ b/libs/sdk-core/src/reverseswap.rs
@@ -367,18 +367,31 @@ impl BTCSendSwap {
         // Depending on the new state, decide next steps and transition to the new state
         for rs in monitored {
             debug!("Checking monitored reverse swap {rs:?}");
-            // (Re-)Broadcast the claim tx for monitored reverse swaps that have a confirmed lockup tx
-            if matches!(self.get_lockup_tx_status(&rs).await?, TxStatus::Confirmed) {
-                info!("Lock tx is confirmed, preparing claim tx");
-                let claim_tx = self.create_claim_tx(&rs).await?;
-                let claim_tx_broadcast_res = self
-                    .chain_service
-                    .broadcast_transaction(serialize(&claim_tx))
-                    .await;
-                match claim_tx_broadcast_res {
-                    Ok(txid) => info!("Claim tx was broadcast with txid {txid}"),
-                    Err(e) => error!("Claim tx failed to broadcast: {e}"),
-                }
+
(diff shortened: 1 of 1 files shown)


ok300:
Will mark it as a draft until I test it in several configurations.

libs/sdk-core/src/reverseswap.rs
```diff
@@ -367,18 +367,31 @@ impl BTCSendSwap {
...
-                    .await;
-                match claim_tx_broadcast_res {
-                    Ok(txid) => info!("Claim tx was broadcast with txid {txid}"),
-                    Err(e) => error!("Claim tx failed to broadcast: {e}"),
-                }
+
+            let lockup_tx_status = self.get_lockup_tx_status(&rs).await?;
+            let claim_tx_status = self.get_claim_tx_status(&rs).await?;
+
+            match claim_tx_status {
```
roeierez, line 374: In a glance it looks correct but one thing is unclear to me: 
The  `execute_pending_reverse_swaps` is only called after `refresh_monitored_reverse_swaps` so the question is why the status wasn't updated to Completed? Shouldn't the provider return Completed here (https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/reverseswap.rs#L475) in this case?

roeierez (changes requested):

roeierez:
The question is now that the state transition was fixed: https://github.com/breez/breez-sdk/pull/496
Do we really need this PR? 
It seems a claim confirmed swaps won't be listed in the monitored anymore and won't enter the `execute_pending_reverse_swaps`

ok300:
Correct. Makes sense, plus I tested and works, so I'll drop this.

PR #493, opened 2023-09-29
Fix rev swap edge case caused by claim tx delays

ok300:
If the rev swap is created and proceeds successfully until the lockup tx, but the claim tx is not broadcasted until after the rev swap timeout is reached (for example, because the SDK node was offline), then the service provider re-claims the locked amount and cancels the HODL invoice.

This scenario resulted in a panic when the device came back online and tried to construct and broadcast the claim tx, because of how the claim tx amount was calculated.

This PR fixes that.

### DIFF
--- a/libs/sdk-core/src/reverseswap.rs
+++ b/libs/sdk-core/src/reverseswap.rs
@@ -280,10 +280,16 @@ impl BTCSendSwap {
                 debug!("Found confirmed txs for lockup address {lockup_addr}: {confirmed_txs:?}");
                 let utxos = get_utxos(lockup_addr.to_string(), confirmed_txs)?;
 
-                let confirmed_amount: u64 = utxos
-                    .confirmed
-                    .iter()
-                    .fold(0, |accum, item| accum + item.value);
+                // To decide the claim tx amount, we use the previously committed to amount
+                // We avoid trying to derive it from confirmed utxos on the lockup address, because
+                // in certain timeout scenarios (e.g. if the claim tx is not broadcast within the
+                // rev swap allocated time), then the service provider will claim the sats back
+                // and cancel the HODL invoice. Practically this results in a new utxo from the lockup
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/reverseswap.rs
```diff
@@ -280,10 +280,16 @@ impl BTCSendSwap {
...
+                // To decide the claim tx amount, we use the previously committed to amount
+                // We avoid trying to derive it from confirmed utxos on the lockup address, because
+                // in certain timeout scenarios (e.g. if the claim tx is not broadcast within the
+                // rev swap allocated time), then the service provider will claim the sats back
+                // and cancel the HODL invoice. Practically this results in a new utxo from the lockup
+                // address, of the same amount as was locked previously. In this scenario, relying
+                // on confirmed utxos to determine the claim tx amount will result in a panic (0 - fees < 0)
+                // Therefore we read the claim tx amount from the originally agreed upon onchain amount,
+                // confirmed by the service provider on rev swap creation.
+                let claim_amount_sat = rs.onchain_amount_sat;
```
roeierez, line 292: Isn't it better to validate that the total utxo value is at least the claim_amount_sat before constructing the transaction?

roeierez (changes requested):

libs/sdk-core/src/reverseswap.rs
roeierez: Also why the status wasn't updated to Canceled in the call to `refresh_monitored_reverse_swaps`?

roeierez: Anyway I agree that to make sure we claim exactly the amount we know we need to take is better than calculating the utxo and if the tx is not valid so we will keep trying untill timeout. LGTM.

roeierez (approved):
LGTM

Issue #456, opened 2023-09-13
Migrate model from `gl_client::pb` to the equivalents in `gl_client::pb::cln`


ok300:
* https://github.com/breez/breez-sdk/pull/455/

A few other relevant structs: `Amount`, `InvoiceStatus`, `OffChainPayment`, `PayStatus`, `Peer`, `WithdrawResponse`, etc.

dangeross:
Can this be closed?

PR #494, opened 2023-10-01
Cln API

roeierez:
This PR moves from the greenlight.proto deprecated API and uses the node.proto instead.
Greenlight are going to retire the greenlight.proto very soon: https://github.com/Blockstream/greenlight/pull/279

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -558,13 +558,13 @@ impl NodeAPI for Greenlight {
         Ok(())
     }
 
-    async fn sweep(&self, to_address: String, fee_rate_sats_per_vbyte: u64) -> Result<Vec<u8>> {
+    async fn sweep(&self, to_address: String, fee_rate_sats_per_vbyte: u32) -> Result<Vec<u8>> {
         let mut client = self.get_node_client().await?;
 
         let request = pb::cln::WithdrawRequest {
             feerate: Some(pb::cln::Feerate {
                 style: Some(pb::cln::feerate::Style::Perkw(
-                    fee_rate_sats_per_vbyte as u32 * 250,
+                    fee_rate_sats_per_vbyte * 250,
                 )),
             }),
             satoshi: Some(pb::cln::AmountOrAll {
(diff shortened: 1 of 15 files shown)


ok300:
Fixes #456

ok300:
Also FYI, the CLN API sometimes runs into this: https://github.com/Blockstream/greenlight/issues/280

libs/sdk-core/src/persist/channels.rs
```diff
@@ -80,22 +84,21 @@ impl SqliteStorage {
                    spendable_msat, 
                    receivable_msat,
                    closed_at,
-                   funding_outnum
+                   funding_outnum,
+                   closed_at,
```
dangeross, line 88: Inserting `closed_at` twice

libs/sdk-core/src/persist/channels.rs
```diff
@@ -108,6 +111,9 @@ impl SqliteStorage {
                     _ => c.closed_at,
                 },
                 c.funding_outnum,
+                c.closed_at,
```
dangeross, line 114: `closed_at` also set in the match statement above:
```
match c.state {
    ChannelState::Opened | ChannelState::PendingOpen => None,
    _ => c.closed_at,
},
```

dangeross (changes requested):
Looks good. Just picked up on some issues inserting/updating channels

libs/sdk-core/src/persist/channels.rs
roeierez: Good one! Done.

libs/sdk-core/src/persist/channels.rs
roeierez: Fixed

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -566,26 +558,24 @@ impl NodeAPI for Greenlight {
         Ok(())
     }
 
-    async fn sweep(
-        &self,
-        to_address: String,
-        fee_rate_sats_per_vbyte: u64,
-    ) -> Result<WithdrawResponse> {
-        let mut client = self.get_client().await?;
+    async fn sweep(&self, to_address: String, fee_rate_sats_per_vbyte: u64) -> Result<Vec<u8>> {
```
JssDWt, line 561: Let's make `fee_rate_sats_per_vbyte` a `u32`.

JssDWt (commented):
Looks good.
Just a nit about the fee rate conversion.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Done. Had to do it all the way through to the udl file.

JssDWt (approved):

PR #496, opened 2023-10-02
Fix rev swap state transitions

ok300:
This PR fixes the handling of the initial reverse swap state, when a node is restored.

It also handles two more edge cases in state transition that result in state Cancelled.


roeierez (approved):
LGTM

PR #497, opened 2023-10-02
Use Request/Response format on sendOnchain & refund

erdemyerebasmaz:
This PR addresses feedback in
- https://github.com/breez/c-breez/issues/629#issuecomment-1742642113

and converts `sendOnchain` & `refund` to use `Request`/`Response` format.

### Changelist:
- Rename `req` params to `request`
  - `req_data` params are renamed to `request_data`
- Reorder `pay_lnurl` & `withdraw_lnurl` params to be consistent with other lnurl functions: `(amount, comment/description, request_data)`

mostly on binding layer with a few exceptions. See `breez_services.rs.`

### sdk-flutter changelist:
- Apply
  - https://github.com/breez/breez-sdk/pull/463 &
  - https://github.com/breez/breez-sdk/pull/478 changes
- Convert expression bodies that stretch to multiple lines to block bodies to increase readability, vice versa
- Update `inProgressSwap()`'s documentation comment

### Todo:
- Apply these changes to 
- [ ] sdk-docs,
- [ ] example apps,
- [ ] c-breez

### Unrelated issues encountered:
Encountered this warning during running `make all` from `sdk-bindings`:
```
   Compiling uniffi-bindgen-go v0.1.2+v.0.23.0 (/Users/erdem/.cargo/git/checkouts/uniffi-bindgen-go-10b6486a739c08f4/62347ac/bindgen)
warning: field `config` is never read
  --> bindgen/src/gen_go/mod.rs:72:5
   |
71 | pub struct GoWrapper<'a> {
   |            --------- field in this struct
72 |     config: Config,
   |     ^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `uniffi-bindgen-go` (lib) generated 1 warning
```


tools/sdk-cli/Cargo.toml
```diff
@@ -16,4 +16,4 @@ serde_json = "1.0"
 tiny-bip39 = "1"
 tokio = { version = "1", features = ["rt-multi-thread"] }
 serde = { version = "1.0", features = ["derive"] }
-clap = { version = "4", features = ["derive"] }
\ No newline at end of file
+clap = { version = "<4.2.*", features = ["derive"] }
```
ok300, line 19: Why not update the local rust version instead?

```
rustup update stable
```

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -633,7 +654,7 @@ interface BlockingBreezServices {
    RecommendedFees recommended_fees();
 
    [Throws=SdkError]
-   BuyBitcoinResponse buy_bitcoin(BuyBitcoinRequest req);
+   BuyBitcoinResponse buy_bitcoin(BuyBitcoinRequest request);
```
ok300, line 657: NIT: IMO `req` and `req_data` is better because it's less verbose (causes shorter lines of code, so less linebreaks when formatting), while the name is still clear enough as to what it means.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -540,16 +561,16 @@ interface BlockingBreezServices {
    Payment send_spontaneous_payment(string node_id, u64 amount_sats);
 
    [Throws=SdkError]
-   ReceivePaymentResponse receive_payment(ReceivePaymentRequest req_data);
+   ReceivePaymentResponse receive_payment(ReceivePaymentRequest request);
 
    [Throws=SdkError]
-   LnUrlPayResult pay_lnurl(LnUrlPayRequestData req_data, u64 amount_sats, string? comment);
+   LnUrlPayResult pay_lnurl(u64 amount_sats, string? comment, LnUrlPayRequestData request_data);
```
ok300, line 567: NIT: IMO the reordering of LNURL args wasn't needed. They were ordered by importance:
- `req_data`: the data returned by `parse` in the first step of LNURL
- `amount`
- `comment` which was optional

ok300 (changes requested):
Looks good, but IMO some of the changes are not necessary.

As to RN bindings: to generate them see the `README` in https://github.com/breez/breez-sdk-rn-generator

dangeross:
Generating RN is quite similar to sdk-flutter. There are instructions here: https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/DEVELOPING.md

tools/sdk-cli/Cargo.toml
erdemyerebasmaz: This was reported by Ademar in a group chat, he did ask if it's OK to upgrade and didn't get answers so I assumed we were targeting 1.69.0 as a requirement.
I will update my rust version 👍

libs/sdk-bindings/src/breez_sdk.udl
erdemyerebasmaz: I see, it wasn't clear from the SDK implementation. 
I will revert these changes.

libs/sdk-bindings/src/breez_sdk.udl
erdemyerebasmaz: We were using `request` in some places. It's all good as long as they are consistent

libs/sdk-bindings/src/breez_sdk.udl
erdemyerebasmaz: Done e48902088783449e148bba6c6b89528c0fa21451

libs/sdk-bindings/src/breez_sdk.udl
erdemyerebasmaz: Done e48902088783449e148bba6c6b89528c0fa21451

tools/sdk-cli/Cargo.toml
erdemyerebasmaz: Reverted e48902088783449e148bba6c6b89528c0fa21451

erdemyerebasmaz:
Breaking changes:
- Use `req` shorthand for `request` params across all bindings.
- Wrap following APIs in Request/Response struct
  - send_onchain, refund, send_payment
  - send_spontaneous_payment, lnurl_pay (as part of #536)
  - lnurl_withdraw (as part of #535, merged to main, not included in this PR)
- Rename `req_data` of lnurl_pay & lnurl_withdraw to `data` as they now belong to a Request struct

libs/sdk-core/src/breez_services.rs
```diff
@@ -226,17 +225,12 @@ impl BreezServices {
     ///
     /// # Arguments
```
ok300, line 226: IMO we can get rid of the `Arguments` doc section for functions that merged all args in a single `..Request` arg.

That doc section became superfluous, the caller now goes to the `..Request` doc to read about the request args.

libs/sdk-core/src/models.rs
```diff
@@ -805,6 +829,27 @@ pub struct SweepResponse {
     pub txid: Vec<u8>,
 }
 
+pub struct SendOnchainRequest {
+    pub amount_sat: u64,
+    pub onchain_recipient_address: String,
+    pub pair_hash: String,
+    pub sat_per_vbyte: u64,
```
ok300, line 836: Here `sat_per_vbyte` is an `u64` but in `RefundRequest` it's an `u32`.

For consistency `u32` could be used for both.

ok300 (changes requested):
Looks good, left a few comments.

One only problem I found is the CLI doesn't compile for me. Can you please check that?

ok300:
> One only problem I found is the CLI doesn't compile for me. Can you please check that?

I think its because when you reverted the `clap` downgrade in e48902088783449e148bba6c6b89528c0fa21451 the `Cargo.lock` was left untouched and still referenced the previous (older) `clap` version.

libs/sdk-core/src/breez_services.rs
erdemyerebasmaz: Done 9d7df482d087253e02d776a4aaddc3810abc4b8f

erdemyerebasmaz:
> > One only problem I found is the CLI doesn't compile for me. Can you please check that?
> 
> I think its because when you reverted the `clap` downgrade in [e489020](https://github.com/breez/breez-sdk/commit/e48902088783449e148bba6c6b89528c0fa21451) the `Cargo.lock` was left untouched and still referenced the previous (older) `clap` version.

Yeah, it has been later addressed:

> Solution:
> 
> Downgrade clap dependency to a Rust 1.69.0 compatible version https://github.com/breez/breez-sdk/commit/f95221a2b5ff630f7264da541783472bf6cb7a70

libs/sdk-core/src/models.rs
erdemyerebasmaz: Done c579ec0da3c0d64d74b2b64bdfb334a81dc01b9f

ok300 (approved):

PR #498, opened 2023-10-03
Persist channel closed_at after lookup

ok300:
If a closed channel's `closed_at` is unknown, and we have to look it up, we now persist it.


roeierez (approved):
LGTM

PR #505, opened 2023-10-05
Remove log library from sdk and move logs for the app

ademar111190:
SDK side of logging library changes https://github.com/breez/c-breez/pull/657


roeierez:
Perhaps not use any logging library in the sdk? We have API for listening to log messages where the user can listen to and has the freedom to choose her ownogging library on the application side. Exposing the log messages as a  broadcast stream sounds good to me.

ubbabeck (approved):
Looks good 

tACK

ademar111190:
Good callout @roeierez I totally agree.
I moved the logs to cbreez and removed the library.

erdemyerebasmaz (approved):
Looks like the right direction 👍

roeierez (approved):
LGTM

PR #510, opened 2023-10-09
Trim `lightning:` from LNInvoice bolt11 response

dangeross:
Trims the "lightning:" from the bolt11 field of LNInvoice if one is passed.

```
sdk> parse lightning:lnbc210n1pj3z6d2pp52wuh79egvd3x67zucx48ktkgfmjxa87x4g34q2t8k08gs3g0kjvqdqqcqzzsxqrrssrzjqvgptfurj3528snx6e3dtwepafxw5fpzdymw9pj20jj09sunnqmwpapyqqqqqqq005qqqqlgqqqqqqgq9qsp5x0jl8kwgshnv4mlmapvnfwzm0djs26n8fz0mpn57z0vh9swqqd2s9qyyssqzcxszyy6p6vyvp763pjg3tpl4suxalmm45tus07uu2e53vm27k9rgec4rca725rygurg9rexhkke0h62ewtfpmd4aevz89enaxfzr5gq4pwer8
```

Before
```
{
  "Bolt11": {
    "invoice": {
      "bolt11": "lightning:lnbc210n1pj3z6d2pp52wuh79egvd3x67zucx48ktkgfmjxa87x4g34q2t8k08gs3g0kjvqdqqcqzzsxqrrssrzjqvgptfurj3528snx6e3dtwepafxw5fpzdymw9pj20jj09sunnqmwpapyqqqqqqq005qqqqlgqqqqqqgq9qsp5x0jl8kwgshnv4mlmapvnfwzm0djs26n8fz0mpn57z0vh9swqqd2s9qyyssqzcxszyy6p6vyvp763pjg3tpl4suxalmm45tus07uu2e53vm27k9rgec4rca725rygurg9rexhkke0h62ewtfpmd4aevz89enaxfzr5gq4pwer8",
      "payee_pubkey": "0203e1b9715f923b4e30ccca52b8045870f95cc2fcf7f0b0f66d2ecac29f069e8e",
       ...
    }
  }
}
```
After
```
{
  "Bolt11": {
    "invoice": {
      "bolt11": "lnbc210n1pj3z6d2pp52wuh79egvd3x67zucx48ktkgfmjxa87x4g34q2t8k08gs3g0kjvqdqqcqzzsxqrrssrzjqvgptfurj3528snx6e3dtwepafxw5fpzdymw9pj20jj09sunnqmwpapyqqqqqqq005qqqqlgqqqqqqgq9qsp5x0jl8kwgshnv4mlmapvnfwzm0djs26n8fz0mpn57z0vh9swqqd2s9qyyssqzcxszyy6p6vyvp763pjg3tpl4suxalmm45tus07uu2e53vm27k9rgec4rca725rygurg9rexhkke0h62ewtfpmd4aevz89enaxfzr5gq4pwer8",
      "payee_pubkey": "0203e1b9715f923b4e30ccca52b8045870f95cc2fcf7f0b0f66d2ecac29f069e8e",
       ...
    }
  }
}
```

### DIFF
--- a/libs/sdk-core/src/invoice.rs
+++ b/libs/sdk-core/src/invoice.rs
@@ -4,6 +4,7 @@ use hex::ToHex;
 use lightning::routing::gossip::RoutingFees;
 use lightning::routing::*;
 use lightning_invoice::*;
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use std::time::UNIX_EPOCH;
@@ -147,7 +148,8 @@ pub fn add_lsp_routing_hints(
 
 /// Parse a BOLT11 payment request and return a structure contains the parsed fields.
 pub fn parse_invoice(bolt11: &str) -> Result<LNInvoice> {
-    let bolt11 = bolt11.strip_prefix("lightning:").unwrap_or(bolt11);
+    let re = Regex::new(r"(?i)^lightning:")?;
+    let bolt11 = re.replace_all(bolt11, "");
     let signed = bolt11.parse::<SignedRawInvoice>()?;
     let invoice = Invoice::from_signed(signed)?;
 
--- a/libs/sdk-core/src/invoice.rs
+++ b/libs/sdk-core/src/invoice.rs
@@ -172,7 +172,7 @@ pub fn parse_invoice(bolt11: &str) -> Result<LNInvoice> {
     let converted_hints = invoice_hints.iter().map(RouteHint::from_ldk_hint).collect();
     // return the parsed invoice
     let ln_invoice = LNInvoice {
-        bolt11: bolt11.to_string(),
+        bolt11: invoice.to_string(),
         payee_pubkey,
         expiry: invoice.expiry_time().as_secs(),
         amount_msat: invoice.amount_milli_satoshis(),
--- a/libs/sdk-core/src/invoice.rs
+++ b/libs/sdk-core/src/invoice.rs
@@ -147,10 +147,8 @@ pub fn add_lsp_routing_hints(
 
 /// Parse a BOLT11 payment request and return a structure contains the parsed fields.
 pub fn parse_invoice(bolt11: &str) -> Result<LNInvoice> {
-    let signed = bolt11
-        .strip_prefix("lightning:")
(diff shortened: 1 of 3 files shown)


libs/sdk-core/src/invoice.rs
```diff
@@ -172,7 +172,7 @@ pub fn parse_invoice(bolt11: &str) -> Result<LNInvoice> {
     let converted_hints = invoice_hints.iter().map(RouteHint::from_ldk_hint).collect();
     // return the parsed invoice
     let ln_invoice = LNInvoice {
-        bolt11: bolt11.to_string(),
+        bolt11: invoice.to_string(),
```
JssDWt, line 175: The resulting invoice string can be slightly different from the actual bolt11 invoice string passed as input by parsing and re-encoding.
I think it's best to do something like:

```rust
let bolt11 = bolt11
        .strip_prefix("lightning:")
        .unwrap_or(bolt11);
let signed = bolt11.parse::<SignedRawInvoice>()?;
```

And pass `bolt11` here.

JssDWt (changes requested):

libs/sdk-core/src/invoice.rs
```diff
@@ -147,10 +147,8 @@ pub fn add_lsp_routing_hints(
 
 /// Parse a BOLT11 payment request and return a structure contains the parsed fields.
 pub fn parse_invoice(bolt11: &str) -> Result<LNInvoice> {
-    let signed = bolt11
-        .strip_prefix("lightning:")
-        .unwrap_or(bolt11)
-        .parse::<SignedRawInvoice>()?;
+    let bolt11 = bolt11.strip_prefix("lightning:").unwrap_or(bolt11);
```
roeierez, line 150: Shouldn't we also strip LIGHTNING (ignore case)?

roeierez (changes requested):

libs/sdk-core/src/invoice.rs
dangeross: Changed to a case insensitive replace

roeierez (approved):
LGTM

JssDWt (approved):
LGTM

PR #515, opened 2023-10-10
Add optional claim_txid to ReverseSwapInfo

### LINKED ISSUE #512: Expose txid in ReverseSwapInfo
It would be nice to be able to be able to retrieve either both the lockup txid and claimtx or  alternatively we return a generic one which shows the current txid.

Not sure what makes more sense to do.

ok300:
The field is dynamically calculated based on the rev swap status.

Fixes #512

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -427,6 +427,7 @@ dictionary ReverseSwapPairInfo {
 dictionary ReverseSwapInfo {
     string id;
     string claim_pubkey;
+    string? lockup_txid;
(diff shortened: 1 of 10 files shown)


dangeross (approved):
Looks good. 

Outside of this PR, but I wonder if it worth pushing status changes to swaps as events also?

roeierez:
> Outside of this PR, but I wonder if it worth pushing status changes to swaps as events also?

Makes lot of sense to me.

roeierez:
@ok300 looks like we need to add here also the lock tx.

roeierez (approved):

PR #520, opened 2023-10-12
Add closing_txid to Channel, ClosedChannelPaymentDetails

### LINKED ISSUE #484: add closing tx field to `ClosedChannelPaymentDetails`
It would be nice to be able to fetch the txid for a closing channel (pending) and also a closed transaction. At them moment we can only fetch the channel opening transaction.

ok300:
`closed_channel_to_transaction` was expanded to 2 fields which could depend on a chainservice lookup:
- `closed_at`
- `closing_txid`

I've refactored it to make it more readable, extracting the lookup methods to separate functions.

This also includes a new optional field `closing_txid` in `Channel`, which is by default empty and is looked up (and persisted) once, just as `closed_at`.

Fixes #484

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -22,7 +22,7 @@ use tonic::{Request, Status};
 
 use crate::backup::{BackupRequest, BackupTransport, BackupWatcher};
 use crate::boltzswap::BoltzApi;
-use crate::chain::{ChainService, MempoolSpace, RecommendedFees};
+use crate::chain::{ChainService, MempoolSpace, Outspend, RecommendedFees};
 use crate::error::{SdkError, SdkResult};
 use crate::fiat::{FiatCurrency, Rate};
 use crate::greenlight::{GLBackupTransport, Greenlight};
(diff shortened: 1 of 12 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1220,55 +1220,90 @@ impl BreezServices {
...
-                        }
-                    }
-                };
-
-                // Persist closed_at, if we had to look it up
-                let mut updated_channel = channel.clone();
-                updated_channel.closed_at = Some(looked_up_channel_closed_at);
-                self.persister.insert_or_update_channel(updated_channel)?;
+                persist_channel_updates = true;
+                self.lookup_chain_service_closed_at(channel.clone()).await?
```
roeierez, line 1278: At this point we need to fetch also the closing tx id so it looks better to make this function `lookup_chain_service_closed_at` return both the closed_at and the txid instead of calling it twice and save the network call.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: Done, I merged the two lookups in one method.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1220,55 +1220,78 @@ impl BreezServices {
...
-                                now_epoch_sec
-                            }
-                            Some(block_time) => block_time,
-                        }
+                        warn!("Blocktime could not be determined for from closing outspend, defaulting closed_at to epoch time");
+                        SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs()
                     }
+                    Some(block_time) => block_time,
                 };
+                let processed_closing_txid = maybe_closing_txid.unwrap_or_default();
```
roeierez, line 1279: It looks like we persisting defaults for both closed_at and the closing_tx_id in case we didn't find the outspend.
It looks like it prevents us from fetching it next time.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: Yes, I thought that's desired behavior. If outspend not found:
- `closed_at` is saved as `now()`
- `closing_txid` is saved as empty string

If we don't persist them in this case, yes we will try to look it up again next time, but until then the fields would be empty. Is that how it should actually be?

roeierez (approved):
LGTM

Issue #417, opened 2023-09-01
Consolidate API units to msats


ok300:
This means replace sats with msats
* externally: in all exposed methods, arguments and return structs
* internally: in all internal methods (fee calculation, etc)

Since msat is the native unit on LN, this gets rid of entire classes of issues (rounding, conversion, etc).

ok300:
`receive_payment` and `lnurl_withdraw` are done in #535

dangeross:
`send_spontaneous_payment` and `lnurl_pay` are done in #536

erdemyerebasmaz:
What's the status on this? Are there any remaining tasks?

ok300:
The only place I found where I'm not sure if a conversion is necessary is

https://github.com/breez/breez-sdk/blob/5adbfac3948352d8823453a75e196cedb7817638/libs/sdk-core/src/models.rs#L841-L846

@JssDWt what do you think, do we need to do the rounding to sat in L843? Since both input and output are in msat?

ok300:
Great, then this is complete and can be closed.

PR #525, opened 2023-10-17
Validate amount parameter

### LINKED ISSUE #522: `send_payment` throws an error if amount_sats and bolt11 amount_msat is set
```
sdk> send_payment lnbc210n1pjje6axpp5lwnk3d6m75av86fdh6rpsnnmajuczyjvrprvlyeqdwp3e46am74qdqqcqzzsxqrrssrzjqvgptfurj3528snx6e3dtwepafxw5fpzdymw9pj20jj09sunnqmwpapyqqqqqqq005qqqqlgqqqqqqgq9qsp5hlx2z9egv2xft3czlzx7gpruh4erngeq9v7tda0sg05ruuuy8vts9qyyssqwnvt8vlmxkwdlz85nsuvp84wg0mxl2z7z8q0z89ej9m4cwrhshlpmy8z9g3nrzpzqje5cr4gv8tgurx9xsh5g6vtvudmeelmkg5wrkcpmdzy9t -a 21
```
```
Error: Failed to send payment: status: Unknown, message: "Error calling method Pay: RpcError { code: Some(-32602), message: \"msatoshi parameter unnecessary\" }", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 16 Oct 2023 07:53:10 GMT", "content-length": "0"} }
```

roeierez:
fixed #522

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -222,21 +222,21 @@ impl BreezServices {
     /// Pay a bolt11 invoice
     ///
     /// Calling `send_payment` ensures that the payment is not already completed, if so it will result in an error.
-    /// If the invoice doesn't specify an amount, the amount is taken from the `amount_sats` arg.
+    /// If the invoice doesn't specify an amount, the amount is taken from the `amount_msat` arg.
     ///
     /// # Arguments
     ///
     /// * `bolt11` - The bolt11 invoice
-    /// * `amount_sats` - The amount to pay in satoshis
+    /// * `amount_msat` - The amount to pay in millisatoshis
     pub async fn send_payment(
         &self,
         bolt11: String,
-        amount_sats: Option<u64>,
+        amount_msat: Option<u64>,
     ) -> SdkResult<Payment> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(bolt11.as_str())?;
         let invoice_amount_msat = parsed_invoice.amount_msat.unwrap_or_default();
-        let provided_amount_msat = amount_sats.unwrap_or_default() * 1000;
+        let provided_amount_msat = amount_msat.unwrap_or_default();
 
         // Ensure amount is provided for zero invoice
         if provided_amount_msat == 0 && invoice_amount_msat == 0 {
@@ -262,7 +262,7 @@ impl BreezServices {
             None => {
                 let payment_res = self
                     .node_api
-                    .send_payment(bolt11.clone(), amount_sats)
+                    .send_payment(bolt11.clone(), amount_msat)
                     .await;
                 self.on_payment_completed(
                     parsed_invoice.payee_pubkey.clone(),
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -235,6 +235,23 @@ impl BreezServices {
     ) -> SdkResult<Payment> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(bolt11.as_str())?;
+        let invoice_amount_msat = parsed_invoice.amount_msat.unwrap_or_default();
+        let provided_amount_msat = amount_sats.unwrap_or_default() * 1000;
+
+        // Ensure amount is provided for zero invoice
+        if provided_amount_msat == 0 && invoice_amount_msat == 0 {
+            return Err(SdkError::SendPaymentFailed {
+                err: "amount must be provided when paying a zero invoice".into(),
+            });
+        }
+
+        // Ensure amount is not provided for invoice that contains amount
+        if provided_amount_msat > 0 && invoice_amount_msat > 0 {
+            return Err(SdkError::SendPaymentFailed {
+                err: "amount should not be provided when paying a non zero invoice".into(),
+            });
+        }
+
         match self
             .persister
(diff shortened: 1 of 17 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -235,6 +235,23 @@ impl BreezServices {
...
+
+        // Ensure amount is provided for zero invoice
+        if provided_amount_msat == 0 && invoice_amount_msat == 0 {
+            return Err(SdkError::SendPaymentFailed {
+                err: "amount must be provided when paying a zero invoice".into(),
+            });
+        }
+
+        // Ensure amount is not provided for invoice that contains amount
+        if provided_amount_msat > 0 && invoice_amount_msat > 0 {
```
dangeross, line 249: Should we consider allowing if non-zero amounts match?

roeierez: I think it is better to be strict here. Greenlight doesn't allow it even if matches.

libs/sdk-core/src/breez_services.rs
```diff
@@ -235,6 +235,23 @@ impl BreezServices {
     ) -> SdkResult<Payment> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(bolt11.as_str())?;
+        let invoice_amount_msat = parsed_invoice.amount_msat.unwrap_or_default();
+        let provided_amount_msat = amount_sats.unwrap_or_default() * 1000;
```
JssDWt, line 239: Maybe this is a good time to change the method signature to amount_msat? I think we'd want to do that anyway?

libs/sdk-core/src/breez_services.rs
JssDWt: I think we should. If the amounts match, we should ignore the provided amount. Because the method signature takes an amount parameter, you expect you have to pass an amount there. Developers will have already parsed the invoice to show the amount to the user. So they already have the amount in a variable. It makes sense to pass it here. But that only really works if the input parameter is in millisatoshi, otherwise it only works full full satoshi amounts.

JssDWt: Ah, Roei's comment wasn't there yet when I answered. No strong opinion tbh.

roeierez: The fact the amount is parsed in the invoice doesn't necessary mean it is required in the `send_payment` and the signature takes optional amount.
Don't you think it feels weird to allow passing a parameter only in the case it is actually redundant? I mean how would we document such behavior? "Amount is allowed if it is the same as in the invoice"?
It would make sense IMO if the node supports passing amount that is greater or equal to the invoice amount but this is not the case here.
Isn't it more coherent (and clear to the user) behavior to require (and allow) amount only in case of zero invoice and avoid this redundancy?
What do you think?

roeierez: @JssDWt comments wasn't there when I answered :)
Anyway let me know your final thoughts.

JssDWt: @roeierez Yeah makes sense. We were getting quite a few questions about this particular issue. I think the error is much better now and should probably lead to less questions already. Maybe make it clearer in the docs too.

ok300: How about supporting both with `SendPaymentRequest` and `SendPaymentZeroAmountInvoice`?

ok300: Nevermind, that would complicate the service interface for very little benefit.

I agree with @JssDWt above.

libs/sdk-core/src/breez_services.rs
ok300: I agree, changing to `msat` is a step towards #417.

roeierez: I agree with both of you. I did change this method signature to msat. Not all other methods.
I think we have 2 more methods where we should change (I am not sure the onchain methods should use msat).
I suggest to do it in a separate PR.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1185,12 +1202,12 @@ impl BreezServices {
...
             .parse_filters(
                 r#"
-                info,
+                debug,
                 breez_sdk_core::input_parser=warn,
                 breez_sdk_core::backup=info,
                 breez_sdk_core::persist::reverseswap=info,
                 breez_sdk_core::reverseswap=info,
-                gl_client=warn,
+                gl_client=debug,
```
dangeross, line 1210: Did you mean to include log filter changes?

JssDWt (approved):
LGTM!

dangeross (approved):
Looks good. Just a comment about log filters below. As you're changing the send_payment interface, you could also move to a SendPaymentRequest struct

libs/sdk-core/src/breez_services.rs
roeierez: Good catch! Actually I didn't but I find myself always change that when I need to debug, I think we can push this change as it enables logging the signer and gl-client node logs.
What do you think?

dangeross: 👍  just making sure

ok300 (approved):

Issue #523, opened 2023-10-16
Default cltv delta for invoices should be 144 blocks


JssDWt:
Currently we take the default cltv delta of the node to create invoices. Greenlight is actually increasing the default cltv delta to 144 blocks, but it could make sense to put this in the SDK as well, to avoid surprises.

Problem is there is too little time to react if a htlc is not fully settled yet, but the signer is offline. Currently the default cltv delta is 9 blocks.

roeierez:
We control the cltv delta only in the case we open a channel. In this case we take it from the lsp information and add it in the routing hints.
I think this should be changed in the lsp info after greenlight changes it.

JssDWt:
@roeierez  This CLTV expiry is the final cltv expiry, which is defined by the recipient. There's also a CLTV delta for the LSP, but adding to the CLTV expiry of the recipient 'bumps' every expiry in the route. The cltv delta of the LSP is already set in the hop hint.

roeierez:
I see. Thanks for clarifying

PR #538, opened 2023-10-20
default cltv_expiry for invoices 144 blocks

JssDWt:
Currently the default cltv expiry of the node is used for invoices. If the default cltv expiry is too low, however, this increases force close risk with a signer that is not always online. Increase the default final cltv expiry to 144 blocks, so there's enough leeway for a client to come back online after a period of time and settle pending htlcs.

Fixes https://github.com/breez/breez-sdk/issues/523

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1792,7 +1792,7 @@ impl Receiver for PaymentReceiver {
                 request.preimage,
                 request.use_description_hash,
                 Some(expiry),
-                request.cltv,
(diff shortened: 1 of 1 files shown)


ok300 (commented):
Might be better to do this fallback inside the NodeAPI `create_invoice` implementation?

That way it covers all current and future calls to `create_invoice`.

dangeross (approved):
Looks good, I think keeping it out of the nope_api impl of create_invoice is good as long as its only called here.

Can also do: `request.cltv.or(Some(144))`

roeierez (approved):

PR #542, opened 2023-10-21
Ensure channels closed details are persisted

roeierez:
As part of optimizing the sync effort I discovered that the closed channel persisted details are not used which causes us to do unnecessary network requests.
Before this PR the channels closed details were always fetched since we persisted the channels fetched from the remote node without considering the `closed_at` and `closed_tx_id` that were already persisted.
The PR ensured this info is merged with the fetched channels before it is persisted preventing overriding this info with None values.


ok300 (commented):
Good catch.

Would it make sense to move it inside `update_channels`, as the new merging logic sort of belongs to that step?

ok300 (approved):

PR #543, opened 2023-10-22
Change RN function param naming to avoid naming clashes

dangeross:
Rerun the RN codegen to avoid function param naming clashes

Generator PR breez/breez-sdk-rn-generator#10


roeierez (approved):
LGTM

PR #546, opened 2023-10-23
Derive Clone trait for some fiat currency structs

ok300:
This is needed to be able to add a Rust snippet in the docs for the equivalent Kotlin snippet: https://sdk-doc.breez.technology/guide/fiat_currencies.html

In some of the loops, a few structs have to be cloned. This PR makes that possible.


roeierez (approved):
LGTM

Issue #526, opened 2023-10-17
Feature request: allow onchain fees to be known before committing to paying them


danielgranhao:
When either using any of the following methods:
* `send_onchain()`
* `sweep()`
* `refund()`

it would improve UX if we could know the fees that will be paid.

This could be done by splitting the methods into preparation and execution steps, such that after preparing, one can decide whether to execute or not based on the fees.

roeierez:
That's right @danielgranhao 
for sweep we are going to introduce prepareSweep: https://github.com/breez/breez-sdk/pull/294
For send_onchain we have https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/breez_services.rs#L630C18-L630C41
for refund perhaps we can add the refund tx weight to the `list_refundables` ?

danielgranhao:
> That's right @danielgranhao
for sweep we are going to introduce prepareSweep: https://github.com/breez/breez-sdk/pull/294
For send_onchain we have https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/breez_services.rs#L630C18-L630C41

Oh, sorry, completely missed that PR and the `fetch_reverse_swap_fees()` method. 

> for refund perhaps we can add the refund tx weight to the list_refundables ?

That would be nice!

danielgranhao:
@roeierez I was working on a PR for adding the refund tx weight to `list_refundables` and realized that this won't lead to an exact prediction of the fees. Without knowing the address the user chooses, the tx weight will just be a guess (of course, very close to the real one, and one could opt for the worst-case scenario).

Wouldn't it be better to go with a "prepare" approach (similar to https://github.com/breez/breez-sdk/pull/294) that would provide the exact fee that will be used?

JssDWt:
> For send_onchain we have https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/breez_services.rs#L630C18-L630C41

Link currently points to a different line of code.

roeierez:
Right @JssDWt I meant this one: https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/breez_services.rs#L647

JssDWt:
Permalinking that https://github.com/breez/breez-sdk/blob/3c602db1fb0799ad96922a618c6dbe180a722c5a/libs/sdk-core/src/breez_services.rs#L647

danielgranhao:
Closed by #294 and #548

PR #548, opened 2023-10-23
Implement `prepare_refund()`

danielgranhao:
Addresses #526.

### DIFF
--- a/libs/sdk-core/src/swap.rs
+++ b/libs/sdk-core/src/swap.rs
@@ -5,8 +5,8 @@ use crate::binding::parse_invoice;
 use crate::chain::{get_utxos, AddressUtxos, ChainService, MempoolSpace, OnchainTx};
 use crate::grpc::{AddFundInitRequest, GetSwapPaymentRequest};
 use crate::{
-    OpeningFeeParams, ReceivePaymentRequest, RefundRequest, RefundResponse,
-    SWAP_PAYMENT_FEE_EXPIRY_SECONDS,
+    OpeningFeeParams, PrepareRefundRequest, PrepareRefundResponse, ReceivePaymentRequest,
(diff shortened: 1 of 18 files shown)


libs/sdk-core/src/swap.rs
```diff
@@ -434,6 +434,31 @@ impl BTCReceiveSwap {
...
+        let swap_info = self
+            .persister
+            .get_swap_info_by_address(req.swap_address.clone())?
+            .ok_or_else(|| anyhow!(format!("swap address {} was not found", req.swap_address)))?;
+
+        let transactions = self
+            .chain_service
+            .address_transactions(req.swap_address.clone())
+            .await?;
+        let utxos = get_utxos(req.swap_address, transactions)?;
```
roeierez, line 450: Perhaps we can extract out this code to a function that can be reused in `refund_swap` function?

roeierez (changes requested):
Thanks a lot, looks good!
Can you please make sure the RN generator code is done with the latest RN generator? We yesterday updated it.

libs/sdk-core/src/swap.rs
danielgranhao: Makes sense. Addressed in [a2308ba](https://github.com/breez/breez-sdk/pull/548/commits/a2308ba88b4b60bc483a1187c071ce5224624c6b)

danielgranhao:
@roeierez 
> Can you please make sure the RN generator code is done with the latest RN generator? We yesterday updated it.

I've just rebased and run the generator again. I didn't see any changes.

dangeross (approved):
RN codegen looks good

roeierez:
@danielgranhao this is ready to be merged. We will merge it right after the conflict is resolved. Thanks.

danielgranhao:
@roeierez Conflicts should be resolved now :)

Issue #541, opened 2023-10-20
Upgrade flutter version


JssDWt:
The sdk-flutter binding currently requires flutter 3.7.12. People are using higher flutter versions for their applications though, making it hard for them to compile the sdk-flutter binding from source. Even though we have the https://github.com/breez/breez-sdk-flutter repo which they could use. If there's an issue with that, it would help if they were able to compile it themselves.

Therefore, update the used flutter version.

PR #549, opened 2023-10-23
Upgrade Flutter requirement to latest stable(3.13.8)

erdemyerebasmaz:
This PR addresses 
- #541 

We have been withholding upgrading Flutter as it requires the huge task of migrating `breezmobile` to null safety. Sticking to 3.7.12 helped us Breez developers to context switch between our Flutter projects, `c-breez`, `sdk-flutter` & `breezmobile` repositories with ease. 

Version management in Flutter is a bit lacking when it comes to this use case but there's a version management CLI tool called fenv, which makes switching between versions much easier. 

We decided forgo this ease in favor of appealing to other Flutter users that use our SDK and use fenv internally.

### Changelist:
- Update Flutter to 3.13.8
  -  Bump Dart range to  '>=3.0.0 <4.0.0'
  -  Bump Flutter range to  '>=3.10.0'
- Update flutter_rust_bridge to 1.80.1 & generate bindings
- Bump sdk-flutter version to 0.3.0

### TODO:
- Detail the changes in CHANGELOG.md


libs/sdk-flutter/pubspec.yaml
```diff
@@ -1,31 +1,31 @@
...
 description: A new flutter plugin project.
-version: 0.2.5
+version: 0.3.0
 publish_to: none
 
 environment:
-  sdk: ">=2.19.6 <3.0.0"
-  flutter: ">=3.7.12"
+  sdk: '>=3.0.0 <4.0.0'
+  flutter: ">=3.10.0"
```
ademar111190, line 8: As we publish a Sdk, keeping the minimum requirement as low as possible is interesting.
So, is it necessary to increase the **minimum** version?

erdemyerebasmaz: Yes, these are the minimum necessary versions.
In Dart 3.0, null safety is built in; you can not turn it off or have a middle-ground, e.g; no `--no-sound-null-safety` flag, and minimum Dart requirement has been upgraded to 3.0 as of Flutter 3.10.0.

erdemyerebasmaz: No, these are the minimum necessary versions due to `flutter_rust_bridge` requirements
`flutter_rust_bridge >=1.76.0 which requires SDK version >=3.0.0 <4.0.0`

which can be downgraded but I think we should move onto Flutter 3.10 & Dart 3.

ubbabeck (approved):
tACK

libs/sdk-flutter/pubspec.yaml
ademar111190: I see; thanks for checking out @erdemyerebasmaz and congrats; that was a hard work

ademar111190 (approved):

roeierez (approved):
LGTM

ubbabeck:
If you work on both breezmobile and c-breez  and have to switch between different flutter version i'd like to recommend [direnv](https://github.com/direnv/direnv) as an alternative to fenv.

erdemyerebasmaz:
> If you work on both breezmobile and c-breez and have to switch between different flutter version i'd like to recommend [direnv](https://github.com/direnv/direnv) as an alternative to fenv.

@breez/breez-developers You are free to use any solution you like for Flutter version management. I would only make sure to include any config files they may have to`.gitignore`.

ubbabeck:
> > If you work on both breezmobile and c-breez and have to switch between different flutter version i'd like to recommend [direnv](https://github.com/direnv/direnv) as an alternative to fenv.
> 
> @breez/breez-developers You are free to use any solution you like for Flutter version management. I would only make sure to include any config files they may have to`.gitignore`.

Good point, i'll open a pr for it.

PR #550, opened 2023-10-24
Add payment hash to `lnurl_pay()` return type

### LINKED ISSUE #540: Feature request: Return payment hash after trying to pay an LNURL-pay
In order to know which [`Payment`](https://breez.github.io/breez-sdk/breez_sdk_core/struct.Payment.html#) should be shown to the user after trying to pay an LNURL-pay, it would be useful if  [`LnUrlPayResult`](https://breez.github.io/breez-sdk/breez_sdk_core/enum.LnUrlPayResult.html#) would include a payment hash. 

This could be done in both success and error cases as long as the LNURL-pay flow progressed enough for the LNURL endpoint to return an invoice.

dleutenegger:
Resolves #540


libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -2997,18 +2997,23 @@ class BreezSDKMapper {
...
+        }
 
         throw SdkError.Generic(message: "Invalid enum variant \(type) for enum LnUrlPayResult")
     }
 
     static func dictionaryOf(lnUrlPayResult: LnUrlPayResult) -> [String: Any?] {
         switch lnUrlPayResult {
         case let .endpointSuccess(
-            data
+            datapaymentHash
```
erdemyerebasmaz, line 3011: Typo?

libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -3018,6 +3023,15 @@ class BreezSDKMapper {
                 "type": "endpointError",
                 "data": dictionaryOf(lnUrlErrorData: data),
             ]
+
+        case let .payError(
+            paymentHashreason
```
erdemyerebasmaz, line 3028: Typo?

dleutenegger: Seemed to be an issue with the rn codegen. I moved the `LnUrlPayResult` data into wrapper structs, which resolved the issue.

dangeross: I'll investigate the RN codegen issue

dangeross: I think your change is the preferred (common) way we handle interface variants anyway. But it seems variants with multiple tuple/struct like data attributes are not being rendered correctly in swift.

libs/sdk-core/src/breez_services.rs
```diff
@@ -339,7 +354,10 @@ impl BreezServices {
                 )?;
 
                 Ok(LnUrlPayResult::EndpointSuccess {
-                    data: maybe_sa_processed,
+                    data: LnUrlSuccessData {
+                        payment_hash: details.payment_hash.clone(),
+                        data: maybe_sa_processed,
```
dangeross, line 359: Can I propose we change `data` to `success_action` to avoid `data.data`?

dangeross (changes requested):
Looks good otherwise. Just a comment around naming

libs/sdk-core/src/lnurl/pay.rs
```diff
@@ -141,12 +140,21 @@ pub(crate) mod model {
...
     pub enum LnUrlPayResult {
-        EndpointSuccess {
-            data: Option<SuccessActionProcessed>,
-        },
-        EndpointError {
-            data: LnUrlErrorData,
-        },
+        EndpointSuccess { data: LnUrlSuccessData },
+        EndpointError { data: LnUrlErrorData },
+        PayError { data: LnUrlPayErrorData },
```
ok300, line 145: @dleutenegger Useful new error type! Can you please add a line about it in the method's rustdoc?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -444,10 +444,20 @@ dictionary BitcoinAddressData {
     string? message;
 };
 
+dictionary LnUrlSuccessData {
```
ok300, line 447: This only applies to LNURL-pay, so a more accurate name is `LnUrlPaySuccessData` IMO.

libs/sdk-core/src/breez_services.rs
```diff
@@ -297,10 +297,25 @@ impl BreezServices {
...
                     amount_msat: None,
                 };
-                let payment = self.send_payment(pay_req).await?.payment;
+
+                let payment = match self.send_payment(pay_req).await {
+                    Ok(p) => Ok(p),
+                    Err(SdkError::SendPaymentFailed { err }) => {
+                        let invoice = parse_invoice(cb.pr.as_str())?;
+
+                        return Ok(LnUrlPayResult::PayError {
```
ok300, line 309: Two points here:

- for consistency, I would remove the `return ...;` and just place the `Ok(..)` on the last line of the block, just like the other two branches of `match`
- is this really an `Ok(..)` and not an `Err(..)`?

libs/sdk-core/src/breez_services.rs
ok300: Would make sense IMO.

ok300 (changes requested):
Left a few comments, otherwise looks good!

libs/sdk-core/src/lnurl/pay.rs
roeierez: This makes me think that perhaps the right way here is to return the errors as rust errors as right now we are mixing both success and error in the success case (We don't use rust error here).
I do understand the change as the mixture has been existed before this change so I am OK to leave it as is but perhaps after we merge the error code PR @dangeross we can think of changing here to use rust errors if that make sense of course?

libs/sdk-core/src/breez_services.rs
dleutenegger: I think this needs to stay as is.

- The return is explicitly required as we want to exit the function and return the `PayError` to the caller.
- It must be an `Ok(..)` instead of an `Err(..)` as this is the current way the method handles errors, as pointed out by @roeierez https://github.com/breez/breez-sdk/pull/550#discussion_r1371648746.

libs/sdk-core/src/lnurl/pay.rs
dangeross: I don't think the error code PR would give a clean way to return the payment hash (and the reason) of the failed payment, which was the intent of the original issue.

libs/sdk-core/src/breez_services.rs
ok300: > The return is explicitly required as we want to exit the function and return the `PayError` to the caller.

We unwrap the result (and throw this error as well, if it would be wrapped as `Err`) just a few lines below in L317.

> It must be an `Ok(..)` instead of an `Err(..)` as this is the current way the method handles errors, as pointed out by @roeierez [Add payment hash to `lnurl_pay()` return type #550 (comment)](https://github.com/breez/breez-sdk/pull/550#discussion_r1371648746).

- The current way (before the PR): we are not mixing success and error in the success case
- The current way (in this PR): we are mixing them with the `Ok(..)` I mentioned above

So changing the `Ok(..)` into an `Err(..)` would address Roei's point around mixing result types. It would also make future migrations to `SdkError` simpler IMO.

ok300: Nevermind, my bad.

You want to return `LnUrlPayResult` in the success case, so `Ok(..)` is correct.

The generic `anyhow::Error` thrown now only takes a string as arg, so cannot be used to wrap anything more meaningful. The `send_payment` step already throws an `SdkError`, but that's converted back to `anyhow::Error` when unwrapping on L137.

All this can be handled better with `SdkError` in the future.

ok300 (approved):

dangeross (approved):
LGTM!

dangeross:
Seems there is a conflict @dleutenegger. Can you take a look?

dangeross:
Still some issues, can you rerun the flutter_rust_bridge?

erdemyerebasmaz:
> Seems there is a conflict @dleutenegger. Can you take a look?
> Still some issues, can you rerun the flutter_rust_bridge?

@dangeross This is likely due to latest Flutter SDK & flutter_rust_bridge_codegen upgrade.
 
@dleutenegger Could you rebase your branch on top of main first then run 
```
make clean
make init
make dev 
```

from sdk-flutter directory.

dangeross:
Hi @dleutenegger, would you mind rebasing and rerunning the RN codegen? I've just merged the error code PR #469 and there is a small conflict.

PR #553, opened 2023-10-24
Allow multiple logs listeners

ademar111190:
Using the broadcast stream controller to allow clients to have multiple listeners of log.

No QA Needed


roeierez (approved):
LGTM

Issue #302, opened 2023-07-07
send_payment can return ClosedChannel?


JssDWt:
The send_payment function returns a `Payment` object. In order to get the preimage for the payment, you need to go into `details`. `details` looks like this:

```
/// Wrapper for the different types of payments
#[derive(PartialEq, Eq, Debug, Clone, Deserialize, Serialize)]
#[serde(untagged)]
pub enum PaymentDetails {
    Ln {
        #[serde(flatten)]
        data: LnPaymentDetails,
    },
    ClosedChannel {
        #[serde(flatten)]
        data: ClosedChannelPaymentDetails,
    },
}
```

I think there is no way a payment will return `ClosedChannelPaymentDetails`? So this signature should only contain `LnPaymentDetails` to make the development easier.

JssDWt:
Superseded by https://github.com/breez/breez-sdk/issues/302

Issue #528, opened 2023-10-17
Feature request: persist swap information with resulting `Payment`


danielgranhao:
If a received LN payment results from a swap, we would like to present the payment slightly differently in the UI. One way of implementing that would be to include swap information in `LnPaymentDetails` similarly to how lnurl info is already included.

roeierez:
Makes sense, see my comment here: https://github.com/breez/breez-sdk/issues/527#issuecomment-1766946629

PR #555, opened 2023-10-25
Include `SwapInfo` in `Payment` resulting from swap

danielgranhao:
Addresses #528

TODO:
- [x] Bindings

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -261,6 +261,7 @@ dictionary LnPaymentDetails {
     string? lnurl_metadata;
     string? ln_address;
     string? lnurl_withdraw_endpoint;
+    SwapInfo? swap_info;
 };
 
 dictionary ClosedChannelPaymentDetails {
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1941,7 +1941,8 @@ pub(crate) mod tests {
     use crate::PaymentType;
     use crate::{
         input_parser, parse_short_channel_id, test_utils::*, BuyBitcoinProvider, BuyBitcoinRequest,
-        InputType, ListPaymentsRequest, PaymentStatus, ReceivePaymentRequest,
+        InputType, ListPaymentsRequest, OpeningFeeParams, PaymentStatus, ReceivePaymentRequest,
+        SwapInfo, SwapStatus,
     };
 
     use super::{PaymentReceiver, Receiver};
@@ -1964,6 +1965,37 @@ pub(crate) mod tests {
 
         let payment_hash_lnurl_withdraw = "2222";
         let payment_hash_with_lnurl_success_action = "3333";
+        let payment_hash_swap: Vec<u8> = vec![1, 2, 3, 4, 5, 6, 7, 8];
+        let swap_info = SwapInfo {
+            bitcoin_address: "123".to_string(),
+            created_at: 12345678,
+            lock_height: 654321,
+            payment_hash: payment_hash_swap.clone(),
+            preimage: vec![],
+            private_key: vec![],
+            public_key: vec![],
+            swapper_public_key: vec![],
+            script: vec![],
+            bolt11: Some("312".into()),
+            paid_sats: 1,
+            confirmed_sats: 1,
+            unconfirmed_sats: 0,
+            status: SwapStatus::Expired,
+            refund_tx_ids: vec![],
+            unconfirmed_tx_ids: vec![],
+            confirmed_tx_ids: vec![],
+            min_allowed_deposit: 5_000,
+            max_allowed_deposit: 1_000_000,
+            last_redeem_error: None,
+            channel_opening_fees: Some(OpeningFeeParams {
+                min_msat: 5_000_000,
+                proportional: 50,
+                valid_until: "date".to_string(),
+                max_idle_time: 12345,
+                max_client_to_self_delay: 234,
+                promise: "promise".to_string(),
+            }),
+        };
         let dummy_transactions = vec![
             Payment {
                 id: "1111".to_string(),
@@ -1985,6 +2017,7 @@ pub(crate) mod tests {
                         lnurl_metadata: None,
                         ln_address: None,
                         lnurl_withdraw_endpoint: None,
+                        swap_info: None,
                     },
                 },
             },
@@ -2008,6 +2041,7 @@ pub(crate) mod tests {
                         lnurl_metadata: None,
                         ln_address: None,
                         lnurl_withdraw_endpoint: Some(test_lnurl_withdraw_endpoint.to_string()),
+                        swap_info: None,
                     },
                 },
             },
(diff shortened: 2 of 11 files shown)


dangeross:
Outside of approval, I think it definitely makes more sense as part of a JOIN query and also adding an INDEX on the bolt11 column.

JssDWt:
I think it makes sense in the current form of the PaymentDetails. Agree with @dangeross on the JOIN, to make this a single querty. 

But tbh I'm not too fond of where the PaymentDetails are headed. It requires logic and knowledge of the different flows in order to extract the relevant information from the PaymentDetails now. It's already like that with the lnurl fields. It should probably be enum based, with a payment type and additional data for that specific type. Refactoring that may also fix https://github.com/breez/breez-sdk/issues/302. What do you think @roeierez?

libs/sdk-core/src/persist/swap.rs
```diff
@@ -245,6 +245,10 @@ impl SqliteStorage {
         self.select_single_swap("swaps.bitcoin_address = ?1", [address])
     }
 
+    pub(crate) fn get_swap_info_by_bolt11(&self, bolt11: &str) -> Result<Option<SwapInfo>> {
```
ok300, line 248: The existing `get_swap_info_by_hash` could probably be used here instead.

If we later add a `swaps_info` INDEX on `payment_hash`, it would benefit this and other callers too.

roeierez:
> But tbh I'm not too fond of where the PaymentDetails are headed. It requires logic and knowledge of the different flows in order to extract the relevant information from the PaymentDetails now. It's already like that with the lnurl fields. It should probably be enum based, with a payment type and additional data for that specific type. Refactoring that may also fix #302. What do you think @roeierez?

Can you elaborate? As far as I understand you are suggesting the details section to be enum but it is already an enum today (PaymentDetails). Or your mainly suggestion to add "type" to ease the differentiation?

libs/sdk-core/src/persist/swap.rs
roeierez: I agree we should use hash instead of bolt11. The main reason is because the bolt11 that exists on the swap is the one that the payer pays and the bolt11 that persisted with the payment is the one that the node settled. 
In case where a channel is opened as part of the swap these can be different but same hash.

JssDWt:
> Can you elaborate? As far as I understand you are suggesting the details section to be enum but it is already an enum today (PaymentDetails). Or your mainly suggestion to add "type" to ease the differentiation?

I think my suggestion would be to go for something like this:
- Maybe split a payment into a receiving side and a sending side, since they have different dynamics.
- Remove ClosedChannel from PaymentDetails, I think it doesn't belong there?
- Make the following fields part of `Payment`
```    
    pub payment_hash: String,
    pub label: String,
    pub destination_pubkey: String,
    pub payment_preimage: String,
    pub keysend: bool,
    pub bolt11: String,
```
- Have a details variant for lnurlp, lightning address, swap

Tbh I'm not sure what is the right way to make this interface for intuitive.

roeierez:
> > Can you elaborate? As far as I understand you are suggesting the details section to be enum but it is already an enum today (PaymentDetails). Or your mainly suggestion to add "type" to ease the differentiation?
> 
> I think my suggestion would be to go for something like this:
> 
> * Maybe split a payment into a receiving side and a sending side, since they have different dynamics.
> * Remove ClosedChannel from PaymentDetails, I think it doesn't belong there?
> * Make the following fields part of `Payment`
> 
> ```
>     pub payment_hash: String,
>     pub label: String,
>     pub destination_pubkey: String,
>     pub payment_preimage: String,
>     pub keysend: bool,
>     pub bolt11: String,
> ```
> 
> * Have a details variant for lnurlp, lightning address, swap
> 
> Tbh I'm not sure what is the right way to make this interface for intuitive.

I think the main confusion here is that the Payment is actually represents PaymentListItem.
We would like all items that affect lightning balance to belong to that list. That was the main guideline for creating one unified type.

JssDWt:
> I think the main confusion here is that the Payment is actually represents PaymentListItem.
> We would like all items that affect lightning balance to belong to that list. That was the main guideline for creating one unified type.

OK, it might still make sense to put the different types of payment details into their own enum variant, so you know a swap has swapdetails, an lnurl payment has lnurl payment details, etc. And you don't need to read the documentation in order to know what fields can exist when.

danielgranhao:
> OK, it might still make sense to put the different types of payment details into their own enum variant, so you know a swap has swapdetails, an lnurl payment has lnurl payment details, etc. And you don't need to read the documentation in order to know what fields can exist when.

I would also be in favor of something like this. Assuming the interface goes in that direction, would it make sense to still merge something like what's proposed here and work on the refactoring in another PR?

JssDWt:
> I would also be in favor of something like this. Assuming the interface goes in that direction, would it make sense to still merge something like what's proposed here and work on the refactoring in another PR?

Yes, makes sense

roeierez:
> > OK, it might still make sense to put the different types of payment details into their own enum variant, so you know a swap has swapdetails, an lnurl payment has lnurl payment details, etc. And you don't need to read the documentation in order to know what fields can exist when.
> 
> I would also be in favor of something like this. Assuming the interface goes in that direction, would it make sense to still merge something like what's proposed here and work on the refactoring in another PR?

Yes of course. Feel free to open an issue separately for discussion on changes/suggestions to the interface. We had some discussions internally and we are happy to open the discussion publicly and get more insights

libs/sdk-core/src/persist/swap.rs
danielgranhao: @ok300 @roeierez Sure, makes sense. I'm working on using a single query, but I'm having issues implementing the join due to the hash being stored as a hex string on the payments table and as a blob on the swaps table. Any suggestions on how to address this?

I see the following options:
* change the payment_hash column type to string on the swaps table (requiring encoding and decoding where necessary in the current places the table it's currently used)
* persist a serialized `SwapInfo` object in `payments_external_info` in a similar fashion to how other objects are persisted at the moment (this should be fine as long as we don't care about changes to the swap information after we've redeemed the swap)

ok300: Good catch.

We could also look into the [hex](https://www.sqlite.org/lang_corefunc.html#hex) and [unhex](https://www.sqlite.org/lang_corefunc.html#unhex) core SQLITE functions. Either of them could probably already be used in the join (e.g. `WHERE hex(payment_hash_blob) == payment_hash_hex`).

If we change the DB schema to converge on a common representation of hashes (for example hex), we could also use one of these functions to migrate data "on-the-fly" in a new migration step.

What do you think?

Btw if we change the schema, it might be better to split this in 2 steps:

- first, DB migration to common hash format, which also migrates existing data
- second, add this PR on top, which can then join without having to convert blob to hex or viceversa

I can do the migration PR if you'd like.

danielgranhao: > We could also look into the [hex](https://www.sqlite.org/lang_corefunc.html#hex) and [unhex](https://www.sqlite.org/lang_corefunc.html#unhex) core SQLITE functions. Either of them could probably already be used in the join (e.g. WHERE hex(payment_hash_blob) == payment_hash_hex).

Oh, I didn't know of these functions and didn't think of searching for them. Thanks!

> Alternatively, if we change the DB schema to converge on a common representation of hashes (for example hex), we could also use one of these functions to migrate data "on-the-fly" in a new migration step.

Hm, probably makes sense to switch to a common representation. Can you elaborate on why it's better to do the migration PR first? I think it would be fine to have a query using `hex()` and remove later.

I've just updated the PR to use the hex function within the payment fetching queries.

ok300: You're right, its also fine to use `hex()` for now and remove it later with the migration.

ok300 (commented):
In principle looks good, but the extra complexity in how the queries are built makes them harder to understand and reason about (in future refactorings, etc).

I personally would prefer not to add the complexity in the query building, even if that means no extra join and therefore one extra query.

What does everyone else think?

roeierez:
> In principle looks good, but the extra complexity in how the queries are built makes them harder to understand and reason about (in future refactorings, etc).
> 
> I personally would prefer not to add the complexity in the query building, even if that means no extra join and therefore one extra query.
> 
> What does everyone else think?

I have to say I tend to agree. Sorry I  had to go all the way just to recognize it. To me it looks like the complexity doesn't worth the join benefits as the previous code was much more simpler.

danielgranhao:
@ok300 @roeierez Sure, makes sense. Any performance implications are also not likely to be significant. I've updated the PR and rebased to make sure the binding generation is up to date.

roeierez (approved):
LGTM! Thanks for your patient @danielgranhao

ok300 (approved):
Looking good!

PR #559, opened 2023-10-26
Add Default trait, util constructor for Request args

ok300:
This PR simplifies the construction of user-facing Request structs:
- derives `Default` where possible
- adds `from_...` constructors for structs where only a few fields are always needed, but the rest are optional

Request structs where all fields are always needed, or where only 1 field is optional, were left out because the ergonomics wouldn't be improved by such a change.

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -751,16 +751,6 @@ pub struct ReceivePaymentRequest {
     pub cltv: Option<u32>,
 }
 
-impl ReceivePaymentRequest {
-    pub fn from_amount_and_description(amount_msat: u64, description: String) -> Self {
-        ReceivePaymentRequest {
-            amount_msat,
-            description,
-            ..Default::default()
-        }
-    }
-}
-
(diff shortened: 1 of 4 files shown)


erdemyerebasmaz (approved):
LGTM

JssDWt (approved):
Looks good.

The `from` implementations are nice, but not necessary. If they were removed, I think it's also good.

libs/sdk-core/src/models.rs
```diff
@@ -750,6 +751,16 @@ pub struct ReceivePaymentRequest {
     pub cltv: Option<u32>,
 }
 
+impl ReceivePaymentRequest {
+    pub fn from_amount_and_description(amount_msat: u64, description: String) -> Self {
```
roeierez, line 755: TBH I do understand the reason for this but I think it is not needed as the explicit way of passing the parameters is already short and easy enough. Unnecessary maintenance overhead IMO.

ok300:
Alright, then I removed the `from` constructors.

Issue #565, opened 2023-10-28
Unresolved reference: asPaymentTypeFilterList


dangeross:
In the RN generated Kotlin code for PaymentTypeFilter the asPaymentTypeFilterList is not implemented

https://github.com/breez/breez-sdk/blob/8bffae04a8717b27799b3599e8a7c9d57120486e/libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKMapper.kt#L734-L741

PR #566, opened 2023-10-28
check git status of react native and flutter

JssDWt:
Run the react native and flutter code generation in the main pull request CI and then run `git status` to see whether there is a diff. There shouldn't be a diff.

Also updates the used flutter version in CI. I noticed the old version was used.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -175,3 +175,95 @@ jobs:
           cargo clippy --tests -- -D warnings -A clippy::uninlined-format-args
           cd ../tools/sdk-cli
           cargo clippy -- -D warnings
+  
+  react-native:
+    name: Check react native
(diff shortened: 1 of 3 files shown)


JssDWt:
I love it that it caught a missed change already
https://github.com/breez/breez-sdk/actions/runs/6676790656/job/18146199727#step:8:21

dangeross:
This is great. A cool next step would be to test the RN generated code against built bindings so it could pickup issues like #565

JssDWt:
> This is great. A cool next step would be to test the RN generated code against built bindings so it could pickup issues like #565

That would be great too indeed. We need a way to compile as little as possible to check the bindings.

roeierez (approved):
LGTM

dangeross (approved):

PR #571, opened 2023-10-30
Fix RN workingDir and disconnect handling

dangeross:
Generated code for breez/breez-sdk-rn-generator#14

> This PR enables different nodes to be run through RN (one at a time) by allowing the workingDir to be set as a subdirectory of the application directory and resetting the breezServices instance when disconnect is called.

### DIFF
--- a/libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
+++ b/libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
@@ -35,6 +35,19 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
         throw SdkException.Generic("BreezServices not initialized")
     }
 
+    @Throws(SdkException::class)
+    private fun ensureWorkingDir(workingDir: String) {
+        try {
+            val workingDirFile = File(workingDir)
+
+            if (!workingDirFile.exists() && !workingDirFile.mkdirs()) {
+                throw SdkException.Generic("Mandatory field workingDir must contain a writable directory")
+            }
+        } catch (e: SecurityException) {
+            throw SdkException.Generic("Mandatory field workingDir must contain a writable directory")
+        }
+    }
+
     @ReactMethod
     fun addListener(eventName: String) {}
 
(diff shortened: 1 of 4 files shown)


JssDWt:
I think looks good.

It could be a bit racy in the `connect` with the gap between the breezServices null check and setting the breezServices if you try to connect two nodes at the same time? But that's not supported anyway, so don't change that.

Setting breezServices to null in disconnect looks fine. Working directory looks fine.

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -157,7 +153,18 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
...
+            if (!configTmp.workingDir.contains(reactApplicationContext.filesDir.toString())) {
+                promise.reject("Generic", "Mandatory field workingDir must contain application directory")
+                return
+            }
+
             val emitter = reactApplicationContext.getJSModule(RCTDeviceEventEmitter::class.java)
+            val workingDir = File(configTmp.workingDir)
+
+            if (!workingDir.exists()) {
+                workingDir.mkdirs()
```
roeierez, line 166: Perhaps we should check if that fails and throw exception?

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
```diff
@@ -157,7 +153,18 @@ class BreezSDKModule(reactContext: ReactApplicationContext) : ReactContextBaseJa
 
         try {
             val configTmp = asConfig(config) ?: run { throw SdkException.Generic("Missing mandatory field config of type Config") }
+
+            if (!configTmp.workingDir.contains(reactApplicationContext.filesDir.toString())) {
```
roeierez, line 157: Not sure if we should limit here the directory path.

roeierez (changes requested):

libs/sdk-react-native/android/src/main/java/com/breezsdk/BreezSDKModule.kt
dangeross: I understand we shouldn't limit too much, the application only has access to the application sandbox directory, documents directory and external file system. Throwing the above exception with a similar error message, would also suffice

dangeross: Updated to not limit the workingDir and provide a better error if creation of the workingDir fails

JssDWt (approved):
LGTM

roeierez (approved):
LGTM

PR #576, opened 2023-11-02
improve error on empty string invoice

JssDWt:
The current error when the caller inputs an empty string for the bolt11 inboice is `Invalid bech32: missing human-readable separator`. This is cryptic. Improve the error for the common case that the invoice is an empty string.

### DIFF
--- a/libs/sdk-core/src/invoice.rs
+++ b/libs/sdk-core/src/invoice.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use bitcoin::secp256k1::{self, PublicKey};
 use hex::ToHex;
 use lightning::routing::gossip::RoutingFees;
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -236,6 +236,11 @@ impl BreezServices {
         &self,
         req: SendPaymentRequest,
     ) -> Result<SendPaymentResponse, SendPaymentError> {
+        if req.bolt11.is_empty() {
```
dangeross, line 239: Perhaps this empty string check could be in the `parse_invoice` function?

JssDWt: Makes sense, done

libs/sdk-core/src/invoice.rs
```diff
@@ -189,6 +190,11 @@ pub fn add_lsp_routing_hints(
 
 /// Parse a BOLT11 payment request and return a structure contains the parsed fields.
 pub fn parse_invoice(bolt11: &str) -> InvoiceResult<LNInvoice> {
+    if bolt11.is_empty() {
```
ok300, line 193:
```suggestion
    if bolt11.trim().is_empty() {
```

Calling [trim()](https://doc.rust-lang.org/std/primitive.str.html#method.trim) also covers strings made up of only whitespace (tabs, space, etc).

ok300 (commented):
Small comment, otherwise LGTM.

libs/sdk-core/src/invoice.rs
JssDWt: 👍 done

PR #578, opened 2023-11-03
Add model docs: time fields are epoch time in seconds

ok300:
To avoid confusion with milliseconds, which is also a common timestamp unit.


erdemyerebasmaz (approved):

JssDWt:
@ok300 this broke CI, because the flutter bindings need updating

PR #586, opened 2023-11-06
Validate bolt11 network

### LINKED ISSUE #462: Feature request: add `Network` field to `LNInvoice`
At the moment, `parse()` returns successfully when decoding a testnet Bolt11 invoice, and there's no way for the consumer of the SDK to learn the network of an invoice before calling `send_payment()` without parsing the invoice using alternative means.

Having the network field would allow us to check it and warn the user the invoice is "invalid" due to a mismatch in networks even before he tries to pay.

dangeross:
This PR adds the invoice network to the LNInvoice response of `parse_input` and `parse_invoice`. It also validates the bolt11 invoice network against the config network when calling `send_payment` or `lnurl_pay`.

- [x] Validate send_payment
- [x] Validate lnurl_pay

Fixes #462

### DIFF
--- a/libs/sdk-core/src/invoice.rs
+++ b/libs/sdk-core/src/invoice.rs
@@ -195,18 +195,9 @@ pub fn add_lsp_routing_hints(
     Ok(invoice_builder.build_raw()?)
 }
 
-// Validate that the BOLT11 payment request matches the network
-pub fn validate_network(bolt11: &str, network: Network) -> InvoiceResult<()> {
-    if bolt11.trim().is_empty() {
-        return Err(InvoiceError::Validation(anyhow!(
-            "bolt11 is an empty string"
-        )));
-    }
-    let re = Regex::new(r"(?i)^lightning:")?;
-    let bolt11 = re.replace_all(bolt11, "");
-    let signed = bolt11.parse::<SignedRawInvoice>()?;
-    let invoice = Invoice::from_signed(signed)?;
-    match invoice.network() == network.into() {
+// Validate that the LNInvoice network matches the provided network
+pub fn validate_network(invoice: LNInvoice, network: Network) -> InvoiceResult<()> {
+    match invoice.network == network {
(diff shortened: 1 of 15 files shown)


ok300 (approved):

libs/sdk-core/src/invoice.rs
```diff
@@ -189,8 +195,17 @@ pub fn add_lsp_routing_hints(
...
+        true => Ok(()),
+        false => Err(InvoiceError::InvalidNetwork(anyhow!(
+            "Invoice network does not match config"
+        ))),
+    }
+}
+
 /// Parse a BOLT11 payment request and return a structure contains the parsed fields.
-pub fn parse_invoice(bolt11: &str) -> InvoiceResult<LNInvoice> {
+pub fn parse_invoice(bolt11: &str, network: Option<Network>) -> InvoiceResult<LNInvoice> {
```
roeierez, line 208: This change makes the binding interface different than the rust interface (we added one parameter only to the rust layer).
I am not sure we should overload the upper layer validation down to this function. Perhaps we should validate on the caller instead?

roeierez (changes requested):

libs/sdk-core/src/invoice.rs
```diff
@@ -189,6 +195,25 @@ pub fn add_lsp_routing_hints(
     Ok(invoice_builder.build_raw()?)
 }
 
+// Validate that the BOLT11 payment request matches the network
+pub fn validate_network(bolt11: &str, network: Network) -> InvoiceResult<()> {
+    if bolt11.trim().is_empty() {
```
roeierez, line 200: We duplicate the parse_invoice code here. Perhaps we should just use it and compare the network with the LNInvoice object?

roeierez (commented):
Looks good and simple! Dropped one simplification suggestion.

roeierez (approved):
LGTM

PR #589, opened 2023-11-06
Persist failed payment amount

### LINKED ISSUE #551: A failed `Payment` for a zero-amount invoice should include amount that was attempted to be sent
When paying a zero-amount invoice and it fails, there's no information about how much was attempted to be paid within the `Payment` struct.

Idea: persist the amount for the last failed attempt and include it in `Payment`.

dangeross:
Persists the last failed payment amount to the payment_external_info table.

Fixes #551

### DIFF
--- a/libs/sdk-core/src/persist/transactions.rs
+++ b/libs/sdk-core/src/persist/transactions.rs
@@ -15,7 +15,7 @@ impl SqliteStorage {
     ///
     /// Note that, if a payment has details of type [LnPaymentDetails] which contain a [SuccessActionProcessed],
     /// then the [LnPaymentDetails] will NOT be persisted. In that case, the [SuccessActionProcessed]
-    /// can be inserted separately via [SqliteStorage::insert_lnurl_payment_external_info].
+    /// can be inserted separately via [SqliteStorage::insert_payment_external_info].
     pub fn insert_or_update_payments(&self, transactions: &[Payment]) -> PersistResult<()> {
         let deleted = self.delete_pending_lightning_payments()?;
         debug!("Deleted {deleted} pending payments");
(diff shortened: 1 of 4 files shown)


dangeross:
Think I need to consider how this will be remotely synced also

roeierez:
> Think I need to consider how this will be remotely synced also

If the amount is needed to be synced then it should be written to the sync db such as when we write to `payments_external_info` table

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -233,11 +238,16 @@ impl SqliteStorage {
...
         let amount_msat = row.get(3)?;
+        let status: PaymentStatus = row.get(5)?;
+        let failed_amount_msat: Option<u64> = row.get(12)?;
         let mut payment = Payment {
             id: row.get(0)?,
             payment_type: PaymentType::from_str(payment_type_str.as_str()).unwrap(),
             payment_time: row.get(2)?,
-            amount_msat,
+            amount_msat: match status {
+                PaymentStatus::Failed => failed_amount_msat.unwrap_or_default(),
```
roeierez, line 248: Don't we have a migration issue for existing failed payments (even with non zero invoices) that currently will get zero value here?

roeierez (changes requested):
Looks good! Just need to clarify the migration.

libs/sdk-core/src/persist/transactions.rs
dangeross: True, it would be 0 for previously failed payments. To my knowledge its not possible to retreive failed payment amounts from GL. The alternative is to add failed_amount_msat: Option<u64> to Payment directly. But it will still not have a value for a failed payment.

roeierez: I am not concerned about existing failed payments of zero invoices. We can't control these but I am worried about existing failed payments of invoices with amount. Shouldn't we fall back to the existing invoice amount to solve this?

dangeross: Ah, that's a good idea. But it is possible as a migration step? I'll look into it

roeierez (approved):
LGTM

ok300 (approved):
A minor suggestion, otherwise looks good.

PR #592, opened 2023-11-07
Fallback to a listed LSP if the default is not available

### LINKED ISSUE #590: No LSP found for id
When an API key is configured to have LSPs that does not include the Breez LSP, an error is always thrown.

```No LSP found for id 03cea51f-b654-4fb0-8e82-eca137f236a0```

This is because the Breez LSP is used as the Config default_lsp_id which is persisted when the BreezService is built then used to get the LSP during receive_payment. This should instead use an LSP from the result of list_lsps to set a default if the Breez LSP is not in the list.

### PR DESCRIPTION
**Pull request by dangeross**
Fallback to the first LSP in the listed LSPs if the persisted LSP is not available. This needs to happen if the Config's default_lsp_id or the currently persisted LSP id is not returned in the list_lsps result.

Fixes #590

### CODE REVIEW

**Review by ok300 — COMMENTED** [review]
With this PR, we have two service methods

- `lsp_id()` (which looks up the current LSP ID from the persister)
- `lsp_info()` (which uses the changed `get_lsp()`)

that sound like they do the same thing but return different parts of the same LSP, but actually they're different.

IMO we should keep them consistent.

Also, if `get_lsp()` now automatically sets an LSP, do we still need the `connect_lsp()` service method?

File: libs/sdk-core/src/breez_services.rs [hunk 1]
```diff
@@ -1896,26 +1896,49 @@ impl Receiver for PaymentReceiver {
...
 
-/// Convenience method to look up LSP info based on current LSP ID
-async fn get_lsp(persister: Arc<SqliteStorage>, lsp: Arc<dyn LspAPI>) -> Result<LspInformation> {
+/// Convenience method to look up LSP info based on current LSP ID and
+/// fallback to the first LSP in the listed LSPs if the current LSP is not available.
+async fn get_lsp(
+    persister: Arc<SqliteStorage>,
+    lsp_api: Arc<dyn LspAPI>,
+) -> Result<LspInformation> {
     let lsp_id = persister.get_lsp_id()?.ok_or(anyhow!("No LSP ID found"))?;
```
roeierez, line 1905: Shouldn't we continue here and try to connect to the first lsp from the list?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
dangeross: Wouldn't it always be set from the config default_lsp_id?... Ah, it could be that it's also set to None right

ok300:
> I think `connect_lsp()` is still needed because it could be that there are multiple LSPs available and they can be switched.

Agree.

> Yeah, with this PR `get_lsp()` is also ensuring that if the persisted LSP is not available, it will fallback to another 

This scenario then becomes possible:

- calling `sdk.lsp_id()?` returns the ID of the persisted LSP, which can be `None`
- calling `sdk.lsp_info()?` applies the new logic: finds `None`, then looks up a new LSP and persists it
- calling `sdk.lsp_id()?` again now returns the new LSP ID

IMO this can be confusing. After all, both these methods look like getters.

I agree that some fallback mechanism can be useful though. Why not add a helper like `get_or_set_lsp_info()` / `get_or_init_lsp_info()`? Would cover your use-case, plus it's obvious it does more than getting, it doesn't change the behavior of existing service methods (keeps backward-compatibility) and avoids any confusion.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1896,26 +1896,49 @@ impl Receiver for PaymentReceiver {
     }
 }
 
-/// Convenience method to look up LSP info based on current LSP ID
-async fn get_lsp(persister: Arc<SqliteStorage>, lsp: Arc<dyn LspAPI>) -> Result<LspInformation> {
+/// Convenience method to look up LSP info based on current LSP ID and
+/// fallback to the first LSP in the listed LSPs if the current LSP is not available.
+async fn get_lsp(
```
roeierez, line 1901: It seems that if we change the selected lsp in this method it may  change it even if the user didn't set it directly.
For example:
1. User set explicit lsp id in the config
2. Once sdk is connected the lsp changes without the user being aware of it. Perhaps the expected behavior in this case is to actually get an error "lsp not found"

I am proposing here an alternative:
1. On our default config remove the constant lsp_id (leave None). Let the user set it if desired.
2. Make the change to select dynamic lsp if not already selected at the `connect_lsp_peer` function that is invoked automatically on every sync and on startup.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
dangeross: Makes sense because it keeps the logic in the connect_lsp_peer fn. Have issues with it at the moment as the node_state is not persisted at that point, which is used to fetch the lsp list

libs/sdk-core/src/breez_services.rs
```diff
@@ -768,6 +768,9 @@ impl BreezServices {
         // update node state and channels state
         self.persister.set_node_state(&new_data.node_state)?;
 
+        // Set the default LSP, if not already
+        self.set_default_lsp_id().await?;
```
roeierez, line 772: @dangeross  Did you see that the `connect_lsp_peer` is called above (line 754)?
I think it would be clean to encapsulate all the logic in that function. What do you think?

dangeross: I separated it out because on the first sync/run the node state isn't set yet during the connect_lsp_peer call

dangeross: If encapsulating it in `connect_lsp_peer` and handling that the node state is not persisted yet, it would mean having to call `connect_lsp_peer` twice, the second being after persisting the node state.

dangeross:
Updated to get the node_pubkey from node_api.start(), then pass the node_pubkey to connect_lsp_peer() to set a default LSP if one is not already set.

roeierez (approved):
LGTM

libs/sdk-core/src/breez_services.rs
```diff
@@ -803,7 +803,13 @@ impl BreezServices {
     }
 
     /// Connects to the selected LSP, if any
```
ok300, line 805:
```suggestion
    /// Connects to the selected LSP peer. If none is selected, this selects the first one from [`list_lsps`] and persists the selection.
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -1896,26 +1903,30 @@ impl Receiver for PaymentReceiver {
     }
 }
 
-/// Convenience method to look up LSP info based on current LSP ID
-async fn get_lsp(persister: Arc<SqliteStorage>, lsp: Arc<dyn LspAPI>) -> Result<LspInformation> {
+/// Convenience method to look up LSP info based on current LSP ID and
+/// fallback to the first LSP in the listed LSPs if the current LSP is not available.
```
ok300, line 1907:
```suggestion
/// Convenience method to look up LSP info based on current LSP ID.
/// If none is selected, it selects and persists the first LSP in the listed LSPs as fallback.
```

ok300 (approved):
A few rustdoc nits for clarity, otherwise LGTM.

libs/sdk-core/src/breez_services.rs
dangeross: I reverted the comment, `get_lsp` behaves as before

libs/sdk-core/src/breez_services.rs
dangeross: Done

Issue #433, opened 2023-09-07
Add a global status API


kingonly:
In case of system is down, general health check.

ok300:
You mean as a webservice on our domain, or as part of the SDK?

kingonly:
As part of the SDK

dangeross:
Fixed in #628

PR #594, opened 2023-11-07
Add lightweight connectivity check on connect()

ok300:
This PR updates `connect()` to include an active, lightweight connectivity check for the various services used.

Before, `connect()` would only create services, to be used when necessary in later calls. A connectivity issue would have only been seen later on, when a sub-service would be used.

An exception was the Greenlight service, which made an active connection during `connect` for `register` and `recover`. In those cases, any connectivity issues to the scheduler would have thrown an error in our `connect`. However, if the node was already registered and credentials already known, no connection to the Greenlight Scheduler was made during our `connect`. With this PR, the Greenlight connectivity is checked regardless of the node init method.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -24,6 +24,7 @@ use tonic::{Request, Status};
 
 use crate::backup::{BackupRequest, BackupTransport, BackupWatcher};
 use crate::chain::{ChainService, MempoolSpace, Outspend, RecommendedFees};
+use crate::connectivity::NeedsConnectivity;
 use crate::error::{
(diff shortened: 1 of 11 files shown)


ok300:
Some (all?) of these checks could be outsourced to Breez Server via a global status API (https://github.com/breez/breez-sdk/issues/433#issuecomment-1709891477).

Maybe that's a good compromise to consider:
- pros: it reduces the connection checks to one connection and therefore speed up `connect`
- cons: it checks connectivity from the Breez Server to these endpoints, not from the caller

The cons would be acceptable IMO, especially for mobile devices with slow or unreliable data network.

ok300:
Superseded by #596.

PR #596, opened 2023-11-08
Schedule signer if `GL::connect()` has seed and creds

ok300:
Include a GL API call on `Greenlight::connect`, for the case when we already know the seed node credentials.

This ensures network connectivity is there and starts a signer, so our `sdk.connect()` returns only with a running signer.


ok300:
When trying to `connect` without network, we immediately get an error:

```
sdk> connect

Error: Service connectivity: (Greenlight) Failed to connect: Service connectivity: status: Unavailable, message: "error trying to connect: dns error: failed to lookup address information: Name or service not known", details: [], metadata: MetadataMap { headers: {} }
```

roeierez (approved):
LGTM

Issue #574, opened 2023-11-01
Roadmap to payment notifications feature


ok300:
###  Issues found

- [x] Background thread timeout has limited effect on physical Android
- [x] Occasional "node not initialized" error when app was stopped and notification wakes background thread
- [x] Occasional empty history when app is opened after notification triggered background thread
  - Solved by https://github.com/breez/c-breez/pull/702
- [x] Errors thrown by `connect` take between 30s - 1m 30s to propagate to the UI (Android)

Low prio

- [ ] Validate the notification before processing it
  - if we receive a notification for an unknown payment_hash, the background thread still keeps checking that payment

### Next steps

- [x] LSPD: Add LSPD notification endpoints to Breez Server
- [x] LSPD: Handle stale URLs on LSP, while still supporting multi-device: https://github.com/breez/lspd/pull/147
- [x] SDK: Update notifications branch (SDK) to use Breez Server notifications
  - Pending https://github.com/breez/breez-sdk/pull/597 
- [x] c-breez: Integrate notification calls: https://github.com/breez/c-breez/pull/717
- [x] c-breez: iOS extension for c-breez: https://github.com/breez/c-breez/issues/562
- [x] c-breez: Allow c-breez to stop the GL node: https://github.com/breez/c-breez/pull/720
- [x] Docs / snippets: how to setup: https://github.com/breez/breez-sdk-docs/pull/108

### Open questions

- [x] "Send notifications to multiple endpoints" https://github.com/breez/breez-sdk/issues/112
- [x] Webhooks: possible on tor? Not at the moment.
- [ ] When more notification types are supported:
  - How should the SDK `register_webhook` signature look like? Should it take as arg the desired notification types?
  - Same for LSP: different method for each notification type?
- [x] How to handle the 5 minute time-window, where the LSP may see the node as online but the signer might be offline, and therefore notifications may not be sent

### Future improvements

- https://github.com/breez/lspd/issues/103

JssDWt:
Breez server now has the PaymentNotifier service:
https://github.com/breez/server/blob/d3d5dc9ea267d9c61ef9b4e9ec9096e63bbe1e08/breez/breez.proto#L83-L85

The signature is a recoverable signature over the double hash of the url. It should be signed by the node id, so lspd can extract the node id from the signature.

PR #597, opened 2023-11-08
Add `register_notifications()` service call

ok300:
Expose `register_notifications()` as SDK service.

- [x] Bindings
- [x] Rustdoc for service method
- [ ] SDK doc example / snippet

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1370,13 +1370,13 @@ impl BreezServices {
         })
     }
 
-    /// Register for webhook callbacks at the given `callback_url` whenever a new payment is received.
+    /// Register for webhook callbacks at the given `webhook_url` whenever a new payment is received.
     ///
     /// More webhook types may be supported in the future.
-    pub async fn register_webhook(&self, callback_url: String) -> SdkResult<()> {
+    pub async fn register_webhook(&self, webhook_url: String) -> SdkResult<()> {
         info!("Registering for webhook notifications");
 
-        let message = callback_url.clone();
+        let message = webhook_url.clone();
         let sign_request = SignMessageRequest { message };
         let sign_response = self.sign_message(sign_request).await?;
 
@@ -1385,7 +1385,7 @@ impl BreezServices {
             .register_payment_notifications(
                 lsp_info.id,
                 lsp_info.lsp_pubkey,
-                callback_url,
+                webhook_url,
                 sign_response.signature,
             )
             .await?;
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -33,9 +33,11 @@ use crate::greenlight::{GLBackupTransport, Greenlight};
 use crate::grpc::channel_opener_client::ChannelOpenerClient;
 use crate::grpc::fund_manager_client::FundManagerClient;
 use crate::grpc::information_client::InformationClient;
+use crate::grpc::payment_notifier_client::PaymentNotifierClient;
 use crate::grpc::signer_client::SignerClient;
 use crate::grpc::swapper_client::SwapperClient;
-use crate::grpc::PaymentInformation;
+use crate::grpc::SubscribeNotificationsRequest;
+use crate::grpc::{PaymentInformation, RegisterPaymentNotificationResponse};
 use crate::invoice::{add_lsp_routing_hints, parse_invoice, LNInvoice, RouteHint, RouteHintHop};
 use crate::lnurl::auth::perform_lnurl_auth;
 use crate::lnurl::pay::model::SuccessAction::Aes;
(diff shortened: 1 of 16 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1370,36 @@ impl BreezServices {
             },
         })
     }
+
+    pub async fn register_notifications(
```
roeierez, line 1374: Isn't it better to get the url directly from the caller? Also perhaps we want to rename this function to `register_webhook` ?
The rational is:
1. This should also be used as a general web hook, not only for mobile
2. The vendor that is running the NDS may decide on a different url format than we introduce here.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1370,36 @@ impl BreezServices {
...
+
+        let message = url.clone();
+        let sign_request = SignMessageRequest { message };
+        let sign_response = self.sign_message(sign_request).await?;
+
+        let subscribe_request = SubscribeNotificationsRequest {
+            url,
+            signature: hex::decode(sign_response.signature)
+                .map_err(|e| anyhow!("Failed to decode signature: {e}"))?,
+        };
```
roeierez, line 1393: Perhaps this logic should go int the lsp_api?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: > This should also be used as a general web hook

Then we might need an extra arg for the type of event that triggers this webhook, like so:

```rust
pub enum WebhookEvent {
    PaymentReceived
}

register_webhook(&self, callback_url: String, event: WebhookEvent) -> SdkResult<()>
```

That might also be needed because we cannot call the same `callback_url` for every notification type, because they have different payloads so the caller needs to choose an appropriate `callback_url`.

Would you agree?

libs/sdk-core/src/breez_services.rs
ok300: Makes sense, done.

libs/sdk-core/src/breez_services.rs
dangeross: Alternatively the event type could be included in the call to the webhook url:

```
<base_url>?token=<token>&type=<type>
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1370,36 @@ impl BreezServices {
...
+    pub async fn register_notifications(
+        &self,
+        platform: String,
+        token: String,
+    ) -> SdkResult<RegisterPaymentNotificationResponse> {
+        info!("Registering notification");
+
+        // TODO Make configurable?
+        let base_url = "https://notifier.breez.technology";
+        let url = format!("/api/v1/notify{base_url}?platform={platform}&token={token}");
```
dangeross: Should this be?

```suggestion
        let url = format!("{base_url}/api/v1/notify?platform={platform}&token={token}");
```

libs/sdk-core/src/grpc/proto/breez.proto
```diff
@@ -395,4 +412,16 @@ message PaymentInformation {
...
+}
+
+
+/////////////////////////////////////////////
+// From notifications.proto
+/////////////////////////////////////////////
+
+message SubscribeNotificationsRequest {
+    string url = 1;
+    bytes signature = 2;
```
JssDWt, line 424: Changed this to string on the LSP side

JssDWt (commented):
Looks good

libs/sdk-core/src/lsp.rs
```diff
@@ -112,6 +115,37 @@ impl LspAPI for BreezServer {
...
+            signature: callback_url_signature,
+        };
+
+        let mut client = self.get_subscription_client().await?;
+
+        let mut buf = Vec::new();
+        buf.reserve(subscribe_request.encoded_len());
+        subscribe_request
+            .encode(&mut buf)
+            .map_err(|e| SdkError::ServiceConnectivity {
```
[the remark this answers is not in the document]
dangeross: `SdkError::Generic` is the only alternative here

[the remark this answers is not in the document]
ok300: Changed to `SdkError::Generic`

libs/sdk-core/src/grpc/proto/breez.proto
ok300: Done, updated to use String.

libs/sdk-core/src/breez_services.rs
ok300: Good point. Both are valid options.

I've added it as an open question in the umbrella issue #574. IMO for now we can leave it as it is, we only have one notification type.

roeierez (approved):
Small Nit otherwise LGTM!

JssDWt:
Let's wait for the lsp change to deploy to be able to test this?

dangeross (approved):

JssDWt (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1369,29 @@ impl BreezServices {
             },
         })
     }
+
+    /// Register for webhook callbacks at the given `callback_url` whenever a new payment is received.
+    ///
+    /// More webhook types may be supported in the future.
+    pub async fn register_webhook(&self, callback_url: String) -> SdkResult<()> {
```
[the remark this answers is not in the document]
ok300: Done

ok300:
> Let's wait for the lsp change to deploy to be able to test this?

@JssDWt I'll merge one the CI tests go through, but will check with you before calling it in notification tests.

PR #598, opened 2023-11-09
adding support for bolt12 offers

### LINKED ISSUE #562: Support paying BOLT12 offers
- [ ] add Bolt12 variant to the input parser. Make it return at least all the necessary fields to be able to call [fetchinvoice](https://docs.corelightning.org/reference/lightning-fetchinvoice).
- [ ] add fetchinvoice to NodeAPI
- [ ] Make a send_payment flow for bolt12. It calls fetchinvoice with the supplied parameters (sane defaults) and then pays the returned invoice

hydra-yse:
resolves #562 1,2
_Note_: In order to support these changes, the `gl_client` crate in sdk-core currently points to a locally-owned repository. Once greenlight and cln_grpc are both updated to support these features, we can opt-out back to the original crates. I suggest changes should be moved into a side branch until then.

### DIFF
--- a/libs/Cargo.toml
+++ b/libs/Cargo.toml
@@ -23,10 +23,10 @@ base64 = "0.13.0"
 bitcoin = "=0.29.2" # Same version as used in gl-client
 # Pin the reqwest dependency until macOS linker issue is fixed: https://github.com/seanmonstar/reqwest/issues/2006
 hex = "0.4"
-lightning = "=0.0.116" # Same version as used in gl-client
-lightning-invoice = "=0.24.0" # Same version as used in gl-client
+lightning = "=0.0.118" # Same version as used in gl-client
+lightning-invoice = "=0.26.0" # Same version as used in gl-client
 log = "0.4"
-mockito = "1"
+mockito = "1.4.0"
 once_cell = "1"
 prost = "^0.11"
 regex = "1.8.1"
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -91,6 +91,7 @@ enum SendPaymentError {
     "InvalidInvoice",
     "InvoiceExpired",
     "InvalidNetwork",
+    "OfferChanged",
     "PaymentFailed",
     "PaymentTimeout",
     "RouteNotFound",
@@ -481,6 +482,21 @@ dictionary StaticBackupResponse {
     sequence<string>? backup;
 };
 
+dictionary CreateOfferRequest {
+    string description;
+    u64? amount_msat = null;
+    u64? absolute_expiry = null;
+    u64? quantity_max = null;
+};
+
+dictionary PayOfferRequest {
+    string offer;
+    u64? amount_msat = null;
+    f64? timeout = null;
+    string? payer_note = null;
+    string? label = null;
+};
+
 dictionary ReceiveOnchainRequest {
     OpeningFeeParams? opening_fee_params = null;
 };
@@ -717,9 +733,26 @@ dictionary RecommendedFees {
 };
 
 [Enum]
+interface Amount {
+    Bitcoin(u64 amount_msat);
+    Currency(string iso4217_code, u64 fractional_amount);
+};
+
+dictionary LNOffer {     
+    string bolt12;
+    sequence<string> chains;
+    string description;
+    string signing_pubkey;
+    Amount? amount;  
+    u64? absolute_expiry;
+    string? issuer;
+};
+
+[Enum]
 interface InputType {
     BitcoinAddress(BitcoinAddressData address);
     Bolt11(LNInvoice invoice);
+    Bolt12Offer(LNOffer offer);
     NodeId(string node_id);
     Url(string url);
     LnUrlPay(LnUrlPayRequestData data);
@@ -752,7 +785,7 @@ dictionary RedeemOnchainFundsResponse {
 };
 
 dictionary SendPaymentRequest {
-    string bolt11;
+    string invoice;
     u64? amount_msat = null;
     string? label = null;
 };
@@ -967,6 +1000,12 @@ interface BlockingBreezServices {
 
    [Throws=SdkError]
    PrepareRedeemOnchainFundsResponse prepare_redeem_onchain_funds(PrepareRedeemOnchainFundsRequest req);
+
+   [Throws=SdkError]
+   string create_offer(CreateOfferRequest req);
+
+   [Throws=SendPaymentError]
+   SendPaymentResponse pay_offer(PayOfferRequest req);
 };
 
 namespace breez_sdk {  
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -261,7 +261,7 @@ impl BreezServices {
         req: SendPaymentRequest,
     ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
-        let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
+        let parsed_invoice = parse_invoice(req.invoice.as_str())?;
         let invoice_amount_msat = parsed_invoice.amount_msat.unwrap_or_default();
         let provided_amount_msat = req.amount_msat.unwrap_or_default();
 
@@ -292,11 +292,7 @@ impl BreezServices {
                 self.persist_pending_payment(&parsed_invoice, amount_msat, req.label.clone())?;
                 let payment_res = self
                     .node_api
-                    .send_payment(
-                        parsed_invoice.bolt11.clone(),
-                        req.amount_msat,
-                        req.label.clone(),
-                    )
+                    .send_payment(req.invoice.clone(), req.amount_msat, req.label.clone())
                     .map_err(Into::into)
                     .await;
                 let payment = self
@@ -356,7 +352,7 @@ impl BreezServices {
             }
             ValidatedCallbackResponse::EndpointSuccess { data: cb } => {
                 let pay_req = SendPaymentRequest {
-                    bolt11: cb.pr.clone(),
+                    invoice: cb.pr.clone(),
                     amount_msat: None,
                     label: req.payment_label,
                 };
@@ -1995,6 +1991,44 @@ impl BreezServices {
           ***Swaps***\n{swaps}\n\n"
         );
         Ok(res)
+    }
+
+    pub async fn create_offer(&self, req: CreateOfferRequest) -> Result<String> {
+        self.start_node().await?;
+        Ok(self.node_api.create_offer(req).await?)
+    }
+
+    pub async fn pay_offer(
+        &self,
+        req: PayOfferRequest,
+    ) -> Result<SendPaymentResponse, SendPaymentError> {
+        self.start_node().await?;
+
+        let fetch_invoice_response = self
+            .node_api
+            .fetch_invoice(FetchInvoiceRequest {
+                offer: req.offer.clone(),
+                amount_msat: req.amount_msat,
+                timeout: req.timeout,
+                payer_note: req.payer_note.to_owned(),
+            })
+            .await?;
+
+        if let Some(_) = fetch_invoice_response.changes {
+            return Err(SendPaymentError::OfferChanged {
+                err: "Offer changed".to_string(),
+            });
+        }
+
+        Ok(self
+            .send_payment(SendPaymentRequest {
+                invoice: fetch_invoice_response.bolt12,
+                // we expect amount_msat from the invoice to be the same as the one in the
+                // PayOfferRequest due to the above check, can omit
+                amount_msat: None,
+                label: req.label,
+            })
+            .await?)
     }
 }
 
--- a/libs/sdk-core/src/bridge_generated.rs
+++ b/libs/sdk-core/src/bridge_generated.rs
@@ -41,6 +41,7 @@ use crate::models::ClosedChannelPaymentDetails;
 use crate::models::Config;
 use crate::models::ConfigureNodeRequest;
 use crate::models::ConnectRequest;
+use crate::models::CreateOfferRequest;
 use crate::models::EnvironmentType;
 use crate::models::GreenlightCredentials;
 use crate::models::GreenlightDeviceCredentials;
@@ -59,6 +60,7 @@ use crate::models::OpenChannelFeeRequest;
 use crate::models::OpenChannelFeeResponse;
 use crate::models::OpeningFeeParams;
 use crate::models::OpeningFeeParamsMenu;
+use crate::models::PayOfferRequest;
 use crate::models::PayOnchainRequest;
 use crate::models::PayOnchainResponse;
 use crate::models::Payment;
@@ -528,6 +530,32 @@ fn wire_receive_payment_impl(
         },
     )
 }
+fn wire_create_offer_impl(port_: MessagePort, req: impl Wire2Api<CreateOfferRequest> + UnwindSafe) {
+    FLUTTER_RUST_BRIDGE_HANDLER.wrap::<_, _, _, String, _>(
+        WrapInfo {
+            debug_name: "create_offer",
+            port: Some(port_),
+            mode: FfiCallMode::Normal,
+        },
+        move || {
+            let api_req = req.wire2api();
+            move |task_callback| create_offer(api_req)
+        },
+    )
+}
+fn wire_pay_offer_impl(port_: MessagePort, req: impl Wire2Api<PayOfferRequest> + UnwindSafe) {
+    FLUTTER_RUST_BRIDGE_HANDLER.wrap::<_, _, _, SendPaymentResponse, _>(
+        WrapInfo {
+            debug_name: "pay_offer",
+            port: Some(port_),
+            mode: FfiCallMode::Normal,
+        },
+        move || {
+            let api_req = req.wire2api();
+            move |task_callback| pay_offer(api_req)
+        },
+    )
+}
 fn wire_lnurl_pay_impl(port_: MessagePort, req: impl Wire2Api<LnUrlPayRequest> + UnwindSafe) {
     FLUTTER_RUST_BRIDGE_HANDLER.wrap::<_, _, _, LnUrlPayResult, _>(
         WrapInfo {
@@ -895,6 +923,9 @@ pub struct mirror_AesSuccessActionDataDecrypted(AesSuccessActionDataDecrypted);
 pub struct mirror_AesSuccessActionDataResult(AesSuccessActionDataResult);
 
 #[derive(Clone)]
+pub struct mirror_Amount(Amount);
+
+#[derive(Clone)]
 pub struct mirror_BitcoinAddressData(BitcoinAddressData);
 
 #[derive(Clone)]
@@ -908,6 +939,9 @@ pub struct mirror_InputType(InputType);
 
 #[derive(Clone)]
 pub struct mirror_LNInvoice(LNInvoice);
+
+#[derive(Clone)]
+pub struct mirror_LNOffer(LNOffer);
 
 #[derive(Clone)]
 pub struct mirror_LnUrlAuthRequestData(LnUrlAuthRequestData);
@@ -979,6 +1013,18 @@ const _: fn() = || {
             let _: String = reason;
         }
     }
+    match None::<Amount>.unwrap() {
+        Amount::Bitcoin { amount_msat } => {
+            let _: u64 = amount_msat;
+        }
+        Amount::Currency {
+            iso4217_code,
+            fractional_amount,
+        } => {
+            let _: String = iso4217_code;
+            let _: u64 = fractional_amount;
+        }
+    }
     {
         let BitcoinAddressData = None::<BitcoinAddressData>.unwrap();
         let _: String = BitcoinAddressData.address;
@@ -1008,6 +1054,9 @@ const _: fn() = || {
         }
         InputType::Bolt11 { invoice } => {
             let _: LNInvoice = invoice;
+        }
+        InputType::Bolt12Offer { offer } => {
+            let _: LNOffer = offer;
         }
         InputType::NodeId { node_id } => {
             let _: String = node_id;
@@ -1042,6 +1091,16 @@ const _: fn() = || {
         let _: Vec<RouteHint> = LNInvoice.routing_hints;
         let _: Vec<u8> = LNInvoice.payment_secret;
         let _: u64 = LNInvoice.min_final_cltv_expiry_delta;
+    }
+    {
+        let LNOffer = None::<LNOffer>.unwrap();
+        let _: String = LNOffer.bolt12;
+        let _: Vec<String> = LNOffer.chains;
+        let _: Option<Amount> = LNOffer.amount;
+        let _: String = LNOffer.description;
+        let _: Option<u64> = LNOffer.absolute_expiry;
+        let _: Option<String> = LNOffer.issuer;
+        let _: String = LNOffer.signing_pubkey;
     }
     {
         let LnUrlAuthRequestData = None::<LnUrlAuthRequestData>.unwrap();
@@ -1315,6 +1374,31 @@ impl support::IntoDartExceptPrimitive for mirror_AesSuccessActionDataResult {}
 impl rust2dart::IntoIntoDart<mirror_AesSuccessActionDataResult> for AesSuccessActionDataResult {
     fn into_into_dart(self) -> mirror_AesSuccessActionDataResult {
         mirror_AesSuccessActionDataResult(self)
+    }
+}
+
+impl support::IntoDart for mirror_Amount {
+    fn into_dart(self) -> support::DartAbi {
+        match self.0 {
+            Amount::Bitcoin { amount_msat } => {
+                vec![0.into_dart(), amount_msat.into_into_dart().into_dart()]
+            }
+            Amount::Currency {
+                iso4217_code,
+                fractional_amount,
+            } => vec![
+                1.into_dart(),
+                iso4217_code.into_into_dart().into_dart(),
+                fractional_amount.into_into_dart().into_dart(),
+            ],
+        }
+        .into_dart()
+    }
+}
+impl support::IntoDartExceptPrimitive for mirror_Amount {}
+impl rust2dart::IntoIntoDart<mirror_Amount> for Amount {
+    fn into_into_dart(self) -> mirror_Amount {
+        mirror_Amount(self)
     }
 }
 
@@ -1594,17 +1678,20 @@ impl support::IntoDart for mirror_InputType {
             InputType::Bolt11 { invoice } => {
                 vec![1.into_dart(), invoice.into_into_dart().into_dart()]
             }
+            InputType::Bolt12Offer { offer } => {
+                vec![2.into_dart(), offer.into_into_dart().into_dart()]
+            }
             InputType::NodeId { node_id } => {
-                vec![2.into_dart(), node_id.into_into_dart().into_dart()]
+                vec![3.into_dart(), node_id.into_into_dart().into_dart()]
             }
-            InputType::Url { url } => vec![3.into_dart(), url.into_into_dart().into_dart()],
-            InputType::LnUrlPay { data } => vec![4.into_dart(), data.into_into_dart().into_dart()],
+            InputType::Url { url } => vec![4.into_dart(), url.into_into_dart().into_dart()],
+            InputType::LnUrlPay { data } => vec![5.into_dart(), data.into_into_dart().into_dart()],
             InputType::LnUrlWithdraw { data } => {
-                vec![5.into_dart(), data.into_into_dart().into_dart()]
+                vec![6.into_dart(), data.into_into_dart().into_dart()]
             }
-            InputType::LnUrlAuth { data } => vec![6.into_dart(), data.into_into_dart().into_dart()],
+            InputType::LnUrlAuth { data } => vec![7.into_dart(), data.into_into_dart().into_dart()],
             InputType::LnUrlError { data } => {
-                vec![7.into_dart(), data.into_into_dart().into_dart()]
+                vec![8.into_dart(), data.into_into_dart().into_dart()]
             }
         }
         .into_dart()
@@ -1660,6 +1747,27 @@ impl support::IntoDartExceptPrimitive for mirror_LNInvoice {}
 impl rust2dart::IntoIntoDart<mirror_LNInvoice> for LNInvoice {
     fn into_into_dart(self) -> mirror_LNInvoice {
         mirror_LNInvoice(self)
+    }
+}
+
+impl support::IntoDart for mirror_LNOffer {
+    fn into_dart(self) -> support::DartAbi {
+        vec![
+            self.0.bolt12.into_into_dart().into_dart(),
+            self.0.chains.into_into_dart().into_dart(),
+            self.0.amount.map(|v| mirror_Amount(v)).into_dart(),
+            self.0.description.into_into_dart().into_dart(),
+            self.0.absolute_expiry.into_dart(),
+            self.0.issuer.into_dart(),
+            self.0.signing_pubkey.into_into_dart().into_dart(),
+        ]
+        .into_dart()
+    }
+}
+impl support::IntoDartExceptPrimitive for mirror_LNOffer {}
+impl rust2dart::IntoIntoDart<mirror_LNOffer> for LNOffer {
+    fn into_into_dart(self) -> mirror_LNOffer {
+        mirror_LNOffer(self)
     }
 }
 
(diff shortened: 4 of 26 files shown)


libs/sdk-core/src/bridge_generated.rs
```diff
@@ -1071,6 +1071,7 @@ impl support::IntoDart for InputType {
                 vec![0.into_dart(), address.into_into_dart().into_dart()]
             }
             Self::Bolt11 { invoice } => vec![1.into_dart(), invoice.into_into_dart().into_dart()],
+            Self::Bolt12Offer { offer } => todo!(),
```
dangeross, line 1074: This needs to be generated. Please look at `sdk-flutter`and `sdk-react-native` to generate the code for these.

dangeross (changes requested):
Looking good @hydra-yse, thanks. I did a pass over the PR, let me know if you need help running the code generators.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -990,6 +991,48 @@ impl NodeAPI for Greenlight {
...
+
+    #[allow(unused_variables)]
+    async fn fetch_invoice(
+        &self,
+        offer: String,
+        amount_msat: Option<u64>,
+        quantity: Option<u64>,
+        recurrence_counter: Option<u64>,
+        recurrence_start: Option<f64>,
+        recurrence_label: Option<String>,
```
JssDWt, line 1003: Maybe it makes sense to remove the recurrence options until there is a clear plan on how recurrence can fit into the SDK. Recurrence has to be programmed on the client side, so inside the SDK or the calling application somehow. I think it makes sense to explicitly enable that once we know how to handle it. Unless you know how to handle it!

libs/sdk-core/src/models.rs
```diff
@@ -816,6 +817,55 @@ pub struct RefundResponse {
...
+            next_period: response.next_period.map(|np| FetchInvoiceNextPeriod {
+                counter: np.counter,
+                starttime: np.starttime,
+                endtime: np.endtime,
+                paywindow_start: np.paywindow_start,
+                paywindow_end: np.paywindow_end,
+            }),
+        }
+    }
+}
```
JssDWt, line 867: If possible put this mapping in greenlight/node_api.rs,

libs/sdk-core/src/invoice.rs
```diff
@@ -56,6 +58,37 @@ impl From<SystemTimeError> for InvoiceError {
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct LNOffer {
+    pub chains: Vec<ChainHash>,
+    pub amount_msats: Option<u64>,
+    pub description: String,
+    pub absolute_expiry: Option<Duration>,
```
JssDWt, line 83: expiry can be of type `u64`, like in LNInvoice

JssDWt (commented):
Left some small remarks.

I think it could be useful to put the actual sending to bolt12 invoices in the same pull request eventually. So you know exactly which fields you need where. Like blinded paths for example are probably useful for sending, but maybe not so useful to expose to the caller.

dangeross (commented):
I made another pass and left comments, looking good.

Are you going to add the bindings interface to the uniffi UDL in sdk-bindings also? If you make changes there, please rerun the flutter/RN codegen.

dangeross:
I know the CLN method is called fetchinvoice, but what do you think to the interface being called `redeem_offer`? fetch_invoice sounds like we're querying a persisted invoice.

hydra-yse:
> fetch_invoice sounds like we're querying a persisted invoice.

Fair point. I can see the reasoning behind naming it `fetch_invoice`, but I agree that ultimately it's more like "redeeming" or "generating" one. As long as it's documented, I guess there is no issue with diverging from the CLN spec.

libs/sdk-core/src/invoice.rs
```diff
@@ -56,6 +56,52 @@ impl From<SystemTimeError> for InvoiceError {
...
+            lightning::offers::offer::Quantity::One => Quantity::One,
+            lightning::offers::offer::Quantity::Unbounded => Quantity::Unbounded,
+            lightning::offers::offer::Quantity::Bounded(n) => Quantity::Bounded(n.into()),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum Amount { 
+    Bitcoin { amount_msats: u64 },
```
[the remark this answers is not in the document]
hydra-yse: I originally named it `amount_msats` to keep naming persistence with the `lightning::offers::offer::Amount` enum, but there is no problem in changing it :+1:

libs/sdk-core/src/models.rs
```diff
@@ -816,6 +817,59 @@ pub struct RefundResponse {
...
+    pub counter: u64,
+    pub start_time: u64,
+    pub end_time: u64,
+    pub paywindow_start: u64,
+    pub paywindow_end: u64,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct FetchInvoiceResponse {
+    pub invoice: String,
```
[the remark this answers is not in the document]
roeierez: I think in the response it is OK to differentiate and use bolt12, bolt11 as it is more clear to the caller.
About `send_payment` as far as I understand we don't expose the bolt12 to the user, users pay an offer and some optional parameters (we can start with only amount IMO) and we call the fetch_invoice as part of the pay process behind the scene.

libs/sdk-core/Cargo.toml
```diff
@@ -64,6 +61,7 @@ miniz_oxide = "0.7.1"
 tokio-stream = "0.1.14"
 serde_with = "3.3.0"
 regex = "1.8.1"
+gl-client = { git = "https://github.com/hydra-yse/greenlight", branch = "20231109-fetchinvoice" }
```
roeierez, line 64: We can now use the gl main branch as this was merged yesterday.

roeierez (changes requested):
Looking good! Some small comments.

libs/sdk-core/src/node_api.rs
```diff
@@ -46,6 +46,18 @@ pub enum NodeError {
...
+    InvalidOffer(anyhow::Error),
+
+    #[error("Offer expired: {0}")]
+    OfferExpired(anyhow::Error),
+
+    #[error("Offer reply error: {0}")]
+    OfferReplyError(anyhow::Error),
+
+    #[error("Offer timeout: {0}")]
+    OfferTimeout(anyhow::Error),
```
JssDWt, line 60: Suggestion to make more clear what this does.
It's a long name I'm suggesting though, maybe we can do better.
```suggestion
    OfferInvoiceRequestTimeout(anyhow::Error),
```

libs/sdk-core/src/node_api.rs
```diff
@@ -46,6 +46,18 @@ pub enum NodeError {
...
     ServiceConnectivity(anyhow::Error),
+
+    #[error("Invalid offer: {0:?}")]
+    InvalidOffer(anyhow::Error),
+
+    #[error("Offer expired: {0}")]
+    OfferExpired(anyhow::Error),
+
+    #[error("Offer reply error: {0}")]
+    OfferReplyError(anyhow::Error),
```
JssDWt, line 57: Suggestion to make more clear what this does.
It's a long name I'm suggesting though, maybe we can do better.

```suggestion
    OfferInvoiceRequestError(anyhow::Error),
```

libs/sdk-core/src/greenlight/error.rs
```diff
@@ -186,7 +186,11 @@ impl From<tonic::Status> for NodeError {
                 JsonRpcErrCode::InvoiceNoDescription => Self::InvoiceNoDescription(status.into()),
                 JsonRpcErrCode::InvoicePreimageAlreadyExists => {
                     Self::InvoicePreimageAlreadyExists(status.into())
-                }
+                },
+                JsonRpcErrCode::OfferExpired => Self::OfferExpired(status.into()),
+                JsonRpcErrCode::OfferBadInvreqReply => Self::OfferReplyError(status.into()),
+                JsonRpcErrCode::OfferRouteNotFound => Self::RouteNotFound(status.into()),
```
JssDWt, line 192: Thinking out loud. I think it may be useful to distinguish between 'payment' route not found and 'messaging' route not found. RouteNotFound is such a general error that pops up too many times. @dangeross what do you think?

libs/sdk-core/src/invoice.rs
```diff
@@ -56,6 +56,52 @@ impl From<SystemTimeError> for InvoiceError {
...
+    fn from(amount: lightning::offers::offer::Amount) -> Self {
+        match amount {
+           lightning::offers::offer::Amount::Bitcoin { amount_msats } => Amount::Bitcoin { amount_msats },
+           lightning::offers::offer::Amount::Currency { iso4217_code, amount } => Amount::Currency { iso4217_code, amount },
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct LNOffer {
```
JssDWt, line 92: This struct needs a field with the raw offer too. Probably called `bolt12`.

libs/sdk-core/src/models.rs
```diff
@@ -816,6 +817,59 @@ pub struct RefundResponse {
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct FetchInvoiceRequest {
+    pub offer: String,
+    pub amount_msat: Option<u64>,
+    pub quantity: Option<u64>,
+    pub timeout: Option<f64>,
+    pub payer_note: Option<String>,
+    // pub recurrence_counter: Option<u64>,
+    // pub recurrence_start: Option<f64>,
+    // pub recurrence_label: Option<String>,
```
JssDWt, line 829:
```suggestion
```

libs/sdk-core/src/test_utils.rs
```diff
@@ -719,6 +729,25 @@ fn sign_invoice(invoice: RawInvoice) -> String {
...
+        .map(|amount| match amount {
+            crate::invoice::Amount::Bitcoin { amount_msats } => amount_msats,
+            crate::invoice::Amount::Currency { .. } => minimum_sats 
+        })
+        .unwrap_or(10000);
+
+    let invoice = create_invoice(offer.description, amount, vec![], None);
+        
+    FetchInvoiceResponse {
+        invoice: invoice.bolt11, // Should be renamed to simply 'bolt' or 'raw_invoice' in the future
```
JssDWt, line 745: The future is now ;-)

libs/sdk-core/src/input_parser.rs
```diff
@@ -392,6 +412,9 @@ pub enum InputType {
     Bolt11 {
         invoice: LNInvoice,
     },
+    Bolt12Offer {
+        offer: LNOffer,
+    },
```
JssDWt, line 417: This needs changes to the `breez_sdk.udl` file in `libs/sdk-bindings/src`

JssDWt (changes requested):
Looking good!
Some small nits about naming.

There's some changes in Cargo.toml/Cargo.lock files that seem unnecessary for this change, let's clean that up before merging.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -133,7 +133,7 @@ dictionary RouteHint {
 };
 
 dictionary LNInvoice {
-    string bolt11;
+    string raw_invoice;
```
roeierez, line 136: Why is this field changed? do we use it for bolt12 also? I couldn't find any reference for such usage in the code.

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
hydra-yse: We should consider whether the `LNInvoice` wrapper should be able to support both Bolt11 and Bolt12 fields at once, given the overlap between the two. Currently CLN makes the invoice field naming distinct between the two types in most methods (e.g. createinvoice, listinvoices etc.) with the exception of outward methods like sendpay and pay. 

The initial reasoning was that, in terms of UX, I believe that the final user is probably not really interested in the type of invoice they are paying (whether bolt11 or 12), so perhaps unifying the optional `bolt11` and `bolt12` parameters into a single field `raw_invoice` could make sense and reduce the need to constantly check which one it is. Additionally, the client would know the type a priori to the fetchinvoice call simply by the scanned QR (offer or bolt11 invoice), therefore there is most probably a way to make this work with a single field.

On the other hand, I can also see instances where the developers would like to know which type of invoice it is at any given time. That is of course more flexible on the developer's side as they are now able to handle the two distinct cases better (such as adding more fields depending on which kind it is [e.g. quantity]) and throw an error, for example, if neither is specified, without having to refer to the original offer to discern so. 

I'll probably roll back this change and specify the two fields as optional, just wanted to let you in on what the thought process was behind that decision

roeierez: Thanks for the details on what you had in mind @hydra-yse 
For me it is even more simple than that. As long as the code doesn't user LNINvoice for bolt12 (and we don't even expose this parsing to the user) then there is no reason to unified.
If/When we implement  and expose such parsing we will have more information to take the right decision.

JssDWt: Agree with Roei. For parsing we can differentiate between bolt11 and bolt12 invoices. For send_payment we can call the `bolt11` argument `payment_request`. The developer would call `send_payment` with either the bolt11 fields from LNInvoice or the `bolt12_offer` field from LNOffer.

libs/sdk-core/src/lnurl/withdraw.rs
```diff
@@ -56,11 +56,15 @@ fn build_withdraw_callback_url(
     req_data: &LnUrlWithdrawRequestData,
     invoice: &LNInvoice,
 ) -> LnUrlResult<String> {
+    let raw_invoice = invoice.bolt11.as_ref()
+        .or(invoice.bolt12.as_ref())
+        .expect("Expected either a bolt11 or bolt12 invoice");
```
JssDWt, line 61: Does lnurl support bolt12?

libs/sdk-core/src/models.rs
```diff
@@ -816,6 +817,56 @@ pub struct RefundResponse {
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct FetchInvoiceRequest {
+    pub offer: String,
+    pub amount_msat: Option<u64>,
+    pub quantity: Option<u64>,
+    pub timeout: Option<f64>,
+    pub payer_note: Option<String>,
+}
+
+impl From<FetchInvoiceRequest> for gl_client::pb::cln::FetchinvoiceRequest {
```
JssDWt, line 829: Please put this mapping inside node_api

JssDWt (commented):
The main questions for me are:
- what to do with the `parse_invoice` interface?
- can we call the `fetch_invoice` _inside_ the `send_payment` method, rather than exposing it to the client?

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1368,16 @@ impl BreezServices {
             },
         })
     }
+
+    pub async fn fetch_invoice(&self, req: FetchInvoiceRequest) -> Result<FetchInvoiceResponse> {
```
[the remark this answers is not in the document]
hydra-yse: 1. User scans the QR code, reading the offer and decoding its contents
2. User gets prompted with mandatory fields to fill (e.g. `quantity`,  `amount_msat` [if offer's `amount` = any] etc.)
3. Client calls `fetch_invoice` with specified parameters, gets a bolt12 invoice in return and displays relevant changes (e.g. `amount_msat`, which _should_ correspond to the specified `amount_msat` field times the `quantity` field, but may have been changed by the recipient for any reason)
4. Client displays the invoice details, if user agrees they execute `send_payment`, and the invoice gets paid

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -990,6 +991,38 @@ impl NodeAPI for Greenlight {
...
+                vendor_removed: changes.vendor_removed,
+                amount_msat: changes.amount_msat.map(|amount| amount.msat),
+            }),
+            next_period: response.next_period.map(|np| FetchInvoiceNextPeriod {
+                counter: np.counter,
+                start_time: np.starttime,
+                end_time: np.endtime,
+                paywindow_start: np.paywindow_start,
+                paywindow_end: np.paywindow_end,
+            }),
```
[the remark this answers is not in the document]
hydra-yse: Next period is indeed related to the offer's recurrence, and since we have omitted it, it makes sense to do the same here. Thanks  :+1:

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -133,7 +133,8 @@ dictionary RouteHint {
 };
 
 dictionary LNInvoice {
-    string bolt11;
+    string? bolt11;
+    string? bolt12;
```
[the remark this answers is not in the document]
hydra-yse: I was also conflicted about the decision. The merging into `LNInvoice` made sense to me given that the invoices are functionally equivalent at the moment, given that we do not handle recurrence and blinded paths yet, plus the fact that CLN's [create_invoice](https://docs.corelightning.org/reference/lightning-createinvoice) command (which is not directly called, as it is more low-level) does exactly the same.

Despite that, some higher level commands like `lightning-invoice` do not allow us to create bolt12 invoices directly, therefore it would make sense to distinguish between those which can and can't with two different interfaces, then using enums to specify either one of them. 

Lastly, regarding the codebase, I agree that having to propagate the changes everywhere is not ideal, yet I've checked that altering `parse_invoice` would require us to make many changes (if not more) as well. At the end of the day though, I think it is worth going forward as you suggested.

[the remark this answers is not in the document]
roeierez: I agree and strongly feel that parsing bolt11 should return a different type than LNInvoice simply because it has different data (even though currently we don't expose/use it).
As for exposing parse_invoice in the udl I agree it is not really necessary (perhaps there are times when the user "knows" that she have a bolt11 in hand and can use this function directly rather than going through the input parser). Doesn't mean we have to do the same mistake twice if we are not sure.

How about:
1. Leave the parse_invoice as is for now (bolt11 only).
2. Not expose parsing bolt12 directly but only via the input parser
2. Use different structure for bolt12 invoices.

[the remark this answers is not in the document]
JssDWt: I think 
- `LNInvoice` could be renamed to `Bolt11Invoice` to take away any ambiguity.
- `parse_invoice` could be renamed to `parse_bolt11_invoice` to make it clear what it's for. Maybe add a `parse_bolt12_invoice` function (internally) if necessary.
- Add a `Bolt12Invoice` type
- The input parser could probably use another branch to parse bolt12 invoices as well.

libs/sdk-core/src/breez_services.rs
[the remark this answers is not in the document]
JssDWt: It's a shame we have to expose all this all the way to the user. It makes the sdk harder to use. It's not just a send_payment flow anymore, but you really have to implement many branches as a developer. I guess there's not much we can do about that though.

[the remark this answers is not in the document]
roeierez: @JssDWt @hydra-yse my suggestion above consolidates the flow I agree we shouldn't make the flow complicated and we should do our best do send_payment would fit both cases (bolt12 and bolt11).
If the internally fetch 8nvoice returns mismatch in amount let's just return error

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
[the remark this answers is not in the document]
roeierez: Just to make sure, we came to a conclusion that we don't need this method at the `BreezServices` level right?

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -990,6 +991,110 @@ impl NodeAPI for Greenlight {
...
+        // Fetch the invoice from the recipient's node
+        let mut client = self.get_node_client().await?;
+        let response = client
+            .fetch_invoice(Into::<cln::FetchinvoiceRequest>::into(req))
+            .await?
+            .into_inner();
+
+        Ok(FetchInvoiceResponse {
+            bolt12: response.invoice,
+            changes: response.changes.map(|changes| FetchInvoiceChanges {
```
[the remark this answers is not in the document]
hydra-yse: For the end user, this is a negative scenario. If the amount changes from what the user expects without them knowing, they would be paying a higher amount than expected without confirmation. 

I'll remove throwing an error when the `description` or `vendor` change, as they are not that relevant (added them for completeness), but definitely will keep the change in the expected amount.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -990,6 +991,110 @@ impl NodeAPI for Greenlight {
...
+            Ok(offer) => offer,
+            Err(_) => {
+                return Err(NodeError::InvalidOffer(anyhow!("Invalid offer")));
+            }
+        };
+
+        let mut field_errors: Vec<FetchInvoiceRequiredField> = vec![];
+
+        if parsed_offer.expects_quantity() && req.quantity.is_none() {
+            field_errors.push(FetchInvoiceRequiredField::Quantity);
```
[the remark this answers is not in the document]
hydra-yse: Indeed, I kept them for the developer side if a change is ever detected. This way, they can check the array of fields and get a convenient enum which they can then use to throw specific errors, or perhaps show a "red field" for the values that need to be changed in order for the offer to succeed. We could throw a more generic error directly, but then the user would not know which values are making the request fail exactly

libs/sdk-core/src/models.rs
```diff
@@ -816,6 +816,63 @@ pub struct RefundResponse {
...
+pub struct FetchInvoiceChanges {
+    pub description_appended: Option<String>,
+    pub description: Option<String>,
+    pub vendor_removed: Option<String>,
+    pub vendor: Option<String>,
+    pub amount_msat: Option<u64>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct FetchInvoiceResponse {
```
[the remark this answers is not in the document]
hydra-yse: The structs must be prefixed with `pub` in order to be accessible from other sub-modules, like the `node_api`

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -403,6 +404,24 @@ dictionary StaticBackupResponse {
     sequence<string>? backup;
 };
 
+dictionary CreateOfferRequest {
+    u64? amount_msat;
+    string description;
+    u64? absolute_expiry;
+    string? issuer;
+    u64? supported_quantity;
+    string? label;
```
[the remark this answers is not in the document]
hydra-yse: Fixed in the following commits :+1:

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -990,6 +991,110 @@ impl NodeAPI for Greenlight {
...
+        }
+
+        if parsed_offer.amount().is_some() && req.amount_msat.is_none() {
+            field_errors.push(FetchInvoiceRequiredField::Amount);
```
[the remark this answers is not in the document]
hydra-yse: Fixed in the following commits :+1:

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -403,6 +404,24 @@ dictionary StaticBackupResponse {
...
+    boolean? single_use;
```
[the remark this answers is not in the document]
hydra-yse: Agreed. For now if a user needs to receive a one-time payment, they can do so with bolt11.

libs/sdk-core/src/test_utils.rs
```diff
@@ -719,6 +737,23 @@ fn sign_invoice(invoice: RawInvoice) -> String {
...
 }
 
+fn fetch_invoice(offer: LNOffer) -> FetchInvoiceResponse {
+    // Mock node receiving FetchInvoiceRequest and building an invoice out of it
+    let currency_amount_msat = 10000u64;
+    let amount = offer
+        .amount
+        .map(|amount| match amount {
+            crate::invoice::Amount::Bitcoin { amount_msat } => amount_msat,
+            crate::invoice::Amount::Currency { .. } => currency_amount_msat,
```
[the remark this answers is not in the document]
hydra-yse: You're right, this was part of the old tests, planning to change it and find a better way to replicate the invoice creation step from the offer using the [lightning](https://docs.rs/lightning/0.0.115/lightning/) library.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -403,6 +404,24 @@ dictionary StaticBackupResponse {
...
+};
+
+dictionary PayOfferRequest {
+    string offer;
+    u64? amount_msat;
+    u64? quantity;
```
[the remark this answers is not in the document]
hydra-yse: After thinking about it, I agree. Implementing quantity doesn't take much (just a few lines of code in fact, since the changes are handled on CLN) though I do agree in more general terms that perhaps it could be handle elsewhere. Leaving it out for now :+1:

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -990,6 +991,58 @@ impl NodeAPI for Greenlight {
...
+        let mut client = self.get_node_client().await?;
+        let response = client
+            .fetch_invoice(Into::<cln::FetchinvoiceRequest>::into(req))
+            .await?
+            .into_inner();
+
+        Ok(FetchInvoiceResponse {
+            bolt12: response.invoice,
+            new_amount_msat: match response.changes {
+                Some(changes) => changes.amount_msat.map(|amount| amount.msat),
```
roeierez, line 1011: Do we need to check if other changes exists and return error? My concern is that it seems we swallow these changes without any indication for the user.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1368,38 @@ impl BreezServices {
             },
         })
     }
+
+    pub async fn create_offer(&self, req: CreateOfferRequest) -> Result<String> {
```
roeierez, line 1372: This API is actually similar to `receive_payment` (that returns bolt11) but for bolt12.
The way it is right now is a bit inconsistent since the `receive_payment` returns a response with the invoice parsed (LNInvoice) and here we return just the raw offer. For consistency perhaps we should return the LNOffer structure here.

libs/sdk-core/src/models.rs
```diff
@@ -816,6 +816,36 @@ pub struct RefundResponse {
...
+    pub amount_msat: Option<u64>,
+    pub timeout: Option<f64>,
+    pub payer_note: Option<String>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct FetchInvoiceResponse {
+    pub bolt12: String,
+    pub new_amount_msat: Option<u64>,
+}
```
roeierez, line 831: These are not publicly exposed right but only needed internally in the sdk right? If so let's change their access modifier accordingly.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -403,6 +404,20 @@ dictionary StaticBackupResponse {
     sequence<string>? backup;
 };
 
+dictionary CreateOfferRequest {
+    u64? amount_msat;
+    string description;
+    u64? absolute_expiry;
+    string? issuer;
```
roeierez, line 411: I would remove the issuer for the first version of bolt12 feature.

roeierez (changes requested):
@hydra-yse this shapes up very nicely and looks really good. I added some comments, most of them are Nits.
I think we are missing two things:
1. Supporting open channels as part of bolt12 received payment.
2. Adding cli commands to create offer and pay offer.

libs/Cargo.toml
```diff
@@ -14,7 +14,3 @@ resolver = "2"
 
 [workspace.package]
 version = "0.2.7"
-
-[workspace.dependencies]
-uniffi = "0.23.0"
-uniffi_macros = "0.23.0"
```
JssDWt, line 20: I assume these are for testing. Make sure this change is reverted before merging.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1368,38 @@ impl BreezServices {
...
+                offer: req.offer.clone(),
+                amount_msat: req.amount_msat,
+                timeout: req.timeout,
+                payer_note: req.payer_note.to_owned()
+            })
+            .await?;
+
+        if let Some(new_amount) = fetch_invoice_response.new_amount_msat {
+            return Err(SendPaymentError::OfferAmountChanged { new_amount });
+        }
```
JssDWt, line 1396: I wonder whether it could be useful to fail on all changes here. Given we don't support changes at all.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1368,6 +1368,38 @@ impl BreezServices {
...
+
+        Ok(self.send_payment(SendPaymentRequest {
+            invoice: fetch_invoice_response.bolt12,
+            amount_msat: req.amount_msat
```
JssDWt, line 1400: Does explicitly setting the amount work here? I think the invoice may have the amount filled already and the call to send_payment may fail?

libs/sdk-core/src/breez_services.rs
```diff
@@ -237,7 +237,7 @@ impl BreezServices {
         req: SendPaymentRequest,
     ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
-        let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
+        let parsed_invoice = parse_invoice(req.invoice.as_str())?;
```
JssDWt, line 240: I believe this still parses a bolt11 invoice?

JssDWt (commented):
It's starting to look very clean, I like it!

libs/sdk-core/src/greenlight/node_api.rs
hydra-yse: For basic support, currently not. Changes in description/vendor can be ignored while a change in the amount could mean the user paying more than the expects, thus an error has to be thrown

libs/sdk-core/src/breez_services.rs
hydra-yse: `pay_offer` internally calls `send_payment`, so I thought it would be appropriate to change the field to `invoice` as we now know it can support both bolt11 and 12.

libs/sdk-core/src/breez_services.rs
hydra-yse: Technically this is a redundant check of the changes. If the target CLN node is working properly, we expect it to return an invoice with the same amount_msat as we specified, thus the call to `send_payment` should not fail. It is not necessary to pass it to the method, so might remove it altogether.

JssDWt: For bolt11 invoices this fails I think, because cln will reject the call to `pay`. Only for zero amount invoices are you allowed to set the amount.

libs/sdk-core/src/breez_services.rs
JssDWt: I mean the `parse_invoice` function expects a bolt11 invoice. So if a bolt12 invoice is passed here, the parsing will fail.

hydra-yse: Right. I think the best solution would be to make `parse_invoice` compatible with either, returning either a bolt11 or a bolt12 invoice. Alternatively, I can just replicate the `send_payment` flow on `pay_offer` and create a separate parsing method for bolt 12 invoices. Let me know what you think

libs/sdk-core/src/breez_services.rs
hydra-yse: Alright, I'll revert the changes to how they were previously and display all of them with an appropriate error

libs/sdk-core/src/breez_services.rs
hydra-yse: Sorry, what I meant was just strictly in regards to the bolt12 flow. I fixed the issue by not passing `amount_msat` at all, as we expect the fetched invoice to have the same amount as our `FetchInvoiceRequest` anyway.

dangeross:
Adding CLN Bolt12 compat tracking issue here
https://github.com/ElementsProject/lightning/issues/7407

PR #600, opened 2023-11-09
ci: fix workflow_call inputs

JssDWt:
https://github.com/breez/breez-sdk/pull/569 Inadvertently broke the workflow_call CI: https://github.com/breez/breez-sdk-docs/actions/runs/6816197496/job/18536990174

This PR re-adds support for workflow_call. The workflow_call parameters are not bounded by input limits, but the json parsing doesn't work on those inputs.


cnixbtc (approved):
Looks good, thanks for fixing! 

Why the double parentheses though? Are they on purpose?

JssDWt:
> Why the double parentheses though? Are they on purpose?

No, they're not on purpose!

PR #601, opened 2023-11-09
Expose the Greenlight node credentials

### LINKED ISSUE #556: Expose Greenlight credentials in the SDK
As a developer, I would like to register/use a Greenlight node with the Breez SDK and get the Greenlight node credentials, so that I can use the gl-client alongside it.

dangeross:
Fixes #556

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -81,72 +81,39 @@ impl Greenlight {
         .to_bytes();
         let encryption_key_slice = encryption_key.as_slice();
 
-        let legacy_encryption_key = Self::legacy_derive_bip32_key(
-            config.network,
-            &signer,
-            vec![ChildNumber::from_hardened_idx(140)?, ChildNumber::from(0)],
-        )?
-        .to_priv()
-        .to_bytes();
-        let legacy_encryption_key_slice = legacy_encryption_key.as_slice();
-
         let register_credentials = match config.node_config.clone() {
             NodeConfig::Greenlight { config } => config,
         };
 
-        // query for the existing credentials
-        let credentials = persister.get_gl_credentials()?;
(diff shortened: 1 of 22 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -563,6 +563,31 @@ impl Greenlight {
...
+        match self.persister.get_gl_credentials()? {
+            Some(creds) => match sym_decrypt(encryption_key_slice, creds.as_slice()) {
+                Some(creds) => {
+                    let credentials: GreenlightCredentials =
+                        serde_json::from_slice(creds.as_slice()).map_err(|e| {
+                            NodeError::Generic(anyhow!("Unable to parse credentials: {e}"))
+                        })?;
+                    Ok(Some(NodeCredentials::Greenlight { credentials }))
+                }
+                None => Ok(None),
```
ok300, line 585: Failing to decrypt should throw an error, see https://github.com/breez/breez-sdk/blob/6b74de2ae9c939789fb7eb5301814cd961e2ac6b/libs/sdk-core/src/greenlight/node_api.rs#L117C56-L117C56

ok300 (commented):
Maybe a call to the new `node_credentials()` can be used to replace part of `connect()`?

For example this section: https://github.com/breez/breez-sdk/blob/6b74de2ae9c939789fb7eb5301814cd961e2ac6b/libs/sdk-core/src/greenlight/node_api.rs#L98-L121

I see `connect()` also has a fallback to a legacy decrypt call. It might make sense to integrate that too in the new call, and replace the `connect()` portion IMO.

dangeross:
> I see `connect()` also has a fallback to a legacy decrypt call. It might make sense to integrate that too in the new call, and replace the `connect()` portion IMO.

I can look into merging them. My thought was after `connect()` the legacy encryption key would no longer be needed because they are re-encrypted using the correct key.

ok300 (approved):

PR #602, opened 2023-11-09
CI: match dispatch inputs in workflow_call

JssDWt:
Unfortunately the [previous fix](https://github.com/breez/breez-sdk/pull/600) didn't work. The json was evaluated anyway. So match the workflow_dispatch input structure in workflow_call.

### DIFF
--- a/.github/workflows/publish-all-platforms.yml
+++ b/.github/workflows/publish-all-platforms.yml
@@ -39,36 +39,17 @@ on:
         description: 'commit/tag/branch reference'
         required: true
         type: string
-      csharp-package-version:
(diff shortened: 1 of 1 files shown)


cnixbtc (approved):
Out of curiosity and sorry if I'm missing something obvious here: How/when do we trigger the `workflow_call` as opposed to the manual trigger?

JssDWt:
> Out of curiosity and sorry if I'm missing something obvious here: How/when do we trigger the `workflow_call` as opposed to the manual trigger?

There's a workflow in breez-sdk-docs here: https://github.com/breez/breez-sdk-docs/blob/main/.github/workflows/main.yml
It's used to create the packages so we can check whether the syntax of the snippets in the docs is ok.

PR #603, opened 2023-11-10
use req for sweepRequest

ubbabeck:
This pr adds the following changes.

- [039247c](https://github.com/breez/breez-sdk/pull/603/commits/039247c4db3ad239a14dd67cb1e70d609c58033d) Which requires `PrepareSweepRequest` as parameter.
- [ba6493b](https://github.com/breez/breez-sdk/pull/603/commits/ba6493bd6b772df156ec13ab932a1cbbbf4f31c9) Changes name of param to `sats_per_vbyte` for consistency for both `Sweep` and `Refund` request and prepare . 

Not sure if `sats_per_vbyte` is the preferred name, but I think we should use the same name for consistency.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -619,7 +619,7 @@ enum BuyBitcoinProvider {
 
 dictionary PrepareSweepRequest {
     string to_address;
-    u64 sats_per_vbyte;
+    u64 sat_per_vbyte;
 };
 
 dictionary PrepareSweepResponse {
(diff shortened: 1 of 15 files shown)


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -678,7 +678,7 @@ dictionary PrepareRefundResponse {
 dictionary RefundRequest {
     string swap_address;
     string to_address;
-    u32 sat_per_vbyte;
+    u32 sats_per_vbyte;
```
JssDWt, line 681: `sat_per_vbyte` is probably fine. Not worth a breaking change on the `RefundRequest` and the `PrepareRefundRequest` imo.

JssDWt (commented):
One comment about breaking change. Otherwise looks good.

libs/sdk-bindings/src/breez_sdk.udl
ubbabeck: Good point. Fixed

JssDWt (approved):
LGTM!

Issue #606, opened 2023-11-12
Python Wheels for Python 3.11 and 3.12 are missing


prusnak:
See https://pypi.org/project/breez-sdk-lib/#files

There are only wheels for Python 3.8, 3.9 and 3.10.

Please make wheels for Python 3.11 and 3.12 available too. Always make sure you submit wheels for all supported Python versions.

prusnak:
ISTM all you need to change is this line:

https://github.com/breez/breez-sdk/blob/main/.github/workflows/publish-python.yml#L32

to read: `python: ["3.8", "3.9", "3.10", "3.11", "3.12"]` and re-run the build.

And then also changing the README.md here: https://github.com/breez/breez-sdk/blob/f5fac7e7713002df2b60532ba740776ec475ce3c/libs/sdk-bindings/bindings-python/README.md#L26

cnixbtc:
Hey @prusnak, thanks for raising this. I'll get back to you on this tomorrow.

cnixbtc:
@prusnak I've added 3.11. and 3.12 support to our release pipeline. Once merged, we'll create a new release on PyPi (likely 0.2.9.post1) which should then include wheels for those Python versions. I'll keep you posted here.

cnixbtc:
Hey @prusnak we just published 0.2.9.post3 which includes wheels for Python 3.11 and 3.12. Please let me know if that solves your issue. You can install this version using:

``` python
pip install breez-sdk==0.2.9.post3
```

prusnak:
Thanks. It works!

PR #608, opened 2023-11-13
Build python wheels for 3.11 and 3.12

cnixbtc:
Treats #606.

Configures the CI to also build wheels for Python 3.11 and 3.12.

After merging, I'd say we re-run the CI for `0.2.9` and use the `0.2.9.post1` [post-release tag](https://peps.python.org/pep-0440/#post-releases) for the Python package. That way we don't have to create a whole new SDK release just for uploading the new wheels while also unblocking #606.

### DIFF
--- a/.github/workflows/publish-python.yml
+++ b/.github/workflows/publish-python.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
       - name: Checkout breez-sdk repo
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


JssDWt (approved):
LGTM!

> After merging, I'd say we re-run the CI for 0.2.9 and use the 0.2.9.post1 [post-release tag](https://peps.python.org/pep-0440/#post-releases) for the Python package. 

Note that the change in the readme won't be included in that 0.2.9.post1 release then. Because the CI will checkout the 0.2.9 tag, which didn't have the readme change yet.

cnixbtc:
> Note that the change in the readme won't be included in that 0.2.9.post1 release then. Because the CI will checkout the 0.2.9 tag, which didn't have the readme change yet.

Good point. Not an ideal situation but I think it's ok since the readme won't show up on PyPi. It's only visible when you browse the repo.

cnixbtc:
Thanks for the fast review @JssDWt. 🙌

PR #612, opened 2023-11-15
Trigger GL stop on SDK disconnect

ok300:


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1210,6 +1210,15 @@ impl NodeAPI for Greenlight {
                     .into_inner();
                 Ok(format!("{resp:?}"))
             }
+            NodeCommand::Stop => {
```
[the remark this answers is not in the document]
ok300: Done, removed the dev command

roeierez (approved):

ok300:
I re-added the dev command, as discussed.

libs/sdk-core/src/breez_services.rs
```diff
@@ -227,6 +227,7 @@ impl BreezServices {
                 err: format!("Shutdown failed: {e}"),
             })?;
         *started = false;
+        self.node_api.stop().await;
```
JssDWt, line 230: @ok300 This will lead to unexpected behavior if you're running the sdk on multiple clients.

ok300: Calling `stop` even on a running client looks to be fine. The only side-effect is the next operation that uses the signer takes a bit longer, because the remote GL scheduler has to be started again.

To be safe, I agree we could remove this, e.g. not call it on `disconnect`. In that case, we can keep `stop` as a dev command and remove it from the `NodeAPI` trait, as its not meant to be called in any service call. @roeierez would you agree?

PR #620, opened 2023-11-16
add ci for swift

JssDWt:
Move the Swift CI into the sdk repo.
Needs a secret to be set: `COCOAPODS_TRUNK_TOKEN`


roeierez (approved):
LGTM! One thing to note is that RN depends on the swift build so right now we will need to publish first the swift and make sure it passes and then run all other platforms unless there is a way to only run RN if swift passes...

JssDWt:
> LGTM! One thing to note is that RN depends on the swift build so right now we will need to publish first the swift and make sure it passes and then run all other platforms unless there is a way to only run RN if swift passes...

Good point. I'll see if I can make React native run only if swift completes, regardless of whether swift was being built or not.

JssDWt:
Closing in favor of #1033

Issue #570, opened 2023-10-30
Bug: in-progress outgoing payment isn't included in payment list


danielgranhao:
### Current behaviour

If `list_payments()` is called while `send_payment()` is running, the payment list doesn't include the new outgoing payment.

### Expected behaviour

If `list_payments()` is called while `send_payment()` is running, the payment list includes the new payment with status `PaymentStatus::Pending`

danielgranhao:
Taking this up if no one has yet

roeierez:
Its yours!

danielgranhao:
I realized that calling `sync()` between the start of the payment and getting the payments list effectively circumvents the issue. I assume that syncing within the implementation of `list_payments` isn't a desirable fix as it makes the method rely on network requests, which up until now it doesn't. Am I correct in this assumption?

roeierez:
Yes, correct.

roeierez:
Can we close this @danielgranhao @dleutenegger  ?

ok300:
Closing as per confirmation above.

PR #621, opened 2023-11-16
Persist payment attempts before starting

danielgranhao:
Partially addresses #570.

The proposed solution isn't applicable to keysend payments as there isn't a payment hash available before requesting the payment to the specified node id.

I am looking for feedback and ideas for addressing the keysend part. Addressing the keysend part in a later PR is also an option.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1,3 +1,4 @@
+use std::cmp::max;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


roeierez:
@danielgranhao how do you intend to use this change? Asking because the send_payment is blocking and the pending payment is populated as part of it. Are you polling in a different thread?

danielgranhao:
> Are you polling in a different thread?

@roeierez Yes, exactly. In the mobile app we allow users to put the payment process "in the background", going back to the payment list. This allows the new in-progress payment to be shown in the payments list.

libs/sdk-core/src/breez_services.rs
```diff
@@ -264,12 +265,15 @@ impl BreezServices {
...
 
+        let amount_msat = max(provided_amount_msat, invoice_amount_msat);
+
         match self
             .persister
             .get_completed_payment_by_hash(&parsed_invoice.payment_hash)?
         {
             Some(_) => Err(SendPaymentError::AlreadyPaid),
             None => {
+                self.persist_new_bolt11_payment_attempt(&parsed_invoice, amount_msat)?;
```
roeierez, line 276: NIT: `persisten_pending_payment` ?

roeierez (approved):
LGTM! Added NIT suggestion

libs/sdk-core/src/breez_services.rs
danielgranhao: Method renamed to `persist_pending_payment` in [d01b703](https://github.com/breez/breez-sdk/pull/621/commits/d01b703f7b0ca40762b57604fa40c35fffd007a2). I assumed the `en` was a typo :) please let me know if it's not.

dangeross (approved):

PR #623, opened 2023-11-16
Persist pending payment amount

### LINKED ISSUE #599: Bug: pending outgoing payment of open bolt11 invoice returns amount inferior to the one attempted to be sent
We've already experienced more than once that when paying an open (aka zero-amount) bolt11 invoice, if it stays pending, the `amount_msat` returned in `Payment` doesn't correspond to the amount that was attempted to be paid. 

In one instance, the payment was for 50k sats, and the `Payment` specified ~48323 sats, and in another, it was for 45k sats, and we got ~3631 sats. Looking at the logs, for the latter, I could see that it was an MPP and one of the parts was of 3631 sats, suggesting this is related.

Not sure if this could also happen for non-open invoices, but we haven't experienced it.

dleutenegger:
Resolves #599 

Builds upon #621

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -283,7 +283,6 @@ impl BreezServices {
                     .on_payment_completed(
                         parsed_invoice.payee_pubkey.clone(),
                         Some(parsed_invoice),
-                        req.amount_msat,
                         payment_res,
                     )
                     .await?;
@@ -304,7 +303,7 @@ impl BreezServices {
             .map_err(Into::into)
             .await;
         let payment = self
-            .on_payment_completed(req.node_id, None, Some(req.amount_msat), payment_res)
+            .on_payment_completed(req.node_id, None, payment_res)
             .await?;
         Ok(SendPaymentResponse { payment })
     }
@@ -334,6 +333,7 @@ impl BreezServices {
                     bolt11: cb.pr.clone(),
                     amount_msat: None,
                 };
+                let invoice = parse_invoice(cb.pr.as_str())?;
 
                 let payment = match self.send_payment(pay_req).await {
                     Ok(p) => Ok(p),
@@ -341,8 +341,6 @@ impl BreezServices {
                         SendPaymentError::InvalidInvoice { .. } => Err(e),
                         SendPaymentError::ServiceConnectivity { .. } => Err(e),
                         _ => {
-                            let invoice = parse_invoice(cb.pr.as_str())?;
-
                             return Ok(LnUrlPayResult::PayError {
                                 data: LnUrlPayErrorData {
                                     payment_hash: invoice.payment_hash,
@@ -394,7 +392,7 @@ impl BreezServices {
                     Some(req.data.metadata_str),
                     req.data.ln_address,
                     None,
-                    None,
+                    invoice.amount_msat,
                 )?;
 
                 Ok(LnUrlPayResult::EndpointSuccess {
@@ -938,6 +936,17 @@ impl BreezServices {
                 },
             },
         }])?;
+
+        if invoice.amount_msat.is_none() {
+            self.persister.insert_payment_external_info(
+                &invoice.payment_hash,
+                None,
+                None,
+                None,
+                None,
+                Some(amount_msat),
+            )?;
+        }
         Ok(())
     }
 
@@ -945,7 +954,6 @@ impl BreezServices {
         &self,
         node_id: String,
         invoice: Option<LNInvoice>,
-        amount_msat: Option<u64>,
         payment_res: Result<PaymentResponse, SendPaymentError>,
     ) -> Result<Payment, SendPaymentError> {
         self.do_sync(payment_res.is_ok()).await?;
@@ -962,16 +970,6 @@ impl BreezServices {
                 }),
             },
             Err(e) => {
-                if let Some(inv) = invoice.clone() {
-                    self.persister.insert_payment_external_info(
-                        &inv.payment_hash,
-                        None,
-                        None,
-                        None,
-                        None,
-                        amount_msat.or(inv.amount_msat),
-                    )?;
-                }
                 self.notify_event_listeners(BreezEvent::PaymentFailed {
                     details: PaymentFailedData {
                         error: e.to_string(),
--- a/libs/sdk-core/src/persist/migrations.rs
+++ b/libs/sdk-core/src/persist/migrations.rs
@@ -531,5 +531,6 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
         END;
         ",
         "ALTER TABLE payments_external_info ADD COLUMN failed_amount_msat INTEGER;",
+        "ALTER TABLE payments_external_info RENAME COLUMN failed_amount_msat TO attempted_amount_msat;",
     ]
 }
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -1613,10 +1613,10 @@ impl TryFrom<cln::ListpaysPays> for Payment {
             payment_type: PaymentType::Sent,
             payment_time: payment.completed_at.unwrap_or(payment.created_at) as i64,
             amount_msat: match status {
-                PaymentStatus::Failed => ln_invoice
+                PaymentStatus::Complete => payment_amount,
+                _ => ln_invoice
                     .as_ref()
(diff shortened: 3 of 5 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -905,6 +909,38 @@ impl BreezServices {
...
+    fn persist_new_bolt11_payment_attempt(
+        &self,
+        invoice: &LNInvoice,
+        amount_msat: u64,
+    ) -> Result<(), SendPaymentError> {
+        self.persister.insert_or_update_payments(&[Payment {
+            id: invoice.payment_hash.clone(),
+            payment_type: PaymentType::Sent,
+            payment_time: SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs() as i64,
+            amount_msat,
```
dangeross, line 921: This can be set to 0 and instead call `persister.insert_payment_external_info()` to store the attempted_amount_msat in the payment_external_info sync table.
```suggestion
            amount_msat: 0,
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -905,6 +909,38 @@ impl BreezServices {
...
+                    bolt11: invoice.bolt11.clone(),
+                    lnurl_success_action: None,
+                    ln_address: None,
+                    lnurl_metadata: None,
+                    lnurl_withdraw_endpoint: None,
+                    swap_info: None,
+                },
+            },
+        }])?;
+        Ok(())
```
dangeross, line 941:
```suggestion
        self.persister.insert_payment_external_info(
            &invoice.payment_hash,
            None,
            None,
            None,
            None,
            amount_msat,
        )?;
        Ok(())
```

dangeross (changes requested):
Looking good, just need to change how the attempted_amount_msat is persisted

libs/sdk-core/src/breez_services.rs
roeierez: > This can be set to 0 and instead call `persister.insert_payment_external_info()` to store the attempted_amount_msat in the payment_external_info sync table.

@dangeross why do we need to sync the payment_external_info with pending payment?

dangeross: Maybe I misunderstood the PR, but it looks like the intention is to store the pending amount in `attempted_amount_msat`. @dleutenegger ?

roeierez: @dangeross either this or the migration was added by mistake because I don't see any update to the sync table.
Let's wait for @dleutenegger answer.

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -531,5 +531,6 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
         END;
         ",
         "ALTER TABLE payments_external_info ADD COLUMN failed_amount_msat INTEGER;",
+        "ALTER TABLE payments_external_info RENAME COLUMN failed_amount_msat TO attempted_amount_msat;",
```
dangeross, line 534: Here @roeierez

roeierez: Yes I saw that but I assumed it is a mistake since I don't see any inserts/updates to this table as part of this PR.

libs/sdk-core/src/breez_services.rs
dleutenegger: @roeierez yes the intention is to persist the `attempted_amount_msat` as @dangeross pointed out [here](https://github.com/breez/breez-sdk/pull/623#discussion_r1396354458), the actual call to persist just got missed in the `persist_new_bolt11_payment_attempt`-method

libs/sdk-core/src/breez_services.rs
dleutenegger: Addressed in [49e5938](https://github.com/breez/breez-sdk/pull/623/commits/49e5938166a2793814d74b33000e387e93a43e9a)

libs/sdk-core/src/breez_services.rs
dleutenegger: Addressed in [49e5938](https://github.com/breez/breez-sdk/pull/623/commits/49e5938166a2793814d74b33000e387e93a43e9a)

libs/sdk-core/src/breez_services.rs
```diff
@@ -905,11 +907,51 @@ impl BreezServices {
...
+                    lnurl_success_action: None,
+                    ln_address: None,
+                    lnurl_metadata: None,
+                    lnurl_withdraw_endpoint: None,
+                    swap_info: None,
+                },
+            },
+        }])?;
+
+        self.persister.insert_payment_external_info(
```
roeierez, line 940: Why do we need to add the amount to the sync db? Is there a reason to sync pending payments?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
dleutenegger: Because the `attempted_amount_msat` is also used for failed open invoices to display the actual amount that was attempted to be paid.
And I assume, although not a 100% sure, that other signers would be able to sync the pending payment from Greenlight as well and thus need the data to display the accurrate amount.

roeierez: But we already handled failed opened invoices. Why persist for every invoice? For success payments the amount is retrieved correctly.
I see disadvantage for this way of sync where the synced data become larger without any real advantage.

dleutenegger: I see your point, but if we don't sync the attempted amount we couldn't display the actual intended amount for pending payments on all remote signers, as `amount_msat` is only the [amount delivered to destination](https://docs.corelightning.org/reference/lightning-listsendpays#return-value) (we could display it on the sending signer if we would just store it locally), which seems problematic to me as this would result it inconsistencies.
The other option I see would be to clean up the synced entry after payment success.

I would like to get your input on this

roeierez: > I see your point, but if we don't sync the attempted amount we couldn't display the actual intended amount for pending payments on all remote signers, as `amount_msat` is only the [amount delivered to destination](https://docs.corelightning.org/reference/lightning-listsendpays#return-value) (we could display it on the sending signer if we would just store it locally), which seems problematic to me as this would result it inconsistencies. The other option I see would be to clean up the synced entry after payment success.
> 
> I would like to get your input on this

The signer is synced with the channel state from greenlight not using our own sync db mechanism to it is irrelevant to the signer sync.
The sdk itself using the sync db to synchronize but it doesn't do that in realtime and we don't have a way to notify remote sdk instances to sync so I don't see how adding it to the sync db contributes here.
In addition immediately after the send_payment is called remote sdk that will sync (actually call list_pays behind the scene) will receive the pending payment as pending so that will should your concern as well.

dleutenegger: Do I understand this correctly; Other signers might get the pending payment (`list_pays`) before they sync the remote DB?

roeierez: > Do I understand this correctly; Other signers might get the pending payment (`list_pays`) before they sync the remote DB?

In short Yes, exactly.
Let me elaborate a bit on the way we sync data across sdk instances which should make it even more clear:
there are two ways we synchronize:
1. Sync remote data that exists on greenlight (payments, invoices, channels, etc...) this data comes from greenlight and is fetched by simple calling greenlight gRpc API. All of it is encapsulated inside the `sync` method.
2. Sync remote data that doesn't exist in greenlight and only the sdk "knows" about it (lnurl metadata, swaps, etc...) for this kind of data we use the sync.db and whenever such data changes locally the sdk ensures it is uploaded to the greenlight storage (using internal `backup` API) so other sdk instances can pull it.

The pending payments are related to the first kind of sync (greenlight data) so in theory we don't need to do anything about it as if one sdk instance pays, other instances will receive it as soon as they call `sync`.
Your requirement was a local requirement as far as I understand, to be able to see the pending payment **immediately** after calling `send_payment`.  For that purpose only adding locally is needed and no need to add it to the sync.db as well.
Hope that makes sense.

roeierez:
@dleutenegger could you please rebase it on top of master, currently there are some conflicts.

roeierez (approved):
LGTM

roeierez:
Will wait for @dangeross to finish his review before merging.

dangeross (approved):
Looks good!

PR #624, opened 2023-11-16
Only include the LSP hop hint if a channel open is needed

### LINKED ISSUE #539: Remove all other hop hints from invoices that need a channel open
If an open channel is needed, because inbound liquidity is too low to receive a payment, a hop hint to the LSP is added to the invoice. If the user's node has channels to other nodes however, these hop hints will still be in the created invoice. That means a multipart payment could route some parts over the lsp, but some parts over the other connected nodes. Routing some parts over the other connected nodes breaks the lsp model, because the assumption of the lsp currently is that all parts of the payment are routed over the lsp.

TL;DR if a channel open is needed, _only_ include a hop hint to the lsp in the invoice.

dangeross:
Fixes #539

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1967,8 +1967,12 @@ impl Receiver for PaymentReceiver {
         // We only create a new invoice if we need to add the lsp hint or change the amount
         if lsp_hint.is_some() || req.amount_msat != destination_invoice_amount_msat {
             // create the large amount invoice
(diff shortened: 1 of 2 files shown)


JssDWt (approved):
Tested with a node with channels to multiple nodes.
Includes only the hop hint to the LSP for amounts > inbound liquidity
Includes hop hints to all nodes for amounts < inbound liquidity.

roeierez (approved):
LGTM

PR #627, opened 2023-11-20
Report payment failures

### LINKED ISSUE #618: Add API to report payment failures
We need an option for the user to report payment failures. We can start with the following information:
node_id
timestamp

Perhaps we can use this API from breezmobile: https://github.com/breez/breez/blob/9a19c9ec635d749313fa563504ff35d2d82f0219/account/payments.go#L797

dangeross:
This PR adds an issue reporting interface to initially support reporting payment failures to the Support API.

I have implemented a generic reporting interface to allow it to extended later by adding extra variants on the req enum param. Perhaps with the option to support reporting other issues around service availability, lsp or swap failures.

Depends on breez/server#35
Fixes #618

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -596,6 +596,26 @@ dictionary LnUrlAuthRequestData {
     string url;
 };
 
+dictionary ReportPaymentFailureDetails {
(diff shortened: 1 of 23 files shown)


roeierez (approved):
LGTM

roeierez (approved):
LGTM

PR #638, opened 2023-11-23
Validate selected LSP

### LINKED ISSUE #634: Unable to receive payments if persisted LSP is removed from LSP list
If there is an existing channel with an LSP (A), but the LSP list for an API key is updated to switch to LSP (B), the node cannot receive payments to LSP (A)'s channel because the LSP lookup of the node's persisted LSP (A) fails.

dangeross:
1. Validate that the selected LSP is still in list_lsps, else select the first LSP in the list. 
2. Only add a route hint if optional short_channel_id is set. This ensures only a route hint is added for channel opening, or if there is an existing channel open with the selected LSP. Route hints are not added for previously used LSPs.

Fixes #634

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1967,7 +1967,7 @@ impl Receiver for PaymentReceiver {
             }
         );
 
-        let mut short_channel_id = parse_short_channel_id("1x0x0")?;
+        let mut short_channel_id = None;
         let mut destination_invoice_amount_msat = req.amount_msat;
 
         let mut channel_opening_fee_params = None;
@@ -1984,6 +1984,7 @@ impl Receiver for PaymentReceiver {
                 None => lsp_info.cheapest_open_channel_fee(expiry)?.clone(),
             };
 
+            short_channel_id = Some(parse_short_channel_id("1x0x0")?);
             channel_opening_fee_params = Some(ofp.clone());
             channel_fees_msat = Some(ofp.get_channel_fees_msat_for(req.amount_msat));
             if let Some(channel_fees_msat) = channel_fees_msat {
@@ -2016,8 +2017,8 @@ impl Receiver for PaymentReceiver {
                         .alias_remote
                         .unwrap_or(active_channel.clone().short_channel_id);
 
-                    short_channel_id = parse_short_channel_id(&hint)?;
-                    info!("Found channel ID: {short_channel_id} {active_channel:?}");
+                    short_channel_id = Some(parse_short_channel_id(&hint)?);
+                    info!("Found channel ID: {short_channel_id:?} {active_channel:?}");
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -917,13 +917,18 @@ impl BreezServices {
...
+    /// This validates if the selected LSP is still in [`list_lsps`].
+    /// If not or no LSP is selected, it selects the first LSP in [`list_lsps`].
     async fn connect_lsp_peer(&self, node_pubkey: String) -> SdkResult<()> {
-        // Sets the LSP id, if not already set
-        if self.persister.get_lsp_id()?.is_none() {
-            if let Some(lsp) = self.lsp_api.list_lsps(node_pubkey).await?.first().cloned() {
-                self.persister.set_lsp_id(lsp.id)?;
-            }
+        let lsps = self.lsp_api.list_lsps(node_pubkey).await?;
+        if let Some(lsp) = self
```
JssDWt, line 925: What if this is `None`, but we do have an lsp persisted already. Will that cause issues?

JssDWt (commented):
This should fix the hop hint problem where a hop hint was created for a lsp you don't have channels with.

Small question about the connect function.

libs/sdk-core/src/breez_services.rs
dangeross: It would mean `list_lsps` is empty. We would be back to `receive_payment` failing, which would also fail if the persisted lsp is not set.

JssDWt (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -917,13 +917,18 @@ impl BreezServices {
...
+    /// Connects to the selected LSP peer.
+    /// This validates if the selected LSP is still in [`list_lsps`].
+    /// If not or no LSP is selected, it selects the first LSP in [`list_lsps`].
     async fn connect_lsp_peer(&self, node_pubkey: String) -> SdkResult<()> {
-        // Sets the LSP id, if not already set
-        if self.persister.get_lsp_id()?.is_none() {
-            if let Some(lsp) = self.lsp_api.list_lsps(node_pubkey).await?.first().cloned() {
-                self.persister.set_lsp_id(lsp.id)?;
-            }
+        let lsps = self.lsp_api.list_lsps(node_pubkey).await?;
```
roeierez, line 924: This works but has a disadvantage of making this network request on every call of this method.
Below we are calling `lsp_info` which already cals `list_lsps` and return None in case the lsp is not in the list.
Perhaps an optimistic approach where only re-selecting if `lsp_info` return `None` is more efficient?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
dangeross: Agreed, it not terribly efficient. The lsp_info call is also used to validate the node state exists. 
I think I can rewrite it so it makes 1 network request, instead of min 1 / max 2 network requests.

libs/sdk-core/src/breez_services.rs
```diff
@@ -917,30 +917,36 @@ impl BreezServices {
...
-                    })?;
+        let lsps = self.lsp_api.list_lsps(node_pubkey).await?;
+        if let Some(lsp) = self
+            .persister
+            .get_lsp_id()?
+            .and_then(|lsp_id| lsps.clone().into_iter().find(|lsp| lsp.id == lsp_id))
+            .or_else(|| lsps.first().cloned())
+        {
+            self.persister.set_lsp_id(lsp.id)?;
+            if let Ok(node_state) = self.node_info() {
```
dangeross, line 932: Now does a maximum of one network request. Let me know what you think @roeierez

roeierez (approved):
LGTM

Issue #639, opened 2023-11-23
Optimize `NodeAPI::pull_transactions`: filter remotely


ok300:
In `pull_transactions` we are calling `list_invoices` [^1] and `list_pays`[^2] with no remote filtering.

After retrieving all invoices and payments, we filter them locally by timestamp, and only return a subset for further processing.

This brings network overhead on every `sync`, especially for nodes with a large history.

It might be worth looking into ways to filter remotely. `list_invoices` already supports a kind of filtering by `created` or `updated` timestamps.

[^1]: https://docs.corelightning.org/reference/lightning-listinvoices
[^2]: https://docs.corelightning.org/reference/lightning-listpays

PR #640, opened 2023-11-23
Log full list of payments on trace instead of debug

ok300:
This reduces the typical log size for a pulse test from 40-50MB to ~1MB.

Related: #639


JssDWt (approved):
That's a lot of MBs, LGTM!

roeierez (approved):

PR #646, opened 2023-11-24
Remove command "stop" from `disconnect` and NodeAPI trait

ok300:
As discussed in https://github.com/breez/breez-sdk/pull/612#discussion_r1395572300 , removing the call to `NodeAPI::stop()` from `disconnect` to avoid unpredictable behavior when using the same seed on multiple clients.

Since it's not meant to be called in any service method because of the above, I also removed it from the trait.

`stop` can still be called as a dev command.


roeierez (approved):
LGTM

JssDWt (approved):
LGTM

PR #652, opened 2023-11-24
Validate RN optional field values

dangeross:
Rerun the RN codegen to add optional field value validation. This update should prevent set optional values being passed to the SDK as **nil** if the value cast/conversion fails. Instead an error will be thrown. 

An example of this is calling `fetchReverseSwapFees` with a float value:
```
await fetchReverseSwapFees({ sendAmountSat: 50000.75 })
```
would return a valid response with a null `totalEstimatedFees`:
```
{
    "feesClaim": 12696, 
    "feesHash": "0a64d2ab96aaa630baa38f7a235511a29f7718759fd7784972397428ce055223", 
    "feesLockup": 14076, 
    "feesPercentage": 0.5, 
    "max": 25000000, 
    "min": 50000, 
    "totalEstimatedFees": null
}
```
whereas it should throw an error:
```
Error: Generic: Unexpected value for optional field sendAmountSat
```

Generator PR: breez/breez-sdk-rn-generator#18

### DIFF
--- a/libs/sdk-react-native/ios/BreezSDKMapper.swift
+++ b/libs/sdk-react-native/ios/BreezSDKMapper.swift
@@ -3,8 +3,12 @@ import Foundation
 
 enum BreezSDKMapper {
     static func asAesSuccessActionDataDecrypted(aesSuccessActionDataDecrypted: [String: Any?]) throws -> AesSuccessActionDataDecrypted {
-        guard let description = aesSuccessActionDataDecrypted["description"] as? String else { throw SdkError.Generic(message: "Missing mandatory field description for type AesSuccessActionDataDecrypted") }
-        guard let plaintext = aesSuccessActionDataDecrypted["plaintext"] as? String else { throw SdkError.Generic(message: "Missing mandatory field plaintext for type AesSuccessActionDataDecrypted") }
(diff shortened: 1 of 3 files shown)


libs/sdk-react-native/ios/BreezSDKMapper.swift
```diff
@@ -2154,7 +2642,13 @@ enum BreezSDKMapper {
...
 
     static func asReverseSwapFeesRequest(reverseSwapFeesRequest: [String: Any?]) throws -> ReverseSwapFeesRequest {
-        let sendAmountSat = reverseSwapFeesRequest["sendAmountSat"] as? UInt64
+        var sendAmountSat: UInt64?
+        if hasNonNilKey(data: reverseSwapFeesRequest, key: "sendAmountSat") {
+            guard let sendAmountSatTmp = reverseSwapFeesRequest["sendAmountSat"] as? UInt64 else {
+                throw SdkError.Generic(message: errUnexpectedValue(fieldName: "sendAmountSat"))
+            }
+            sendAmountSat = sendAmountSatTmp
+        }
```
dangeross, line 2651: Check using `hasNonNilKey` and guard the cast throwing an error if nil

roeierez (approved):
LGTM

PR #664, opened 2023-11-30
Add BreezSDK.swift as an on demand resource on Flutter plugin

erdemyerebasmaz:
Importing `breez-sdk-swift` package natively on a **Flutter** project that also depended on `breez-sdk-flutter` Flutter plugin was causing `.xcframework` dependency conflicts. 

To overcome this issue and prevent unnecessary import of the Swift package, we decided to add `BreezSDK.swift` as an on demand resource on **Flutter** plugin. 

With this change, `BreezSDK.swift` will be accessible natively by adding it as a **Compile Source** under **Build Phases** through **XCode**.

### Changelist
- Add `BreezSDK.swift `as an on demand resource on **Flutter** plugin
  - Both `.podspec` files now use the same format with this change, which can make reviewing it a bit hard. To clarify, only addition is: 
  `s.on_demand_resources = { 'BreezSDK' => 'bindings-swift/Sources/BreezSDK/BreezSDK.swift' }`
- Use **Flutter** 3.16.1 on CI workflows
- Update dependencies to latest
- Updated flutter plugin & pod version

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -245,7 +245,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: '3.13.9'
+          flutter-version: '3.16.1'
           channel: 'stable'
 
       - name: Flutter bridge codegen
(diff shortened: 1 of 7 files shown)


JssDWt (approved):
Looks good.

Question: we update the pubspec.yaml version in the publish-flutter CI, but not the podspec version. Should we?

erdemyerebasmaz:
@JssDWt 

> Looks good.
> 
> Question: we update the pubspec.yaml version in the publish-flutter CI, but not the podspec version. Should we?

I believe we should to avoid any possible cache issues when installing pods.

> @erdemyerebasmaz Do you think it's a good idea to use latest stable flutter version in CI?

Yes, we aim to keep up to date with Flutter's latest stable channel. 

We can add `continue-on-error: true` to `publish-flutter` job of `publish-all-platforms.yml`, which seems like the only time consuming workflow we wouldn't want to exit due to a Flutter version issue.

PR #672, opened 2023-12-08
Add Kotlin bindings to flutter CI

JssDWt:
In https://github.com/breez/breez-sdk/pull/670 the kotlin bindings are added to the sdk-flutter plugin. This change integrates that into the CI. The breez_sdk.kt file was committed, I removed that. It's not necessary to add generated files into source control.


libs/sdk-flutter/.gitignore
```diff
@@ -1,2 +1,3 @@
 build/
-.flutter-version
\ No newline at end of file
+.flutter-version
+android/src/main/kotlin/breez_sdk/
```
erdemyerebasmaz, line 3: Should we delete and ignore swift bindings as well?

erdemyerebasmaz (approved):
Looks good

libs/sdk-flutter/.gitignore
JssDWt: Yeah makes sense, will do

JssDWt: Good one, ignoring the swift bindings now too.
They weren't committed (I think, or I'm losing my eyesight ;-))

PR #673, opened 2023-12-08
Persist payment failure error

### LINKED ISSUE #626: Feature request: provide payment failure reason for failed payments
It would be useful if, when sending a payment fails, besides providing the failure reason as an error returned by e.g., `send_payment()`, the failure reason was also included within `Payment` when this payment is fetched using `list_payments()` or `payment_by_hash()`.

dangeross:
This PR adds the last payment failure error reason into the external info table. An added benefit of this is the error information is also included when reporting a payment failure.

Fixes #626

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -392,11 +392,14 @@ impl BreezServices {
                 // Store SA (if available) + LN Address in separate table, associated to payment_hash
                 self.persister.insert_payment_external_info(
                     &details.payment_hash,
-                    maybe_sa_processed.as_ref(),
-                    Some(req.data.metadata_str),
-                    req.data.ln_address,
-                    None,
-                    invoice.amount_msat,
+                    PaymentExternalInfo {
+                        lnurl_pay_success_action: maybe_sa_processed.clone(),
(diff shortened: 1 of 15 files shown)


ok300 (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -1012,6 +1022,19 @@ impl BreezServices {
                 Ok(payment)
             }
             Err(e) => {
+                if let Some(invoice) = invoice.clone() {
+                    // Only persist errors specific to a node or route related failure
+                    match e.clone() {
+                        SendPaymentError::PaymentFailed { err }
```
roeierez, line 1028: Perhaps we should persist any error? I am concerned we will miss errors when we add more errors to the enum.
Also worst case we will add an entry here without a failure in greenlight that will never be used.
What do you think?

roeierez (changes requested):
One comment, otherwise LGTM.

libs/sdk-core/src/breez_services.rs
dangeross: Agreed and done. We can always refine later.

roeierez (approved):
LGTM

PR #674, opened 2023-12-10
Payment expiry for pending payments

ubbabeck:
This pr aims to add expiry time to pending payments.

Fixes [#679](https://github.com/breez/c-breez/issues/679)

## For reviews

I am new to rust, so if you see something which is not best practise or could be done functionally or in a _rustier_ way  please let me know.


libs/sdk-core/src/breez_services.rs
```diff
@@ -903,7 +903,29 @@ impl BreezServices {
...
+                        {
+                            info!(
+                                "htlc payment is {:?}",
+                                payment.status == PaymentStatus::Pending
+                            );
+                        }
+                    }
+                }
+            }
+        }
```
roeierez, line 926: This whole logic of matching the htlc to the specific pending payment should be inside the NodeAPI implementation.

ubbabeck: Done moved to NodeAPI

libs/sdk-core/src/breez_services.rs
```diff
@@ -910,8 +910,21 @@ impl BreezServices {
...
+            .clone()
+            .into_iter()
+            .filter(|c| c.status == PaymentStatus::Pending)
+            .collect();
+        let mut payments: Vec<Payment> = vec![];
+        if !pending_payments.is_empty() {
+            info!("pending paymetns are not empty {:?}", pending_payments);
+            payments.extend(self.node_api.htlc_expiry(pending_payments).await?);
+        }
+        info!("payment list consist of {:?}", payments);
```
roeierez, line 924: I think all this needs to be inside node_api. We don't need another method for `htlc_expiry` and we rather ensure the `pull_changed` method returns the payments with the right expiry populated.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ubbabeck: Moved this logic.

ubbabeck:
Changes since last commit include converting block difference to unix time and persisting the payment_expiry to the db.

What needs to be fixed is:

- [ ] When a payment consists of multiple htlcs, we only need to care about the one expiring first
- [ ] Don't iterate over already pending payments more than once, atm it goes though this on every sync.

ubbabeck:
Instead of converting number of block to a unix timestamp it might be better to let return the block when the htlc is expiring. 

The reasoning is this : 

- even though the blocktime is approximately 10min it's usually quicker. 
- Developers might want to display block expiry rather than the calculated time, this calculation can be handled in the UI
- Relying on a timestamp timeout can backfire if notifications are sent too late. Even though we'll eventually add a webhook which handles this.
- The htlc will most likely be settled long before this timeout.

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1452,13 +1452,53 @@ async fn pull_transactions(
         .map(TryInto::try_into)
         .collect();
 
+    let res = c
```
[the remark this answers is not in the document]
ubbabeck: True, fixed.

libs/sdk-core/src/models.rs
```diff
@@ -1098,6 +1109,23 @@ pub struct Channel {
     ///
     /// This may be empty for older closed channels, if it was not possible to retrieve the closing txid.
     pub closing_txid: Option<String>,
+
+    pub htlc: Option<Vec<Htlc>>,
```
roeierez, line 1113: I see this field is never None in the code so we can omit the Option as the vector can be empty instead.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -282,6 +282,7 @@ dictionary LnPaymentDetails {
     string? ln_address;
     string? lnurl_withdraw_endpoint;
     SwapInfo? swap_info;
+    u32? payment_expiry;
```
roeierez, line 285: Perhaps `pending_expiration_block` ?

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
ubbabeck: Agree, renamed.

libs/sdk-core/src/models.rs

roeierez (approved):
LGTM

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1452,13 +1458,40 @@ async fn pull_transactions(
         .map(TryInto::try_into)
         .collect();
 
+    let outbound_transactions: NodeResult<Vec<Payment>> =
+        htlc_expiry(outbound_transactions.unwrap(), htlc_list);
```
dangeross, line 1462: To avoid a potential panic
```suggestion
        htlc_expiry(outbound_transactions?, htlc_list);
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1452,13 +1458,40 @@ async fn pull_transactions(
...
+
     let mut transactions: Vec<Payment> = Vec::new();
     transactions.extend(received_transactions?);
     transactions.extend(outbound_transactions?);
 
     Ok(transactions)
 }
 
+fn htlc_expiry(payments: Vec<Payment>, htlc_list: Vec<Htlc>) -> NodeResult<Vec<Payment>> {
```
dangeross, line 1471: The fn name doesn't really tell me what the function does, I had to look at the code. Maybe rename to `update_payment_expirations`

libs/sdk-core/src/models.rs
```diff
@@ -1098,6 +1109,23 @@ pub struct Channel {
     ///
     /// This may be empty for older closed channels, if it was not possible to retrieve the closing txid.
     pub closing_txid: Option<String>,
+
+    pub htlc: Vec<Htlc>,
```
dangeross, line 1113: I think this should be plural
```suggestion
    pub htlcs: Vec<Htlc>,
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1452,13 +1458,40 @@ async fn pull_transactions(
...
+    let mut pending_payments: Vec<Payment> = Vec::new();
```
dangeross, line 1472: From what I understand in `pull_transactions` the payments could be pending or completed, maybe rename to something more general like `payments_res`. I assume if the payment is completed there will be no matching HTLC, so no need to filter.

libs/sdk-core/src/models.rs
```diff
@@ -1098,6 +1109,23 @@ pub struct Channel {
...
+    pub payment_hash: Vec<u8>,
+}
+
+impl Htlc {
+    pub fn from(expiry: u32, payment_hash: Vec<u8>) -> Self {
+        Htlc {
+            expiry,
+            payment_hash,
+        }
+    }
```
dangeross, line 1128: I don't think this is used, could be removed

dangeross (changes requested):

libs/sdk-core/src/models.rs
ubbabeck: I did not remove it and instead use it when listing channel peers. Not sure which approach is preferred.

dangeross (approved):
LGTM!

PR #678, opened 2023-12-11
Don't override pending payments.

roeierez:
When we are using insert_or_update_payments we delete pending payments. Historically it was because this function only was called at the end of the sync. Now that we call it from different flows it is not longer correct. Added a flag for deleting pending payments and only set it within the sync.


dangeross (approved):
LGTM!

PR #679, opened 2023-12-11
Add extra TLVs to send spontaneous payment

### LINKED ISSUE #643: Permit adding TLV records to a spontaneous keysend payment
Currently only the node id and amount can be specified for a spontaneous payment. It would open up some use cases, in which no proof of payment is really needed, if TLV records could be added too.

dangeross:
This PR allows optional TLVs to be added to a spontaneous payment.

Fixes #643


roeierez (approved):
LGTM

Issue #666, opened 2023-12-02
Lowercase lightning address (LUD-16) before validation


andrei-21:
According to the standard LUD-16 capital letters are not allowed in the username part. However in practice email addresses are _considered_ case insensitive, since lightning addresses look very similar people may expect similar behavior. I suggest to lowercase a lightning address before validation and return the canonical username (e.g. `Satoshi@bitcoin.org` will be `satoshi@bitcoin.org`).

It will solve the problem that smartphones' keyboards sometimes uppercase words and can make a lightning address invalid.

PR #681, opened 2023-12-11
change: lud-16 assersions to ignore cases

rustaceanrob:
Users may submit LN addresses like `User@domain.com`, addressed in #666. Instead of throwing an exception the username is automatically lower-cased before validation. For example,`User@domain.com` -> `https://domain.com/.well-known/lnurlp/user`


dangeross (approved):
Thanks for the PR @rustaceanrob! LGTM!

roeierez (approved):
LGTM

Issue #680, opened 2023-12-11
Order the `list_fiat_currencies` response


dangeross:
The result of `list_fiat_currencies` is returned in different order each time.

ok300:
Fixed in #686

PR #682, opened 2023-12-12
enhancement: sort fiat currencies by name

rustaceanrob:
The result of `list_fiat_currencies` was unsorted as addressed in issue #680. Changed to sort by the canonical name of the currency.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -587,12 +587,13 @@ impl BreezServices {
         Ok(response)
     }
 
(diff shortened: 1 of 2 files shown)


ok300:
Looks good!

There are a few other places that are worth updating as part of this PR:
- `fetch_fiat_rates` (just underneath the method you changed) could use a similar sorting
- the rustdoc for both `fetch_fiat_rates` and `list_fiat_currencies`, located in `breez_services.rs`, could be updated to indicate the returned lists are sorted

Could you add those here as well?

rustaceanrob:
Can do!

PR #683, opened 2023-12-12
Revert "Schedule signer if `GL::connect()` has seed and creds"

ok300:
Reverts breez/breez-sdk#596

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -116,11 +116,7 @@ impl Greenlight {
                 match encrypted_creds {
                     Some(c) => {
                         persister.set_gl_credentials(c)?;
-                        let gl_node = Greenlight::new(config, seed, creds, persister).await?;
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -116,11 +116,7 @@ impl Greenlight {
                 match encrypted_creds {
                     Some(c) => {
                         persister.set_gl_credentials(c)?;
-                        let gl_node = Greenlight::new(config, seed, creds, persister).await?;
-                        // Make small API call to ensure the signer can be scheduled. This will fail if there are any network issues.
-                        // The other two scenarios (recover, register) already include a network call.
-                        gl_node.start().await?;
-                        Ok(gl_node)
+                        Greenlight::new(config, seed, creds, persister).await
```
roeierez, line 119: It seems that the `Greenlight::new` doesn't need the `async` modifier and we can get rid of the `await` here.

roeierez (changes requested):
Caught something along the way, not seems to be related to the PR but worth fixing.
Other than that LGTM

PR #684, opened 2023-12-12
rename sweep to redeem_onchain_funds

ubbabeck:
In order to avoid confusion about sweep which implies sending funds from the closing address to the internal wallet. By renaming it to withdraw we indicate that it is sending the funds out.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -658,7 +658,7 @@ enum BuyBitcoinProvider {
 
 dictionary PrepareSweepRequest {
     string to_address;
-    u64 sat_per_vbyte;
+    u32 sat_per_vbyte;
 };
 
 dictionary PrepareSweepResponse {
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -656,22 +656,22 @@ enum BuyBitcoinProvider {
     "Moonpay",
 };
 
-dictionary PrepareSweepRequest {
+dictionary PrepareRedeemOnchainFundsRequest {
     string to_address;
     u32 sat_per_vbyte;
 };
 
-dictionary PrepareSweepResponse {
-    u64 sweep_tx_weight;
-    u64 sweep_tx_fee_sat;
+dictionary PrepareRedeemOnchainFundsResponse {
+    u64 tx_weight;
+    u64 tx_fee_sat;
 };
 
-dictionary SweepRequest {
+dictionary RedeemOnchainFundsRequest {
     string to_address;
     u32 sat_per_vbyte;
 };
 
-dictionary SweepResponse {
+dictionary RedeemOnchainFundsResponse {
     sequence<u8> txid;
 };
 
@@ -777,7 +777,7 @@ interface BlockingBreezServices {
    sequence<Payment> list_payments(ListPaymentsRequest req);
 
    [Throws=SdkError]
-   SweepResponse sweep(SweepRequest req);
+   RedeemOnchainFundsResponse redeem_onchain_funds(RedeemOnchainFundsRequest req);
 
    [Throws=SdkError]
    sequence<Rate> fetch_fiat_rates();
@@ -852,7 +852,7 @@ interface BlockingBreezServices {
    BuyBitcoinResponse buy_bitcoin(BuyBitcoinRequest req);
 
    [Throws=SdkError]
-   PrepareSweepResponse prepare_sweep(PrepareSweepRequest req);
+   PrepareRedeemOnchainFundsResponse prepare_redeem_onchain_funds(PrepareRedeemOnchainFundsRequest req);
 };
 
 namespace breez_sdk {  
--- a/libs/sdk-bindings/src/uniffi_binding.rs
+++ b/libs/sdk-bindings/src/uniffi_binding.rs
@@ -14,15 +14,16 @@ use breez_sdk_core::{
     MaxReverseSwapAmountResponse, MessageSuccessActionData, MetadataItem, Network, NodeConfig,
     NodeCredentials, NodeState, OpenChannelFeeRequest, OpenChannelFeeResponse, OpeningFeeParams,
     OpeningFeeParamsMenu, Payment, PaymentDetails, PaymentFailedData, PaymentStatus, PaymentType,
-    PaymentTypeFilter, PrepareRefundRequest, PrepareRefundResponse, PrepareSweepRequest,
-    PrepareSweepResponse, Rate, ReceiveOnchainRequest, ReceivePaymentRequest,
-    ReceivePaymentResponse, RecommendedFees, RefundRequest, RefundResponse, ReportIssueRequest,
+    PaymentTypeFilter, PrepareRedeemOnchainFundsRequest, PrepareRedeemOnchainFundsResponse,
+    PrepareRefundRequest, PrepareRefundResponse, Rate, ReceiveOnchainRequest,
+    ReceivePaymentRequest, ReceivePaymentResponse, RecommendedFees, RedeemOnchainFundsRequest,
+    RedeemOnchainFundsResponse, RefundRequest, RefundResponse, ReportIssueRequest,
     ReportPaymentFailureDetails, ReverseSwapFeesRequest, ReverseSwapInfo, ReverseSwapPairInfo,
     ReverseSwapStatus, RouteHint, RouteHintHop, SendOnchainRequest, SendOnchainResponse,
     SendPaymentRequest, SendPaymentResponse, SendSpontaneousPaymentRequest,
     ServiceHealthCheckResponse, SignMessageRequest, SignMessageResponse, StaticBackupRequest,
-    StaticBackupResponse, SuccessActionProcessed, SwapInfo, SwapStatus, SweepRequest,
-    SweepResponse, Symbol, TlvEntry, UnspentTransactionOutput, UrlSuccessActionData,
+    StaticBackupResponse, SuccessActionProcessed, SwapInfo, SwapStatus, Symbol, TlvEntry,
+    UnspentTransactionOutput, UrlSuccessActionData,
 };
 use log::{Level, LevelFilter, Metadata, Record};
 use once_cell::sync::{Lazy, OnceCell};
@@ -195,8 +196,11 @@ impl BlockingBreezServices {
         rt().block_on(self.breez_services.report_issue(req))
     }
 
-    pub fn sweep(&self, req: SweepRequest) -> SdkResult<SweepResponse> {
-        rt().block_on(self.breez_services.sweep(req))
+    pub fn redeem_onchain_funds(
+        &self,
+        req: RedeemOnchainFundsRequest,
+    ) -> SdkResult<RedeemOnchainFundsResponse> {
+        rt().block_on(self.breez_services.redeem_onchain_funds(req))
     }
 
     pub fn fetch_fiat_rates(&self) -> SdkResult<Vec<Rate>> {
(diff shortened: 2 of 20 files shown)


JssDWt:
I find `withdraw` more confusing than `sweep` tbh. We have lnurl withdraw for example, which means receiving funds into your sdk balance. Withdrawing something is usually incoming. The terms deposit/withdraw are generally confusing and should be avoided altogether imo. Send and receive are clearer, because the direction is clear.

Sweep means claim onchain funds that are already yours. Since you cannot access your onchain funds through the sdk, and you have to provide an address, the ring for me is "claim funds that are already yours into this specific address".

kingonly:
> I find `withdraw` more confusing than `sweep` tbh. We have lnurl withdraw for example, which means receiving funds into your sdk balance. Withdrawing something is usually incoming. The terms deposit/withdraw are generally confusing and should be avoided altogether imo. Send and receive are clearer, because the direction is clear.
> 
> Sweep means claim onchain funds that are already yours. Since you cannot access your onchain funds through the sdk, and you have to provide an address, the ring for me is "claim funds that are already yours into this specific address".

The problem is that "sweep" has a specific semantics in Lightning (moved funds from a closed channel to the node's wallet). Our API is different, it's about moving funds from the node's wallet to another BTC address. That's why "withdraw on-chain" or something would be better imo.

kingonly:
"transfer on chain funds"?

dangeross:
I agree with @JssDWt, withdraw can be confusing because it used elsewhere as an incoming/receiving action. I think `transfer_onchain` could work like @kingonly said

kingonly:
Not sure we can do better with transfer. I agree withdraw isn't great, but sweep is very confusing to Lightning devs and I don't think we should overload it with new semantics. Open to suggestions.

ubbabeck:
I agree that withdraw also can cause confusion since this is a lighting first sdk. One suggestion is to use transaction as it is reserved for on-chain and payment is reserved for lighting. 

Suggestions  `sweepTransaction` or `withdrawTransaction`. Thought they might be a bit too verbose when combined eg.  `PrepareWithdrawTransactionReqeust` or `PrepareSweepTransactionRequest`.

kingonly:
It doesn't help. Sweep transaction is even worse because it refers to a very specific transaction (moving the funds from a closed channel to an the node's on-chain wallet). transfer_onchain_funds (which echo the 'unexpected funds' UI is the best option so far imo.

ubbabeck:
Since the last commit I've renamed the sweep to transfer_onchain_funds and prepare_sweep to prepare_transfer_onchain_funds. 

Do we also wish to use the same type of unsigned integer for both of the methods? 

Which of u32 or u64 is preferred?

dangeross:
u32 is the most commonly used for sat_per_vbyte.

My other options are `remit_onchain` or `forward_onchain`

ok300:
To indicate we're dealing with onchain, we could include `utxo` in the name, like `sweep_utxos` or `transfer_utxos`.

One step further, to include directionality: `transfer_utxos_to` (although IMO `sweep_utoxs` includes direction and separates it from LN semantics).

kingonly:
on-chain is better and clearer than utxo.
transfer_onchain is still better the remit or forward.

roeierez (approved):
LGTM

libs/sdk-bindings/src/uniffi_binding.rs
```diff
@@ -195,7 +196,7 @@ impl BlockingBreezServices {
         rt().block_on(self.breez_services.report_issue(req))
     }
 
-    pub fn sweep(&self, req: SweepRequest) -> SdkResult<SweepResponse> {
+    pub fn sweep(&self, req: RedeemOnchainFundsRequest) -> SdkResult<RedeemOnchainFundsResponse> {
         rt().block_on(self.breez_services.sweep(req))
     }
```
dangeross, line 201:
```suggestion
    pub fn redeem_onchain_funds(&self, req: RedeemOnchainFundsRequest) -> SdkResult<RedeemOnchainFundsResponse> {
        rt().block_on(self.breez_services.redeem_onchain_funds(req))
    }
```

libs/sdk-core/src/binding.rs
```diff
@@ -360,15 +361,22 @@ pub fn buy_bitcoin(req: BuyBitcoinRequest) -> Result<BuyBitcoinResponse> {
 }
 
 /// See [BreezServices::sweep]
```
dangeross, line 363:
```suggestion
/// See [BreezServices::redeem_onchain_funds]
```

libs/sdk-core/src/binding.rs
```diff
@@ -360,15 +361,22 @@ pub fn buy_bitcoin(req: BuyBitcoinRequest) -> Result<BuyBitcoinResponse> {
...
-pub fn sweep(req: SweepRequest) -> Result<SweepResponse> {
+pub fn sweep(req: RedeemOnchainFundsRequest) -> Result<RedeemOnchainFundsResponse> {
     block_on(async { get_breez_services().await?.sweep(req).await })
         .map_err(anyhow::Error::new::<SdkError>)
 }
```
dangeross, line 367:
```suggestion
pub fn redeem_onchain_funds(req: RedeemOnchainFundsRequest) -> Result<RedeemOnchainFundsResponse> {
    block_on(async { get_breez_services().await?.redeem_onchain_funds(req).await })
        .map_err(anyhow::Error::new::<SdkError>)
}
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -576,20 +576,26 @@ impl BreezServices {
         Ok(self.persister.get_payment_by_hash(&hash)?)
     }
 
-    /// Sweep on-chain funds to the specified on-chain address, with the given feerate
-    pub async fn sweep(&self, req: SweepRequest) -> SdkResult<SweepResponse> {
+    /// RedeemOnchainFunds on-chain funds to the specified on-chain address, with the given feerate
+    pub async fn sweep(
```
dangeross, line 580:
```suggestion
    pub async fn redeem_onchain_funds(
```

libs/sdk-core/src/node_api.rs
```diff
@@ -90,7 +90,7 @@ pub trait NodeAPI: Send + Sync {
         last_hop: Option<&RouteHintHop>,
     ) -> NodeResult<Vec<MaxChannelAmount>>;
     async fn sweep(&self, to_address: String, sat_per_vbyte: u32) -> NodeResult<Vec<u8>>;
```
dangeross, line 92:
```suggestion
    async fn redeem_onchain_funds(&self, to_address: String, sat_per_vbyte: u32) -> NodeResult<Vec<u8>>;
```

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -777,7 +777,7 @@ interface BlockingBreezServices {
    sequence<Payment> list_payments(ListPaymentsRequest req);
 
    [Throws=SdkError]
-   SweepResponse sweep(SweepRequest req);
+   RedeemOnchainFundsResponse sweep(RedeemOnchainFundsRequest req);
```
dangeross, line 780:
```suggestion
   RedeemOnchainFundsResponse redeem_onchain_funds(RedeemOnchainFundsRequest req);
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -576,20 +576,26 @@ impl BreezServices {
         Ok(self.persister.get_payment_by_hash(&hash)?)
     }
 
-    /// Sweep on-chain funds to the specified on-chain address, with the given feerate
-    pub async fn sweep(&self, req: SweepRequest) -> SdkResult<SweepResponse> {
+    /// RedeemOnchainFunds on-chain funds to the specified on-chain address, with the given feerate
```
dangeross, line 579:
```suggestion
    /// Redeem on-chain funds from closed channels to the specified on-chain address, with the given feerate
```

dangeross (changes requested):
Seems to be a couple of places where functions are still named `sweep`.

libs/sdk-bindings/src/uniffi_binding.rs
ubbabeck: Oh, my bad. Thanks, fixed with the latest push.

libs/sdk-core/src/binding.rs
ubbabeck: Done

libs/sdk-core/src/binding.rs
ubbabeck: Done

libs/sdk-core/src/breez_services.rs
ubbabeck: Updated

libs/sdk-core/src/models.rs
```diff
@@ -1298,21 +1298,21 @@ pub enum BuyBitcoinProvider {
     Moonpay,
 }
 
-/// We need to prepare a sweep transaction to know what fee will be charged in satoshis this
-/// model holds the request data which consists of the address to sweep to and the fee rate in
+/// We need to prepare a redeem_onchain_funds transaction to know what fee will be charged in satoshis this
+/// model holds the request data which consists of the address to redeem_onchain_funds to and the fee rate in
```
dangeross, line 1302:
```suggestion
/// We need to prepare a redeem_onchain_funds transaction to know what fee will be charged in satoshis. 
/// This model holds the request data which consists of the address to redeem on-chain funds to and the fee rate in
```

tools/sdk-cli/src/command_handlers.rs
```diff
@@ -232,29 +232,29 @@ pub(crate) async fn handle_command(
...
             sat_per_vbyte,
         } => {
             sdk()?
-                .sweep(SweepRequest {
+                .redeem_onchain_funds(RedeemOnchainFundsRequest {
                     to_address,
                     sat_per_vbyte,
                 })
                 .await?;
             Ok("Onchain funds were swept successfully".to_string())
```
dangeross, line 245: Maybe change this so we can see the txid
```suggestion
            serde_json::to_string_pretty(&resp).map_err(|e| e.into())
```

libs/sdk-core/src/models.rs
```diff
@@ -1298,21 +1298,21 @@ pub enum BuyBitcoinProvider {
...
-/// We need to prepare a sweep transaction to know what a fee it will be charged in satoshis
+/// We need to prepare a redeem_onchain_funds transaction to know what a fee it will be charged in satoshis
 /// this model holds the response data, which consists of the weight and the absolute fee in sats
 #[derive(PartialEq, Eq, Debug, Clone, Deserialize, Serialize)]
-pub struct PrepareSweepResponse {
-    pub sweep_tx_weight: u64,
-    pub sweep_tx_fee_sat: u64,
+pub struct PrepareRedeemOnchainFundsResponse {
+    pub redeem_onchain_funds_tx_weight: u64,
+    pub redeem_onchain_funds_tx_fee_sat: u64,
```
dangeross, line 1315: Same renaming as above

dangeross (changes requested):
Looking good. Just a couple more bits I missed, sorry 🙂

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -656,22 +656,22 @@ enum BuyBitcoinProvider {
...
-    u64 sat_per_vbyte;
+    u32 sat_per_vbyte;
 };
 
-dictionary PrepareSweepResponse {
-    u64 sweep_tx_weight;
-    u64 sweep_tx_fee_sat;
+dictionary PrepareRedeemOnchainFundsResponse {
+    u64 redeem_onchain_funds_tx_weight;
+    u64 redeem_onchain_funds_tx_fee_sat;
```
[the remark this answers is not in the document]
ubbabeck: Thanks, much more digestible.

libs/sdk-core/src/models.rs

tools/sdk-cli/src/command_handlers.rs
ubbabeck: Added

libs/sdk-core/src/models.rs

dangeross (approved):
LGTM! 🚀

PR #685, opened 2023-12-12
set group_id in send_pay

JssDWt:
The group_id was not set, which is the probable cause for this error:
image

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -835,7 +835,7 @@ impl NodeAPI for Greenlight {
                     payment_secret: Some(invoice.payment_secret.clone()),
                     partid: Some(part_id),
                     localinvreqid: None,
-                    groupid: None,
+                    groupid: Some(group_id),
                 })
                 .await?;
             part_id += 1;


roeierez:
@JssDWt  Looks good and good catch.
This fix is good anyway but according to the docs the group_id is not mandatory.Also it worked for me several times without the group_id.

Perhaps we are missing something else that may cause this error?
Looking at the code I would suspect that the `max_amount_per_channel` was empty and the loop that use send_pay had zero iterations and perhaps we are calling `wait_send_pay` without sending from the first place?

JssDWt:
@roeierez I think you're right.
This person said the button was pressed twice. So potentially the payment was made in the first attempt and the second attempt we run into this error. I might actually be caused by the group id if that is the case. send_pay returns success if you try to pay the same payment hash with the same amount again. I think we should do the same kind of lookup as we do for send_payment: AlreadyPaid.

dangeross:
> I think we should do the same kind of lookup as we do for send_payment: AlreadyPaid. Or perhaps a check whether a payment for this payment hash is already in-flight, since it is a hodl invoice being paid.

So throw an error if the payment is pending or complete?

JssDWt:
> So throw an error if the payment is pending or complete?

The user says both payments are taken from the balance, so I think I'm wrong in the above. Something else must be going on. That should mean there are two different invoices being paid. That should also match the pressing the button twice. Calling send_onchain twice, creating two invoices.

I don't see how that should lead to this error then, other than a race in Greenlight, or the groupid missing.

roeierez:
Looking at this again comparing to core lightning code I do think this will solve the issue we have.
There is clearly a mismatch in the group_id between the send_pay and wait_send_pay and according to core lightning code the payment won't be resolved:  I am looking again on the PR for that and I now think it will solve the issue.
I see in core lightning that if there is a mismatch between the group_id between adding a payment and fetching a payment it will not be fetched.
https://github.com/ElementsProject/lightning/blob/a48e2258c7b653bb03ccbbd8c7f4c99a7b514ac3/wallet/wallet.c#L3615

roeierez (approved):
LGTM

PR #691, opened 2023-12-15
Retry signer loop on startup, in case of connection error

ok300:
This PR offers an alternative solution to #596.

More specifically, the problem it solves is that `connect` called with known seed and credentials, but without network connectivity, can lead to the signer loop exiting a few minutes later. Thus, even if network connectivity is restored, the SDK instance won't be able to use the signer.

The fix proposed here is to add an outer loop in `start_signer`. This outer loop starts the signer loop and checks whether it exits due to a connectivity problem. If so, it restarts it.


ok300:
Trying a simpler version inside the signer loop.

PR #692, opened 2023-12-16
Bumping lightning and lightning-invoice versions

### LINKED ISSUE #689: BOLT12: bumping `lightning` version to version 0.0.118
The `parse_invoice` method currently does not support parsing BOLT12 invoices. To fix this, we need to create a separate method that allows us to do so. This feature is currently present in the latest version of the [lightning crate](https://docs.rs/lightning/0.0.118/lightning/), but requires version bumping, which in turn breaks some of the current functionalities.

hydra-yse:
Fixes #689


roeierez (approved):
LGTM

PR #705, opened 2024-01-02
[sdk-flutter] fetch node data on new block

ubbabeck:
We currently do not emit the node info when a new block arrives for the flutter sdk. This is useful for calculating the payment expiry for pending payments.

This pr also includes small fixes for the sdk and for the makefile.

### DIFF
--- a/libs/sdk-flutter/lib/breez_sdk.dart
+++ b/libs/sdk-flutter/lib/breez_sdk.dart
@@ -24,6 +24,9 @@ class BreezSDK {
   void initialize() {
     /// Listen to BreezEvent's(new block, invoice paid, synced)
     _lnToolkit.breezEventsStream().listen((event) async {
+      if (event is BreezEvent_NewBlock) {
+        await fetchNodeData();
+      }
       if (event is BreezEvent_InvoicePaid) {
         _invoicePaidStream.add(event.details);
         await fetchNodeData();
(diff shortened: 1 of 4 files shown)


ok300 (approved):

erdemyerebasmaz (approved):
LGTM

libs/sdk-flutter/lib/breez_sdk.dart
```diff
@@ -24,6 +24,9 @@ class BreezSDK {
   void initialize() {
     /// Listen to BreezEvent's(new block, invoice paid, synced)
     _lnToolkit.breezEventsStream().listen((event) async {
+      if (event is BreezEvent_NewBlock) {
+        await fetchNodeData();
```
roeierez, line 28: As far as I see the sync in the sdk is always called on new block so I am not sure we need that.
We already listen to sync and fetching the node data.

roeierez (changes requested):

libs/sdk-flutter/lib/breez_sdk.dart
ubbabeck: True, but when testing this the latest node date isn't received by the flutter side in c-breez even though if it's present in the sdk.

roeierez: We need to understand why.

ubbabeck: Disregard what I wrote, the new state is emitted when a new block arrives.

PR #706, opened 2024-01-04
Allow custom payment metadata inside SDK storage

### LINKED ISSUE #635: Feature request: Allow SDK users to store additional metadata inside SDK storage
It would be useful for SDK users to be able to persist additional metadata inside the SDK storage.

Some metadata might be **associated with payments** (e.g. fiat exchange rate, timezone, user category).
Also it would be nice if:
* the SDK allows to search by such data using `list_payments()`
* SDK encrypts the data (since such data might be private)
* the SDK sync the data between different instances

Probably a solution might be a field allowing to store a JSON-like structure or multiple fields storing key-value pairs.

Another usecase is to persist data **not associated with a particular payment** like users settings (e.g. fiat currency, language).

hydra-yse:
Closes #635

This PR aims to address modifying metadata strictly related to payments. A similar feature is already present in the persister:
https://github.com/breez/breez-sdk/blob/e6323f57acf0a5aad58974695b60ce07b0b77bfe/libs/sdk-core/src/persist/transactions.rs#L69-L74

Yet the method expects a pre-defined struct. This could be changed to accept any kind of JSON-seralized string, as Sqlite also [supports JSON operators and fields](https://www.sqlite.org/json1.html).  This would be the cleaner approach, rather than creating a separate method with equivalent functionality, but it has the following caveats:

- Currently the `insert_payment_external_info` method is also used internally for metadata such as lnurl-related fields, or errors. Merging the two into a single metadata column could lead to the possibility of developers overwriting these values, if they do not call the proper `update_payment_external_info` method (the two methods are split due to sqlite's technical limitations, if "insert" is called, the whole column will be replaced with the specified object). This is not really clean and prone to error for the end developer, and I am looking forwards to possible alternatives (possibly not involving an additional database read).

**Edit:** ~To address the first caveat, a SQL trigger can be created which automatically generates an empty JSON object whenever a payment is added to the `payments` table. Thus, one can simply call insert and update that row with the corresponding fields.~ This does not seem to be possible as the two tables belong to different databases. The best solution does indeed seem to perform an additional read to check whether the row exists already.


hydra-yse:
Based on your comments, I opted for keeping the struct and adding an `external_metadata` column instead. Let me know if the name sounds repetitive/needs change. I also added some utilities like update/retrieve functions that coerce directly into the desired struct.

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -180,6 +180,7 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
        ",
        "ALTER TABLE payments_external_info ADD COLUMN ln_address TEXT;",
        "ALTER TABLE payments_external_info ADD COLUMN lnurl_metadata TEXT;",
+       "ALTER TABLE payments_external_info ADD COLUMN external_metadata TEXT;",
```
dangeross, line 183: We cannot insert or edit migrations defined already in the `current_migrations` / `current_sync_migrations`, as these migrations have already been run on existing clients using the SDK. Any migration needed has to be added to the end of the respective list.

So this line needs to be moved to the end of the `current_sync_migrations` list of migrations and you can remove the other changes here.

dangeross (changes requested):

libs/sdk-core/src/persist/migrations.rs
hydra-yse: Understood, thanks Ross

libs/sdk-core/src/breez_services.rs
```diff
@@ -574,6 +574,27 @@ impl BreezServices {
         Ok(self.persister.get_payment_by_hash(&hash)?)
     }
 
+    /// Fetch the external metadata of a payment.
+    /// Useful for persisting external, platform-specific information,
+    /// unrelated to the SDK
+    pub async fn payment_external_metadata<T: serde::de::DeserializeOwned>(
```
roeierez, line 580: I think we need to add a "metadata" field to the payment structure (returned by list_payments). Otherwise the user would have to iterate over the list and fetch one by one.
If we do so the advantage is that we can remove this method as we already have `payment_by_hash`.

libs/sdk-core/src/breez_services.rs
```diff
@@ -574,6 +574,27 @@ impl BreezServices {
...
+        &self,
+        hash: String,
+    ) -> SdkResult<T> {
+        Ok(self.persister.get_payment_external_metadata(hash)?)
+    }
+
+    /// Set the external metadata of a payment as a valid JSON string
+    pub async fn set_payment_external_metadata(
```
roeierez, line 588: NIT: Perhaps we can shorten this to `set_payment_metadata` ?
Also missing the udl file implementation (In case you wanted to hold this until the approach is approved then ignore this comment).

roeierez (changes requested):
Looks nice and simple. Added come comment to simplify a bit more.

libs/sdk-core/src/persist/sync.rs
```diff
@@ -170,7 +170,8 @@ impl SqliteStorage {
           lnurl_metadata,
           lnurl_withdraw_endpoint,
           attempted_amount_msat,
-          attempted_error
+          attempted_error,
+          external_metadata
```
hydra-yse, line 174: As a side note, I was wondering whether or not it would would be useful to keep the `external_metadata` in the backup, as its usefulness could vary based on what the implementation is like. Given the small size, perhaps it is fine for now?

roeierez: I agree.  It only I think it is fine but also it seems mandatory since we want this metadata to sync accross apps/devices

libs/sdk-core/src/breez_services.rs
```diff
@@ -574,6 +574,13 @@ impl BreezServices {
         Ok(self.persister.get_payment_by_hash(&hash)?)
     }
 
+    /// Set the external metadata of a payment as a valid JSON string
+    pub async fn set_payment_metadata(&self, hash: String, metadata: String) -> SdkResult<()> {
```
roeierez, line 578: If we require the string to be json then we should validate it is a valid json otherwise we will end up with invalid json in the db.
Also I have noticed int the original issue it was mentioned that it would be nice to be able to search payments for a specific metadata value. This can be done using JSON1 sqlite extension (https://sqlite.org/json1.html). I think we can defer this specific search requirement to a following PR but we need to verify our current solution will be able to supports it.

libs/sdk-core/src/persist/sync.rs
```diff
@@ -170,7 +170,8 @@ impl SqliteStorage {
...
          FROM remote_sync.payments_external_info
          WHERE payment_id NOT IN (SELECT payment_id FROM sync.payments_external_info);",
```
roeierez, line 176: Syncing metadata is a bit different since we allow updating it in contrary to other columns that are created once and immutable.
This creates a problematic scenario when one device sets the metadata and after sync to a remote one subsequent updates won't be synced again (since the payment already exists).
One way to solve this is use the updated_at timestamp to sync the same way we do with the swap fees (`sync_swaps_fees_local`)

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
hydra-yse: You're right about verifying the JSON. I read that SQLite's `json_set` did checks before inserting, but I just tested out that it is not the case. Regarding searching for specific fields, I believe we can support it right away by exposing a method using `json_extract`.

dangeross (commented):
LGTM so far, just need to resolve the JSON validation and sync updating that Roei mentioned.

libs/sdk-core/src/models.rs
```diff
@@ -542,12 +542,32 @@ impl From<Network> for bitcoin::network::constants::Network {
     }
 }
 
+#[derive(PartialEq)]
+pub enum JsonFieldValue {
+    IsNull,
+    Bool { value: bool },
+    Number { value: u64 },
+    String { value: String },
+}
```
hydra-yse, line 551: Just wanted to leave two quick notes regarding the enum:
1. The choice to have the values specified in such a way (using a struct with `value` field) is due to UniFFI's limitations in regards to enums.
2. When `IsNull` is specified, due to limitations in SQLite's json methods when a field does not exist in an object, it will be treated as if it were null (e.g. in the following example, querying for "field" IsNull will return both objects): 
```js
  obj1 = {
     "field": null  
  }
  obj2 = {}
  ```
This is not a problem in most scenarios, especially when all the payment metadatas retain the same structure, but I believe it should be properly documented anyway.

File: libs/sdk-bindings/src/breez_sdk.udl [hunk 8]
```diff
@@ -193,10 +193,20 @@ dictionary CheckMessageResponse {
...
+    Number(u64 value);
+    String(string value);
+};
+
+[Enum]
+interface PaymentTypeFilter {
+    Sent();
+    Received();
+    ClosedChannel();
+    Metadata(string key, JsonFieldValue value);
```
roeierez, line 209: Perhaps we can simplify this and use strings only instead of JsonFieldValue ?

roeierez: Also we should think about this interface a bit more as the sent/received/closed channels are used as union in the query and metadata is used as filter.

hydra-yse: Addressed both points in the following commit. Would've been nice to use a `HashMap<String, String>` instead of a `Vec<PaymentMetadata>` for the filters, but apparently flutter_rust_bridge does not seem to support them (https://github.com/fzyzcjy/flutter_rust_bridge/issues/329).

libs/sdk-core/src/persist/sync.rs
```diff
@@ -162,17 +162,20 @@ impl SqliteStorage {
...
+          remote_sync.payments_external_info.lnurl_metadata,
+          remote_sync.payments_external_info.lnurl_withdraw_endpoint,
+          remote_sync.payments_external_info.attempted_amount_msat,
+          remote_sync.payments_external_info.attempted_error,
+          remote_sync.payments_external_info.external_metadata,
+          remote_sync.payments_external_info.updated_at
          FROM remote_sync.payments_external_info
-         WHERE payment_id NOT IN (SELECT payment_id FROM sync.payments_external_info);",
+         LEFT JOIN sync.payments_external_info ON sync.payments_external_info.payment_id = remote_sync.payments_external_info.payment_id
+         WHERE remote_sync.payments_external_info.updated_at > sync.payments_external_info.updated_at;",
```
roeierez, line 178: The problem with this approach is that if one device updates the metadata and wasn't synced other devices will override their data on next sync (since they would update the whole item)
Perhaps it is better to use a different table for metadata?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -199,6 +199,11 @@ enum PaymentTypeFilter {
     "ClosedChannel",
 };
 
+dictionary PaymentMetadata {
```
roeierez, line 202: NIT: rename to PaymentMetadataFilter

libs/sdk-core/src/breez_services.rs
roeierez, line 578: I still don't see json validation and even if we do so what happen if the user uses a nested json? Do we allow this? Won't it break the search?
How about passing here an array of metadata where every metadata has `key` and `value` fields where both are strings? We can validate for duplicate keys and it makes the search a lot simpler since we only use strings as values.
@dangeross what do you think?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
hydra-yse: I added the JSON validation inside the `update_payment_external_metadata` method, see [here](https://github.com/hydra-yse/breez-sdk/blob/febab11e462fe20d8f334eaf5e178711150bcc68/libs/sdk-core/src/persist/transactions.rs#L110-L111). Regarding passing a vector of `PaymentMetadata`, I agree, since we can then also reduce everything to one SQL query.

libs/sdk-bindings/src/breez_sdk.udl
hydra-yse: Given the other reply about passing a vector to `set_payment_metadata`, maybe we should keep this as PaymentMetadata and use it for both filtering and inserting. What do you think?

roeierez: The filter structure might change in the future ( we may for example add support for operators (contains, equals, etc...).
I think it is better to have a separate structure even though currently they look the same, they represent different things.

libs/sdk-core/src/breez_services.rs
dangeross: As @hydra-yse says, you can query within the nested JSON using `json_extract` with a key like `$.parent.child`, even with arrays `$.parent.arr[*].child`. I don't have a problem with the metadata being JSON, I think the question is what is the best approach to use to sync the metadata. A key/value store in a separate table is probably a better approach to keep the payment syncing clean. With a separate table when syncing we can control the approach to updating we want to take based on the metadata payment_hash/key, be it write once or write many (based on timestamp).

roeierez: I agree a separate table is the correct approach as the sync strategy is different.
My initial thinking was one row per payment with a json field.
When you referring to key/value store do you mean one row per metadata key so there is one to many relation between payment to metadata? This will complicate the interface as we will need to support also delete.

dangeross: Ah yeah, I was thinking one to many with one row per key/value but didn't consider the delete case. Then one row per payment_hash/json seems more appropriate.

ok300: With JSON, the question of validation remains. AFAIK duplicate keys are valid in JSON, but would be invalid in our scenario. So we'd have to explicitly check and deny that.

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -345,6 +391,28 @@ fn filter_to_where_clause(
...
+                        "CAST(json_extract(metadata, '$.{}') AS TEXT) = '{}'",
+                        key,
+                        value
+                    )
+                };
+
+                where_clause.push(query);
+            });
+        }
+    }
```
hydra-yse, line 414: Hope you like this approach. After our conversation, I first attempted implementing strings-only, yet realized that if we pass a JSON object for filtering as well (with the desired JSONPath-value fields) I can pretty easily coerce them into the right type! This should support the majority of cases, and is pretty intuitive for the end developer.

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -415,7 +415,14 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
...
        "ALTER TABLE channels ADD COLUMN alias_remote TEXT;",
-       "ALTER TABLE channels ADD COLUMN closing_txid TEXT;"
+       "ALTER TABLE channels ADD COLUMN closing_txid TEXT;",
+       "
+       CREATE TABLE IF NOT EXISTS payments_metadata (
+        payment_id TEXT NOT NULL PRIMARY KEY,
+        metadata TEXT,
+        updated_at TEXT DEFAULT CURRENT_TIMESTAMP
+       ) STRICT;
+       "
```
dangeross, line 425: I think the table only needs creating in the sync storage

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -101,6 +104,38 @@ impl SqliteStorage {
...
+    pub fn set_payment_external_metadata(
+        &self,
+        payment_hash: String,
+        new_metadata: String,
+    ) -> PersistResult<()> {
+        if new_metadata.len() > METADATA_MAX_SIZE_KB {
+            return Err(PersistError::Generic(
+                anyhow::anyhow!("Max metadata size ({}MB) has been exceeded", METADATA_MAX_SIZE_KB / 1000)
+            ));
+        }
```
dangeross, line 117: Could replace this with the ensure_sdk! macro
```suggestion
        ensure_sdk!(
            new_metadata.len() <= METADATA_MAX_SIZE_KB,
            PersistError::Generic(
                anyhow!("Max metadata size ({}MB) has been exceeded", METADATA_MAX_SIZE_KB / 1000)
            )
        );         
```

libs/sdk-core/src/persist/transactions.rs
dangeross: Sounds good. So something like?
```
{ "path.to.field": true }
```

libs/sdk-core/src/models.rs
```diff
@@ -640,6 +641,7 @@ pub struct PaymentExternalInfo {
 #[derive(Default)]
 pub struct ListPaymentsRequest {
     pub filters: Option<Vec<PaymentTypeFilter>>,
+    pub metadata_filters: Option<String>,
```
dangeross, line 644: I'm not sure if filter or filters is better here. I lean towards metadata_filter because its not a list.
```suggestion
    pub metadata_filter: Option<String>,
```

dangeross (changes requested):
Looking pretty good!

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -8,6 +8,9 @@ use rusqlite::{named_params, params, OptionalExtension};
 use std::collections::HashSet;
 
 use std::str::FromStr;
+use serde_json::{Value, Map};
+
+const METADATA_MAX_SIZE_KB: usize = 10000;
```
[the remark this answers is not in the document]
hydra-yse: No, it's actually kilobytes. Now that I think about it though, it probably is too much. Let's set a max of 10kb, and see how it goes, since the string encoding is UTF-8 and may be heavier for some locales (up to 4 bytes per character, meaning 2500 characters)

libs/sdk-core/src/persist/transactions.rs
hydra-yse: Didn't know about it. Thanks Ross!

libs/sdk-core/src/persist/transactions.rs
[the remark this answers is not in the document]
dangeross: 👍 I thought 10MB might be too much 😄

yaslama:
Imo, the best way is to use the -> operator and pass 2 strings to the filter: the first one is a json path, the second one a json representation
You can try for instance to run the following (command line sqlite3 or https://sqlite.org/fiddle/):
```
CREATE TABLE t(a json,b int);
INSERT INTO t(a,b) VALUES('{"a":2,"c":[4,5,{"f":7}]}',123),('[11,22,33,44]',456),(NULL,789),('"aa"',012);
INSERT INTO t(a,b) VALUES('{"a":true,"c":[4,5,{"f":7}]}',789);
INSERT INTO t(a,b) VALUES('{"a":1,"c":[4,5,{"f":7}]}',987);

select * from t where a->'$.c[2]'='{"f":7}';
select * from t where a->'$.c[2].f'='7';
select * from t where a->'$.a'='1';
select * from t where a->'$.a'='true';
```

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -345,6 +393,21 @@ fn filter_to_where_clause(
...
+        if let Ok(map) = serde_json::from_str::<Map<String, Value>>(&filters) {
+            map.iter().for_each(|(key, value)| {
+                where_clause.push(format!(
+                    "metadata->'$.{}' = '{}'",
+                    key,
+                    match value {
+                        Value::String(s) => format!(r#""{s}""#),
+                        _ => value.to_string(),
+                    }
+                ))
```
hydra-yse, line 406: Thanks @yaslama for the reply. This does indeed simplify even more our approach, although I noticed we still need to wrap JSON strings inside some double quotes in order to make this work as intended. Still, a huge improvement over checking for null and boolean fields, as now it can be done with only one query 👍

yaslama: > although I noticed we still need to wrap JSON strings inside some double quotes in order to make this work as intended

I don't understand this sentence. In order to be a valid json, a string has to be enclosed by double quotes. So if the second parameter (the value of the filter) is required to be a json, why it's a problem?

hydra-yse: Since we need to parse the filter string (a JSON object) and split it into multiple where-clauses for each field, we need to iterate over a `Map<String, Value>`. When a field has a `Value::String`, and we call the method `to_string()` (used in `format!` macro), the string gets printed out without the double quotes, which is a problem for the SQL query as it only works as follows: `SELECT * FROM metadata->'$.{fieldName}' = '"{value}"'`

roeierez: @hydra-yse I think What Akiba's suggestion is how to solve the issue of searching in non string values.
And by solving this it allows us to go back to the previously suggested interface for filtering instead of passing a json string as filter.

We can pass a list of `PaymentMetadataFilter` where the struct is defined like this:
```
struct PaymentMetadataFilter {
  search_path: String,
  search_value: String
}
```

Then it is trivial to compose the where clause using the arrow operator and we get two benefits:
1. The interface is clear (we pass a list of filters with documented fields)
2. It will search in all values (including booleans, numbers)

What do you think?

hydra-yse: Sure, they would both work well. On one hand I like the simplicity of being able to pass a JSON string without requiring any additional data structure, but I also see how setting the filters this way would make it very similar (and thus more familiar) to the `type_filters`. I'll continue as you suggested

libs/sdk-core/src/models.rs
```diff
@@ -550,6 +550,15 @@ pub enum PaymentTypeFilter {
     ClosedChannel,
 }
 
+/// A metadata filter which can be applied when retrieving the transaction list
+pub struct PaymentMetadataFilter {
+    /// Specifies which field to apply the filter on, using the JSON path format
+    pub search_path: String,
+    /// Specifies which JSON value to filter for.
+    /// As such, strings must be wrappe with quotes ("") in order to be properly filtered
```
roeierez, line 558: I don't understand the comment here. Should the user add quotes as part of the value here?

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -1,14 +1,17 @@
...
+use crate::{ensure_sdk, models::*};
 use rusqlite::types::{FromSql, FromSqlError, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
 use rusqlite::Row;
 use rusqlite::{named_params, params, OptionalExtension};
 use std::collections::HashSet;
 
+use serde_json::{Map, Value};
 use std::str::FromStr;
 
+const METADATA_MAX_LEN: usize = 10000;
```
roeierez, line 13: I would start with 1k, we can always expand later.

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -345,6 +393,20 @@ fn filter_to_where_clause(
...
     }
 
+    if let Some(filters) = metadata_filters {
+        filters.iter().for_each(
+            |PaymentMetadataFilter {
+                 search_path,
+                 search_value,
+             }| {
+                where_clause.push(format!(
+                    "metadata->'$.{}' = '{}'",
```
roeierez, line 403: Since we are formatting here a string using external input we should validate the input for security reasons.
Best would be to use a prepared statement and pass these values as parameters (at least the value).
We can validate the json path using https://docs.rs/serde_json_path/latest/serde_json_path/struct.JsonPath.html

roeierez (changes requested):
Looks good. I think we are very close.

libs/sdk-core/src/models.rs
hydra-yse: Yes, if the value inside the JSON object is a JSON string, the developer must add double quotes in order to match for it. The `->` operator simply creates a one-to-one match, so "string" can only be matched by the literal "string".

roeierez: I see, I think it will create a confusion. Users will not be aware of that, they will just pass a string...
@yaslama any idea on how to make the experience here more straight forward?

roeierez: @hydra-yse @yaslama perhaps we can trim the quotes before the matching?

`trim(metadata->'$.{}', '"') = '{}'`

libs/sdk-core/src/persist/transactions.rs
hydra-yse: I agree using a prepared statement, because that is enough to prevent SQL injections, but is parsing the JSON path necessary? The worst that can happen if someone specifies a bad path would just be a bad filter that does not catch anything, which is not an issue on our side.

What do you think?

roeierez: I think If we use a prepared statement then we don't need to parse the json path.

libs/sdk-core/src/models.rs
roeierez: @hydra-yse  Now thinking again my suggestion will make a problem in case the user pass json (nested).
Ok, I am fine with the quotes as we document the value should be a valid json (where strings should be quoted)

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -394,17 +404,9 @@ fn filter_to_where_clause(
...
-                 search_value,
-             }| {
-                where_clause.push(format!(
-                    "metadata->'$.{}' = '{}'",
-                    search_path, search_value
-                ));
-            },
-        );
+        filters.iter().for_each(|_| {
+            where_clause.push("metadata->'$.?'='?'".to_string());
```
dangeross, line 408: I wonder if these single quotes are the issue:
```suggestion
            where_clause.push("metadata->?=?".to_string());
```
then when building the params:
```
metadata_filters.iter().for_each(|PaymentMetadataFilter { search_path, search_value }| {
    params.push(format!("$.{}", search_path));
    params.push(search_value.clone());
})
```

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -196,8 +238,22 @@ impl SqliteStorage {
...
         )?;
 
+        let mut params: HashMap::<String, String> = HashMap::new();
+
+        if let Some(metadata_filters) = &req.metadata_filters {
+            metadata_filters.iter().enumerate().for_each(|(i, PaymentMetadataFilter { search_path, search_value })| {
+                params.insert(format!(":search_path_{i}"), format!("'$.{search_path}'"));
+                params.insert(format!(":search_value_{i}"), format!("'{search_value}'"));
+            })
+        }
```
hydra-yse, line 248: Based on the solution proposed by @roeierez and @dangeross , I decided to opt for their approach. This still does not work, but at the same time the parser does not complain of any parsing errors. I am thinking the issue is now related to the JSON Path, as maybe it is not evaluated as such due to the fact that the `$` is part of the prepared statement, perhaps? I have verified the values get indeed passed to `query_map` as intended, therefore I imagine it to be a SQL-related issue.

roeierez: The test you added did pass using the commit I posted: https://github.com/breez/breez-sdk/commit/7a324252a22e2954ad93cb83e55890b1c836cdbc

hydra-yse: @roeierez Thanks for the swift reply, turns out the quotes were not needed at all. Made it work with the HashMap, pushing rn!

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -199,6 +199,11 @@ enum PaymentTypeFilter {
     "ClosedChannel",
 };
 
+dictionary PaymentMetadataFilter {
+    string search_path;
```
roeierez, line 203: NIT: perhaps json_path to be more explicit?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -199,6 +199,11 @@ enum PaymentTypeFilter {
...
+    string search_value;
```
roeierez, line 204: Same here, `json_value` ?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -199,6 +199,11 @@ enum PaymentTypeFilter {
     "ClosedChannel",
 };
 
+dictionary PaymentMetadataFilter {
```
roeierez, line 202: NIT: rename to `MetadataFilter` ? This looks generic enough to apply (if needed) later on other entities.

roeierez (approved):
LGTM! I added several NIT suggestions about renaming fields, other than that good to go IMO.

dangeross (approved):
Looks good @hydra-yse, just one small typo

PR #707, opened 2024-01-04
only iterate over pending payments for update_payment_expirations

ubbabeck:
minor enhancement for #674. We only need to go through payments that have status pending.


roeierez (approved):
LGTM

PR #708, opened 2024-01-04
Chain service: Add support for an optional fallback mempool.space URL

ok300:
This PR adds an optional fallback URL for a 2nd mempool.space instance.

It is called automatically if the call to the main mempool instance fails. If the fallback call also fails, the error is propagated to the caller.

### DIFF
--- a/libs/sdk-core/src/chain.rs
+++ b/libs/sdk-core/src/chain.rs
@@ -1,4 +1,5 @@
-use crate::input_parser::get_parse_and_log_response;
+use crate::input_parser::{get_parse_and_log_response, post_and_log_response};
+
 use anyhow::{anyhow, Result};
 use bitcoin::hashes::hex::FromHex;
 use bitcoin::{OutPoint, Txid};
@@ -124,6 +125,7 @@ pub(crate) fn get_utxos(address: String, transactions: Vec<OnchainTx>) -> Result
 #[derive(Clone)]
 pub(crate) struct MempoolSpace {
     pub(crate) base_url: String,
+    pub(crate) fallback_base_url: Option<String>,
 }
 
 /// Wrapper containing the result of the recommended fees query, in sat/vByte, based on mempool.space data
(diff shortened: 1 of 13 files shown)


dangeross:
Should we seek permission to use Emzy`s mempool.space instance as the fallback? Maybe a better alternative is to set the default to None and allow devs to set the fallback as they see fit

dangeross:
Code LGTM, I just question if the `mempoolspace_fallback_url` should be optional in the Config and that it should be set by the developer using the SDK. If mempool.space is down, there is then network load going another single point of failure.

ok300:
The fallback is more useful if the default has a value. And Emzy's instance is public, so it seems reasonable we could start using it. But I'm flexible on both points.

As background: in my tests, I've seen short periods of HTTP 404 or 503 errors every now and then on the main mempool instance. It usually lasts only a few seconds, then it works again. This PR is an enhancement proposal for to better handle these kinds of cases.

libs/sdk-core/src/chain.rs
```diff
@@ -124,6 +125,7 @@ pub(crate) fn get_utxos(address: String, transactions: Vec<OnchainTx>) -> Result
 #[derive(Clone)]
 pub(crate) struct MempoolSpace {
     pub(crate) base_url: String,
+    pub(crate) fallback_base_url: Option<String>,
```
dangeross, line 128: Is there any point `fallback_base_url` being optional if the Config `mempoolspace_fallback_url` is not optional?

yaslama:
If we want to able to control from Breez the default url, the best way way is to add a rpc function (into beez server) which returns a list of default urls. If no url is configured, the sdk calls this rpc and tries the first one and if it fails the second one, etc..

libs/sdk-core/src/chain.rs
ok300: Good point, fixed.

roeierez:
@ok300 if we decided to take the values from the server, can we close this one?

ok300:
Yes. I added an issue in the LSPD repo instead.

yaslama:
> Yes. I added an issue in the LSPD repo instead.

Why in lspd? It's in breez server
https://github.com/breez/breez-sdk/pull/708#issuecomment-1880068799

ok300:
I thought the idea is who runs an LSP may want to control the default mempool instances for their users.

I moved it to the breez server repo: https://github.com/breez/server/issues/41

PR #709, opened 2024-01-05
Improve sync payments

### LINKED ISSUE #704: Improve sync payments
Recently we have changed the logic post payment to add the payment to the db and then trigger sync (it was vice versa previously).
This may be buggy on some rare times as the sync starts always from the last payment time in the db.
In theory a received payment may may be received while there are sill received payments that were not synced before that payment and then preventing from the next sync to fetch these prior payments.
The solution maybe as simple as saving the last sync time separately in the db and update it on every sync.

dangeross:
Synchronises payments using the cached `last_payment_timestamp` from the previous sync.

~~**Open question:** Maybe its better to cache the `last_payment_timestamp` as the `last_sync_time` at the end on the sync. Then we know for sure that that is the last payment time to sync from in the next sync and payments can be safely added to the table outside of the sync without messing up the last_sync_time. This also avoids clock errors between the SDK and GL when syncing based on the local clock.~~

Fixes #704

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -879,11 +879,7 @@ impl BreezServices {
         self.connect_lsp_peer(node_pubkey).await?;
 
         // First query the changes since last sync time.
-        let since_timestamp = self
-            .persister
-            .get_last_sync_time()?
-            .or_else(|| self.persister.last_payment_timestamp().ok())
-            .unwrap_or(0);
+        let since_timestamp = self.persister.get_last_sync_time()?.unwrap_or(0);
         let new_data = &self
             .node_api
             .pull_changed(since_timestamp, balance_changed)
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -888,10 +888,6 @@ impl BreezServices {
             .node_api
             .pull_changed(since_timestamp, balance_changed)
             .await?;
-        let last_sync_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_secs();
 
         debug!(
             "pull changed time={:?} {:?}",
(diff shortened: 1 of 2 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -879,11 +879,19 @@ impl BreezServices {
...
+        let since_timestamp = self
+            .persister
+            .get_last_sync_time()?
+            .or_else(|| self.persister.last_payment_timestamp().ok())
+            .unwrap_or(0);
         let new_data = &self
             .node_api
             .pull_changed(since_timestamp, balance_changed)
             .await?;
+        let last_sync_time = SystemTime::now()
```
roeierez, line 891: I think it is better to use the last_sync_time as the newest payment, received or sent (but not pending), time.
I think the time should be taken from gl rather the local time because:
1. It avoids any clock sync between local time and gl (such as if a payment would be later added exactly at that gap).
2. Eventually we want to have an endpoint in gl that filters based on timestamp.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
dangeross: 👍  these were my concerns too

libs/sdk-core/src/breez_services.rs
```diff
@@ -879,7 +879,11 @@ impl BreezServices {
         self.connect_lsp_peer(node_pubkey).await?;
 
         // First query the changes since last sync time.
-        let since_timestamp = self.persister.last_payment_timestamp().unwrap_or(0);
+        let since_timestamp = self
+            .persister
+            .get_last_sync_time()?
+            .or_else(|| self.persister.last_payment_timestamp().ok())
```
roeierez, line 885: Do we need the fallback to `last_payment_timestamp` ?

roeierez (changes requested):
One comment, otherwise LGTM.

libs/sdk-core/src/breez_services.rs
dangeross: The fallback is for the first sync that happens before the `last_sync_time` is set. We could alternatively set it when building the BreezService instance. What do you think?

roeierez: I see there is already fallback to zero, that's why I was asking.

dangeross: Yeah, it could be the first run/node without payments, so also needs a fallback if there is no payment also. Any recommendations?

roeierez: How about leaving the zero as single fallback? Worst case on migration it will sync all payments one time.

dangeross: I'm fine with that. First sync after update will definely be a full payment sync.

roeierez (approved):
LGTM

Issue #537, opened 2023-10-19
Document fields of `SwapInfo`


danielgranhao:
Without reviewing the source code, it's unclear how to use some of the fields of `SwapInfo` or how the dynamic ones behave (when they are filled in, when they change, etc).

Example: when using `list_refundables()`, I was unsure if the amount that's available to be refunded was given simply by `swap_info.confirmed_sats` or if I had to add the unconfirmed ones or even subtract `paid_sats`. The documentation ideally would make it clear that it's simply the value in `swap_info.confirmed_sats`.

PR #712, opened 2024-01-08
document fields of SwapInfo

ubbabeck:
documents #537

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -1169,32 +1169,52 @@ impl TryFrom<i32> for SwapStatus {
 /// saved to the persistent storage.
 ///
 /// The SwapInfo has a status which changes accordingly, documented in [SwapStatus].
+///
+
 #[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize)]
 pub struct SwapInfo {
-    //static immutable data
+    /// Bitcoin address for this swap. Sats sent to this address will be swapped.
     pub bitcoin_address: String,
+    /// Relative time lock start, received from [SwapperAPI::create_swap].
     pub created_at: i64,
+    /// Relative time lock for the timeout for the script to be redeemed before swap fails.
     pub lock_height: i64,
+    /// sha256 hash of preimage to used in the claim sript.
     pub payment_hash: Vec<u8>,
+    /// Secret to claim the swap.
     pub preimage: Vec<u8>,
+    /// Secret claim key for the bitcoin address.
     pub private_key: Vec<u8>,
+    /// Public key in binary format of the private claim private key.
     pub public_key: Vec<u8>,
+    /// The public key in binary format from the swapping service. Received from [SwapperAPI::create_swap].
     pub swapper_public_key: Vec<u8>,
+    /// The lockingsscript for the generated bitcoin address. Received from [SwapperAPI::create_swap].
     pub script: Vec<u8>,
 
-    // dynamic data
+    /// bolt11 invoice to claim the sent funds.
     pub bolt11: Option<String>,
+    /// Amount of sats payed to the claim address.
     pub paid_sats: u64,
+    /// Confirmed onchain sats to be claim with an bolt11 invoice or refunded if swap fails.
     pub confirmed_sats: u64,
+    /// Unconfirmed sats waiting to be confirmed onchain.
     pub unconfirmed_sats: u64,
+    /// Shows the current status of the swap, either `Initial` or `Expired`.
     pub status: SwapStatus,
+    /// Transaction IDs for failed swap attempts.
     pub refund_tx_ids: Vec<String>,
+    /// Refund transaction IDs for ongoing swap awaiting confirmation.
     pub unconfirmed_tx_ids: Vec<String>,
+    /// Transaction IDs that have been confirmed on-chain.
     pub confirmed_tx_ids: Vec<String>,
+    /// The minimum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].   
     pub min_allowed_deposit: i64,
+    /// The maximum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
     pub max_allowed_deposit: i64,
+    /// Error reason for when swap fails.
     pub last_redeem_error: Option<String>,
-    /// The dynamic fees that may have been used when the swap was created
+    /// The dynamic fees which is present if a channel opening is needed.
     pub channel_opening_fees: Option<OpeningFeeParams>,
 }
 
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -1214,7 +1214,7 @@ pub struct SwapInfo {
     pub max_allowed_deposit: i64,
     /// Error reason for when swap fails.
     pub last_redeem_error: Option<String>,
-    /// The dynamic fees which is present if a channel opening is needed.
+    /// The dynamic fees which is set if a channel opening is needed.
     pub channel_opening_fees: Option<OpeningFeeParams>,
 }
 
--- a/libs/sdk-flutter/lib/bridge_generated.dart
+++ b/libs/sdk-flutter/lib/bridge_generated.dart
@@ -1672,29 +1672,69 @@ sealed class SuccessActionProcessed with _$SuccessActionProcessed {
 /// saved to the persistent storage.
 ///
 /// The SwapInfo has a status which changes accordingly, documented in [SwapStatus].
+///
 class SwapInfo {
+  /// Bitcoin address for this swap. Sats sent to this address will be swapped.
   final String bitcoinAddress;
+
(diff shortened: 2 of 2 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,56 @@ impl TryFrom<i32> for SwapStatus {
...
     pub paid_sats: u64,
+    /// Confirmed onchain sats to be claim with an bolt11 invoice or refunded if swap fails.
     pub confirmed_sats: u64,
+    /// Unconfirmed sats waiting to be confirmed onchain.
     pub unconfirmed_sats: u64,
+    /// Shows the current status of the swap, either `Initial` or `Expired`.
     pub status: SwapStatus,
+    /// Transaction ID for failed swap attempts.
     pub refund_tx_ids: Vec<String>,
+    /// Transaction ID for ongoing swap awaiting confirmation.
```
ok300, line 1211:
```suggestion
    /// Transaction IDs for ongoing swap awaiting confirmation.
```

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,56 @@ impl TryFrom<i32> for SwapStatus {
...
     pub unconfirmed_tx_ids: Vec<String>,
+    /// Transaction ID that have been confirmed on-chain.
```
ok300, line 1213:
```suggestion
    /// Transaction IDs that have been confirmed on-chain.
```

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,56 @@ impl TryFrom<i32> for SwapStatus {
...
     pub confirmed_tx_ids: Vec<String>,
+    /// The minimum amount the one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
```
ok300, line 1215:
```suggestion
    /// The minimum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
```

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,56 @@ impl TryFrom<i32> for SwapStatus {
...
     pub min_allowed_deposit: i64,
+    /// The maximum amount the one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
```
ok300, line 1217:
```suggestion
    /// The maximum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
```

ok300 (commented):
A few NITs, but generally looks good!

libs/sdk-core/src/models.rs
```diff
@@ -1170,51 +1170,48 @@ impl TryFrom<i32> for SwapStatus {
...
+    /// Amount of sats payed to the claim address. Dynamic field
     pub paid_sats: u64,
-    /// Confirmed onchain sats to be claim with an bolt11 invoice or refunded if swap fails.
+    /// Confirmed onchain sats to be claim with an bolt11 invoice or refunded if swap fails. Dynamic field.
     pub confirmed_sats: u64,
-    /// Unconfirmed sats waiting to be confirmed onchain.
+    /// Unconfirmed sats waiting to be confirmed onchain. Dynamic field.
     pub unconfirmed_sats: u64,
-    /// Shows the current status of the swap, either `Initial` or `Expired`.
+    /// Shows the current status of the swap, either `Initial` or `Expired`. Dynamic field. Dynamic field.
```
ok300, line 1204:
```suggestion
    /// Shows the current status of the swap, either `Initial` or `Expired`. Dynamic field.
```

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,56 @@ impl TryFrom<i32> for SwapStatus {
 /// saved to the persistent storage.
 ///
 /// The SwapInfo has a status which changes accordingly, documented in [SwapStatus].
+///
+/// Static imutable fields are [`bitcoin_address`](SwapInfo::bitcoin_address), [`created_at`](SwapInfo::created_at), [`lock_height`](SwapInfo::lock_height), [`payment_hash`](SwapInfo::payment_hash), [`preimage`](SwapInfo::preimage), [`private_key`](SwapInfo::private_key), [`public_key`](SwapInfo::public_key), [`swapper_public_key`](SwapInfo::swapper_public_key), [`script`](SwapInfo::script), [`min_allowed_deposit`](SwapInfo::min_allowed_deposit) and [`max_allowed_deposit`](SwapInfo::max_allowed_deposit).
```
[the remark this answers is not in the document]
ubbabeck: Good point, added the short description and whenever the fields are dynamic or static

libs/sdk-core/src/models.rs
ubbabeck: Whoops, should have seen that. Thanks.

ok300 (approved):
LGTM

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,53 @@ impl TryFrom<i32> for SwapStatus {
 /// saved to the persistent storage.
 ///
 /// The SwapInfo has a status which changes accordingly, documented in [SwapStatus].
+///
+/// Static fields are those which don't change for the lifetime of a swap, whereas dynamic fields may change.
+
 #[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize)]
 pub struct SwapInfo {
-    //static immutable data
+    /// Bitcoin address for this swap. Sats sent to this address will be swapped. Static field.
```
roeierez, line 1177: Static field doesn't seem the right term. Perhaps "immutable field" more fits here although I don't feel we should add that info.

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,53 @@ impl TryFrom<i32> for SwapStatus {
...
     pub public_key: Vec<u8>,
+    /// The public key in binary format from the swapping service. Received from [SwapperAPI::create_swap]. Static field.
     pub swapper_public_key: Vec<u8>,
+    /// The lockingsscript for the generated bitcoin address. Received from [SwapperAPI::create_swap]. Static field.
     pub script: Vec<u8>,
 
-    // dynamic data
+    /// bolt11 invoice to claim the sent funds. Dynamic field.
     pub bolt11: Option<String>,
+    /// Amount of sats payed to the claim address. Dynamic field
```
roeierez, line 1198: Same here for Dynamic field. I would remove this, too low level and implementation details.

roeierez (changes requested):

libs/sdk-core/src/models.rs
ubbabeck: Good point, might be needless to add this info. Removed

libs/sdk-core/src/models.rs

libs/sdk-core/src/models.rs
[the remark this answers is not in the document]
ubbabeck: Removed not longer relevant ref (https://github.com/breez/breez-sdk/pull/712#discussion_r1449926815)

libs/sdk-core/src/models.rs
```diff
@@ -1169,32 +1169,52 @@ impl TryFrom<i32> for SwapStatus {
...
+    /// Transaction IDs that have been confirmed on-chain.
     pub confirmed_tx_ids: Vec<String>,
+    /// The minimum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].   
     pub min_allowed_deposit: i64,
+    /// The maximum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
     pub max_allowed_deposit: i64,
+    /// Error reason for when swap fails.
     pub last_redeem_error: Option<String>,
-    /// The dynamic fees that may have been used when the swap was created
+    /// The dynamic fees which is present if a channel opening is needed.
```
roeierez, line 1217: "which is present" => "Which is set" ?

libs/sdk-flutter/lib/bridge_generated.dart
```diff
@@ -1672,29 +1672,69 @@ sealed class SuccessActionProcessed with _$SuccessActionProcessed {
...
   final int minAllowedDeposit;
+
+  /// The maximum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
   final int maxAllowedDeposit;
+
+  /// Error reason for when swap fails.
   final String? lastRedeemError;
 
-  /// The dynamic fees that may have been used when the swap was created
+  /// The dynamic fees which is present if a channel opening is needed.
```
roeierez, line 1737: "Which is set" ?

roeierez (changes requested):
Small two syntax changes. Other than that Looks good.

libs/sdk-flutter/lib/bridge_generated.dart
ubbabeck: Added.

libs/sdk-core/src/models.rs
ubbabeck: Added

Issue #667, opened 2023-12-04
Feature request: expose unconfirmed onchain balance


danielgranhao:
It would be helpful to get this to let the user know that, eventually, he will be able to recover some funds using `sweep()`.

PR #714, opened 2024-01-09
new `pending_balance` field to NodeState

ubbabeck:
implements #667

Calculates the pending balance and adds it to the node state.


ubbabeck:
When testing this by closing a channel the node was most likely out off sync as it did not recognise the funds  either pending nor confimred_onchain_balance before after 2 block confirmations.

ubbabeck:
> You can check if a node is out of sync with `node_info`. The [result](https://breez.github.io/breez-sdk/breez_sdk_core/struct.NodeState.html) contains the node's block height. If that's lagging, the node is indeed out of sync.

Thank I looked through the logs now and can confirm that the node was laging ~50 block behind the tip for when the transaction was first seen in the mempool.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -441,6 +438,20 @@ impl Greenlight {
         Ok(on_chain_balance)
     }
 
+    async fn pending_onchain_balance(&self, funds: cln::ListfundsResponse) -> Result<u64> {
+        let pending_onchain_balance = funds.outputs.iter().fold(0, |a, b| {
+            if b.status != 0 || !b.reserved {
```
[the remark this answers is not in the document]
roeierez: Can you explain the logic here? It looks like we consider utxos that are confirmed or not reserved as pending balance.

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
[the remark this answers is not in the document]
ubbabeck: This logic was flawed. Since the channel is not closed yet then there is not much help in using the output of list funds. Instead we need to go thought all of the channels that are closed pending and sum up `our_amount_msat` for each channel.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -441,6 +438,32 @@ impl Greenlight {
         Ok(on_chain_balance)
     }
 
+    async fn pending_onchain_balance(&self, funds: cln::ListfundsResponse) -> Result<u64> {
+        let pending_onchain_balance = funds.channels.iter().fold(0, |a, b| match b.state() {
```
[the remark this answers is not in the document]
ubbabeck: I will add this again, do we want to add the outputs where it's not confirmed or reserved ?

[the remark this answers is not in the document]
ubbabeck: When looking at the behaviour of cln the outputs list are employed when channel opening is taking place. On the other hand when channel closes are happening it the `channels` field which are populated. The match statement covers also the force close attempt, the `AwaitingUnilateral` is triggered on force close. After this is done the the funds will be accessible for the peer how didn't initiate the force close. Which means it will be showing up in the listfunds. 

So it is under my impression that the log above also includes the force closed channels and it's sweep.



Here are the state changes for the peer who had their channel force closed on them .

```
"state_changes": [
            {
               "timestamp": "2024-01-16T10:39:03.829Z",
               "old_state": "CHANNELD_AWAITING_LOCKIN",
               "new_state": "CHANNELD_NORMAL",
               "cause": "user",
               "message": "Lockin complete"
            },
            {
               "timestamp": "2024-01-16T10:41:32.330Z",
               "old_state": "CHANNELD_NORMAL",
               "new_state": "AWAITING_UNILATERAL",
               "cause": "protocol",
               "message": "channeld: received ERROR error channel 607e907fccd85848a2e0a8cf25e010bf67771879afc064b099abd4cb7bab39f0: Forcibly closed by `close` command timeout"
            },
            {
               "timestamp": "2024-01-16T10:44:01.368Z",
               "old_state": "AWAITING_UNILATERAL",
               "new_state": "FUNDING_SPEND_SEEN",
               "cause": "onchain",
               "message": "Onchain funding spend"
            },
            {
               "timestamp": "2024-01-16T10:44:01.376Z",
               "old_state": "FUNDING_SPEND_SEEN",
               "new_state": "ONCHAIN",
               "cause": "onchain",
               "message": "Onchain init reply"
            }
         ],
"status": [
            "CHANNELD_NORMAL:Received error channel 607e907fccd85848a2e0a8cf25e010bf67771879afc064b099abd4cb7bab39f0: Forcibly closed by `close` command timeout",
            "ONCHAIN:Tracking their unilateral close",
            "ONCHAIN:All outputs resolved: waiting 99 more blocks before forgetting channel"
         ],
```

The peer closing the channel
```
"state_changes": [
            {
               "timestamp": "2024-01-16T10:39:03.827Z",
               "old_state": "CHANNELD_AWAITING_LOCKIN",
               "new_state": "CHANNELD_NORMAL",
               "cause": "remote",
               "message": "Lockin complete"
            },
            {
               "timestamp": "2024-01-16T10:40:55.058Z",
               "old_state": "CHANNELD_NORMAL",
               "new_state": "CHANNELD_SHUTTING_DOWN",
               "cause": "user",
               "message": "User or plugin invoked close command"
            },
            {
               "timestamp": "2024-01-16T10:40:59.887Z",
               "old_state": "CHANNELD_SHUTTING_DOWN",
               "new_state": "AWAITING_UNILATERAL",
               "cause": "user",
               "message": "Forcibly closed by `close` command timeout"
            },
            {
               "timestamp": "2024-01-16T10:44:03.927Z",
               "old_state": "AWAITING_UNILATERAL",
               "new_state": "FUNDING_SPEND_SEEN",
               "cause": "user",
               "message": "Onchain funding spend"
            },
            {
               "timestamp": "2024-01-16T10:44:03.931Z",
               "old_state": "FUNDING_SPEND_SEEN",
               "new_state": "ONCHAIN",
               "cause": "user",
               "message": "Onchain init reply"
            }
         ],
         "status": [
            "ONCHAIN:Tracking our own unilateral close",
            "ONCHAIN:3 outputs unresolved: in 5 blocks will spend DELAYED_OUTPUT_TO_US (1b916ba0213d85138af0d67760d05aa26591975e2fc6d230ba7f21fb2e944fbb:0) using OUR_DELAYED_RETURN_TO_WALLET"
         ],
```

ubbabeck:
There is a case for the node who initiates the unilateral close which isn't handled yet. 

During the delayed to self the funds shows as status 'onchain' in channels, but not in outputs until before the timelock has passed. In the current implementation this isn't added to the pending onchain funds. 

I think this can be done by using listpeerchannels

ubbabeck:
> There is a case for the node who initiates the unilateral close which isn't handled yet.
> 
> During the delayed to self the funds shows as status 'onchain' in channels, but not in outputs until before the timelock has passed. In the current implementation this isn't added to the pending onchain funds.
> 
> I think this can be done by using listpeerchannels

When investigating how we close channels to our lsp I see that we do not support unilateral closes. Should we still handle this case as I user in theory can manually close a channel using the dev-commands? 
```
let response = client
                    .close(cln::CloseRequest {
                        id: hex::encode(chan_id),
                        unilateraltimeout: None,
                        destination: None,
                        fee_negotiation_step: None,
                        wrong_funding: None,
                        force_lease_closed: None,
                        feerange: vec![],
                    })
```

roeierez:
> When investigating how we close channels to our lsp I see that we do not support unilateral closes. Should we still handle this case as I user in theory can manually close a channel using the dev-commands?

What do you mean by not supporting? the lSP can close the channel unilaterally without negotiating with the user node. Also vice versa.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -710,7 +723,8 @@ impl NodeAPI for Greenlight {
 
         // calculate onchain balance
         let onchain_balance = self.on_chain_balance(funds.clone()).await?;
-        let utxos = self.utxos(funds).await?;
+        let pending_onchain_balance = self.pending_onchain_balance(funds.clone()).await?;
```
JssDWt, line 726: I think we should be able to borrow `funds`, so it doesn't need to be cloned.
```suggestion
        let pending_onchain_balance = self.pending_onchain_balance(&funds).await?;
```

JssDWt (commented):
small nit

libs/sdk-core/src/greenlight/node_api.rs
ubbabeck: Done, should I add this for the `onchain_balance` as well?

ubbabeck:
> > When investigating how we close channels to our lsp I see that we do not support unilateral closes. Should we still handle this case as I user in theory can manually close a channel using the dev-commands?
> 
> What do you mean by not supporting? the lSP can close the channel unilaterally without negotiating with the user node. Also vice versa.

Understand, I added functionality for this in a28c77e7fb635a0906d829d0308ed1fb5854234c and also this enabled us to use the `min_to_us_msat` field so that users do not get disappointed if their onchain  funds are lower than expected.

libs/sdk-core/src/greenlight/node_api.rs
ubbabeck: Added it for `onchain_balance` too lmk if I should drop this commit.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -441,6 +438,28 @@ impl Greenlight {
...
+            | FundingSpendSeen => a + b.min_to_us_msat.clone().unwrap_or_default().msat,
+            Onchain => {
+                if b.closer() == cln::ChannelSide::Local
+                    && b.status[b.status.len() - 1].contains("DELAYED_OUTPUT_TO_US")
+                {
+                    a + b.min_to_us_msat.clone().unwrap_or_default().msat
+                } else {
+                    a
+                }
+            }
```
JssDWt, line 456: Can you explain why 'onchain' gets a specific case?
1) 'Onchain' means this was a unilateral close attempt by the peer. In that case, you should not get a delayed output? If you initiate the unilateral close yourself (AwaitingUnilateral), you get a delayed output. Or does this have to do when htlcs are involved?

2) Why is the balance of the delayed outputs added to the pending balance, but not the balance of the non-delayed outputs?

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: Adding to onchain_balance too is great 👍

libs/sdk-core/src/greenlight/node_api.rs
ubbabeck: 1. When the user unilateral closes the channel it will as you write get a status as `AwaitingUnilateral` first, but when the closing transaction is confirmed onchain the channel gets a status as `Onchain`. Though for the peer causing they'll have to wait X amount of blocks before they can spend those funds, which is reflected in the status.

2. Adding the balance of the delayed outputs to the pending balance since they cannot be spent. Also if earlier states shows a pending balance it makes sense to add them here too as the funds cannot be spent due to the delayed outputs. Adding the balance of the non delayed outputs does not make sense to add as a pending balance since they can be spent.

JssDWt: OK thanks, makes sense.
Maybe add a comment in the code there to explain what it's doing, because it's hard to figure out frmo cln documentation why we're doing what we're doing here.

ubbabeck: Good point, added a comment

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -441,6 +438,33 @@ impl Greenlight {
...
+            | FundingSpendSeen => a + b.min_to_us_msat.clone().unwrap_or_default().msat,
+
+            // When we  unilaterally close the channel it will get status as `AwaitingUnilateral`
+            // first, but when the closing transaction is confirmed onchain the funds receive status
+            // as `Onchain`. Though if we closed the channel we'll have to wait for the timelock to
+            // pass before the funds can be spent.
+            Onchain => {
+                if b.closer() == cln::ChannelSide::Local
+                    && b.status[b.status.len() - 1].contains("DELAYED_OUTPUT_TO_US")
+                {
```
dangeross, line 456: What do you think about using [last()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.last) here?
```suggestion
                if b.closer() == cln::ChannelSide::Local
                    && b.status.last().is_some_and(|status| status.contains("DELAYED_OUTPUT_TO_US")
                {
```

ubbabeck: Thank you that looks neater.

.github/workflows/main.yml
```diff
@@ -178,7 +178,7 @@ jobs:
   
   react-native:
     name: Check react native
-    runs-on: macOS-latest
+    runs-on: macOS-13
```
ok300, line 181: The RN / `ktlint` issue was solved with #762.

No need to change the macOS version anymore, just rebase on latest master and it should work.

ubbabeck: Thanks @ok300! Rebased to master

dangeross (approved):
LGTM!

roeierez (approved):
LGTM

erdemyerebasmaz:
@roeierez @ubbabeck 
Do we want to utilize this field on C-Breez? If so, let's create a [new issue](https://github.com/breez/c-breez/issues/new) there.

ubbabeck:
> Do we want to utilize this field on C-Breez? If so, let's create a [new issue](https://github.com/breez/c-breez/issues/new) there.

No @erdemyerebasmaz, @kingonly said that we don't need to add it in c-breez.

PR #717, opened 2024-01-10
Use Android bindings via Gradle in Flutter package

cnixbtc:
This is a precursor to publishing the flutter package to [pub.dev](https://pub.dev/) which has a 100MB package size limit. In order to meet that size limit, we need to remove the binary bindings which are currently bundled with the flutter package. Instead, we need to pull them in from our Gradle repo. We already do that for the iOS part: There we pull the binaries from CocoaPods.

_What I did to test:_

- Building the 0.2.12 version of the breez-sdk for flutter (but with gradle dependency instead of bundled jniLibs: https://github.com/breez/breez-sdk/actions/runs/7479393930
- Setting up a fresh Flutter project
- Calling the SDK method `mnemonicToSeed` via Flutter in `main.dart` to see if the Flutter APIs work.
- Calling the SDK method `mnemonicToSeed` via Kotlin in `MainActivity.kt` to see if the bundled Kotlin APIs work

I don't have a working setup of c-breez at the moment. The ultimate test would be to see if c-breez runs normally using the new setup. Will try to get that done soon. In the meantime, please let me know your feedback on the PR! 🙏

### DIFF
--- a/.github/workflows/publish-flutter.yml
+++ b/.github/workflows/publish-flutter.yml
@@ -79,7 +79,7 @@ jobs:
           sed -i.bak -e "s/s.version          = .*/s.version          = '${{ inputs.package-version }}'/" ios/breez_sdk.podspec
           rm pubspec.yaml.bak
           rm android/build.gradle.bak
-          ios/breez_sdk.podspec.bak
+          rm ios/breez_sdk.podspec.bak
 
       - name: Archive flutter release
         uses: actions/upload-artifact@v3
--- a/.github/workflows/publish-flutter.yml
+++ b/.github/workflows/publish-flutter.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Checkout breez-sdk-flutter repo
         uses: actions/checkout@v3
         with:
-          repository: breez/breez-sdk-flutter          
+          repository: breez/breez-sdk-flutter
           ssh-key: ${{ secrets.REPO_SSH_KEY }}
           fetch-depth: 0
           path: flutter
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-flutter.yml
```diff
@@ -50,22 +50,17 @@ jobs:
       - name: Copy package files
         working-directory: flutter
         run: |
-          rm -f ../breez-sdk/libs/sdk-flutter/ios/breez_sdk.podspec.dev
```
cnixbtc, line 53: This file doesn't exist anymore.

roeierez:
@cnixbtc thanks for that and for the explanation.
One question, how will this work for local development where we build the sdk locally and want to use that in c-breez?

cnixbtc:
> One question, how will this work for local development where we build the sdk locally and want to use that in c-breez?

Good question @roeierez. Local development should not be affected by this change. The dependency on the Gradle version of the binary bindings (as opposed to the locally build `jniLibs`) is declared in `build.gradle.production`. The previous `build.gradle` that doesn't have this dependency is unchanged and will be used when depending on the package locally. Only in CI will the `build.gradle` be replaced with the contents of `build.gradle.production`. That means for now this change will only affect people that pull in the lutter package via Git from `breez/breez-sdk-flutter` (and in the future from pub.dev). Hope this makes sense. I haven't tested this part yet though. Will do so as soon as I have c-breez setup to build locally.

roeierez (approved):
LGTM

cnixbtc:
Thanks for the review! 🎉 

Just to document what I did to test in addition to what's listed in the desciption:

- Replicated the [last Android build](https://github.com/breez/c-breez/actions/runs/7410761485) of c-breez
  - c-breez @ [5c24869f](https://github.com/breez/c-breez/commit/5c24869fa72f50f2311114a61283646759508dcd)
  - breez-sdk @ [e6323f57](https://github.com/breez/breez-sdk/commit/e6323f57acf0a5aad58974695b60ce07b0b77bfe)
  - Building and running c-breez works and uses the locally built `jniLibs/*` and `breez_sdk.kt` language bindings
- Applied changes from this PR (adding the `build.gradle.production` gradle file)
  - Re-ran the build
  - c-breez still uses the locally built Android bindings as expected (development mode)
- Deleted the locally built bindings and used the `build.gradle.production` file (using Maven instead of locally built bindings like the production release will do)
  - c-breez builds and runs despite the locally built bindigns not present anymore because it now pulls them from Maven

PR #718, opened 2024-01-11
Use a non-zero timeout in our HTTP clients

ok300:
The default `reqwest` Client, which we use for HTTP requests, has no timeout. This means a faulty endpoint we connect to may hang our client indefinitely. This PR adds a default timeout to all `reqwest` clients we use.

Related to https://github.com/breez/monitoring-setup/pull/18

### DIFF
--- a/libs/sdk-core/src/input_parser.rs
+++ b/libs/sdk-core/src/input_parser.rs
@@ -1,4 +1,5 @@
 use std::str::FromStr;
+use std::time::Duration;
 
 use anyhow::{anyhow, Result};
 use bip21::Uri;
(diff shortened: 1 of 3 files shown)


libs/sdk-core/src/input_parser.rs
```diff
@@ -374,6 +375,14 @@ fn lnurl_decode(encoded: &str) -> LnUrlResult<(String, String, Option<String>)>
     }
 }
 
+/// Creates an HTTP client with a built-in connection timeout
+pub(crate) fn get_reqwest_client() -> Result<reqwest::Client> {
+    reqwest::Client::builder()
+        .timeout(Duration::from_secs(60))
```
roeierez, line 381: I would use a lower number here, I suggest 30 seconds as we have in breez mobile.

roeierez (changes requested):

libs/sdk-core/src/input_parser.rs
ok300: Done

PR #723, opened 2024-01-17
Add payment and backup exceptions

ademar111190:
Enrich exception data.

This PR is a complement to https://github.com/breez/c-breez/pull/760 as I need to access the payment hash there.
Using the same structure for Backup exceptions as well; I think we can use always we want to broadcast an exception.


roeierez (approved):
LGTM

PR #727, opened 2024-01-18
Set log stream natively on Android

erdemyerebasmaz:
Set log stream natively on Android.

Required changes for
- https://github.com/breez/c-breez/pull/762

Todo: Expose attached LogStream instance's shared flow

### DIFF
--- a/libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
+++ b/libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
@@ -1,23 +1,15 @@
 package com.breez.breez_sdk
 
-import android.os.Handler
-import android.os.Looper
+import com.breez.breez_sdk.SdkLogListener
 import androidx.annotation.NonNull
-import breez_sdk.LogEntry
 import breez_sdk.LogStream
 import breez_sdk.setLogStream
-
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.plugin.common.EventChannel
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.asSharedFlow
-import kotlinx.coroutines.flow.launchIn
-import kotlinx.coroutines.flow.onEach
 
 /** BreezSDKPlugin */
 class BreezSDKPlugin : FlutterPlugin, MethodCallHandler, EventChannel.StreamHandler {
@@ -28,7 +20,7 @@ class BreezSDKPlugin : FlutterPlugin, MethodCallHandler, EventChannel.StreamHand
     private lateinit var channel: MethodChannel
     private var eventChannel: EventChannel? = null
     private var eventSink: EventChannel.EventSink? = null
-    private var nodeLogStream: LogStream? = null
+    private var nodeLogStream: SdkLogListener? = null
 
     override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         channel = MethodChannel(flutterPluginBinding.binaryMessenger, "breez_sdk")
@@ -62,36 +54,15 @@ class BreezSDKPlugin : FlutterPlugin, MethodCallHandler, EventChannel.StreamHand
         eventChannel = null
     }
 
-    private fun setNodeLogStream(): LogStream? {
+    fun setNodeLogStream(): SdkLogListener {
         // Set Log Stream
         if (nodeLogStream == null) {
             try {
-                nodeLogStream = SdkLogListener()
+                nodeLogStream = SdkLogListener(eventSink!!)
                 setLogStream(nodeLogStream!!)
             } catch (ex: Throwable) {
             }
         }
         return nodeLogStream!!
-    }
-
-    inner class SdkLogListener : LogStream {
-        // Event has to be handled on main thread.
-        private var handler = Handler(Looper.getMainLooper())
-        private val _sharedFlow = MutableSharedFlow<LogEntry>() // private mutable shared flow
-
-        override fun log(l: LogEntry) {
-            produceLog(l)
-
-            val runnable = Runnable {
-                val data = mapOf("level" to l.level, "line" to l.line)
-                eventSink?.success(data)
-            }
-
-            handler.post(runnable)
-        }
-
-        private fun produceLog(log: LogEntry) = _sharedFlow.tryEmit(log)
-
-        fun subscribe(scope: CoroutineScope, block: suspend (LogEntry) -> Unit) = _sharedFlow.onEach(block).launchIn(scope)
     }
 }
--- a/libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
+++ b/libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
@@ -1,35 +1,97 @@
 package com.breez.breez_sdk
 
+import android.os.Handler
+import android.os.Looper
 import androidx.annotation.NonNull
+import breez_sdk.LogEntry
+import breez_sdk.LogStream
+import breez_sdk.setLogStream
 
 import io.flutter.embedding.engine.plugins.FlutterPlugin
+import io.flutter.plugin.common.EventChannel
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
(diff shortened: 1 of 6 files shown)


libs/sdk-flutter/lib/breez_sdk.dart
```diff
@@ -22,7 +23,7 @@ class BreezSDK {
 
   Stream<Payment> get paymentResultStream => _paymentResultStream.stream;
 
-  void initialize() {
+  void initialize({bool usesNativeNodeLogs = false}) {
```
roeierez, line 26: This doesn't sound clear to me. Also it requires documentation to help the user understand whey the logstream is not emitting any log events...
If we want to use the native API to listen to log stream then we should do it always and wire it to the flutter code using plugin.

roeierez (changes requested):

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
```diff
@@ -1,35 +1,97 @@
...
+    override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
+        eventSink = events
+    }
+
+    override fun onCancel(arguments: Any?) {
+        eventSink = null
+        eventChannel = null
+    }
+
+    private fun setNodeLogStream(): LogStream? {
```
roeierez, line 65: As is I don't see a way we can get the log stream from the worker unless we have a public method somewhere that can be called both from the worker and the plugin.

roeierez (changes requested):

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
erdemyerebasmaz: `LogEntry`'s are emitted via a `SharedFlow` and we `subscribe()` to it to listen to it from the worker.

I didn't had a chance to test these changes but you're right, the node log stream being used should be public and we should access that instance's subscribe method.

roeierez: So perhaps taking the SdkListener outside the plugin and Adding the ability to static initialize it (and subscribe to the sdk) via a public static method that can be used from both the flutter plugin and the worker.
The MutableShareflow seems a right decision and every side may add its own subscription to it?

erdemyerebasmaz: That's right, there can be many subscribers to the log event flow.

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
```diff
@@ -1,35 +1,82 @@
...
+    override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
+        eventSink = events
+    }
+
+    override fun onCancel(arguments: Any?) {
+        eventSink = null
+        eventChannel = null
+    }
+
+    fun setNodeLogStream(): SdkLogListener {
```
roeierez, line 63: This method is not static. It should be static and outside the plugin in order for the worker to be able to initialize the  singleton los stream.

roeierez (changes requested):

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
erdemyerebasmaz: Yes, this function is not static, my thinking was that the worker has access to the [plugin instance](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/plugins/PluginRegistry.html#get(java.lang.Class)) itself, so it's not being re-initialized, and it's always attached to Flutter engine beforehand.

I'll move the event channel outside and convert into a static function.

erdemyerebasmaz: Done.
[Convert setNodeLogStream to a static function](https://github.com/breez/breez-sdk/pull/727/commits/cb70ebe563ec58940cd21e92ebbb3d2c153b77db)

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
```diff
@@ -1,35 +1,87 @@
...
+    override fun onListen(arguments: Any?, events: EventChannel.EventSink?) {
+        eventSink = events
+    }
+
+    override fun onCancel(arguments: Any?) {
+        eventSink = null
+        eventChannel = null
+    }
+
+    companion object {
```
roeierez, line 68: I think we should move this out of the plugin in a different class that handles the static log intialization.

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
```diff
@@ -1,35 +1,87 @@
...
+    private var eventSink: EventChannel.EventSink? = null
+
+    override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
+        channel = MethodChannel(flutterPluginBinding.binaryMessenger, "breez_sdk")
+        channel.setMethodCallHandler(this)
+
+        eventChannel =
+            EventChannel(flutterPluginBinding.binaryMessenger, "breez_sdk_node_logs")
+        val nodeLogStream = setNodeLogStream()
+        nodeLogStream.subscribe(CoroutineScope(Dispatchers.Main)) { l: LogEntry ->
```
roeierez, line 36: We should also handle unsubscribe when the engine detached.

roeierez (changes requested):

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
roeierez: Or perhaps it should unsubscribed at `onCancel` method?

erdemyerebasmaz: Done as part of: 59cc59973a11716aaa965a416eebc4e1876efa48

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
erdemyerebasmaz: Done as part of: 59cc59973a11716aaa965a416eebc4e1876efa48

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/SdkLogInitializer.kt
```diff
@@ -0,0 +1,26 @@
...
+                e.printStackTrace()
+                nodeLogStream = null
+                throw e
+            }
+        }
+        return nodeLogStream!!
+    }
+
+    fun unsubscribeNodeLogStream() {
+        nodeLogStream?.unsubscribe()
```
roeierez, line 24: This will cancel all subscriptions, won't it? We only want to unsubscribe the current one in case there are multiple subscriptions.

roeierez (changes requested):

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/SdkLogInitializer.kt
erdemyerebasmaz: Yeah, my bad- forgot to pass the CoroutineScope that subscribed to the log events.

libs/sdk-flutter/android/src/main/kotlin/com/breez/breez_sdk/BreezSDKPlugin.kt
```diff
@@ -1,35 +1,63 @@
...
+
+        eventChannel = EventChannel(flutterPluginBinding.binaryMessenger, "breez_sdk_node_logs")
+        val nodeLogStream = SdkLogInitializer.initializeNodeLogStream()
+        nodeLogStream.subscribe(scope) { l: LogEntry ->
+            val runnable = Runnable {
+                val data = mapOf("level" to l.level, "line" to l.line)
+                eventSink?.success(data)
+            }
+
+            handler.post(runnable)
```
roeierez, line 37: @erdemyerebasmaz  The subscribed scope seems to already use the main thread, can we omit the handler here and instead call directly the eventSink without the Runnable?

erdemyerebasmaz: Yes, good catch!
Done f1d89da3820aa17a37bb6d19d200d71559004df8

roeierez (approved):
LGTM

Issue #527, opened 2023-10-17
Feature request: include on-chain sends in payment list


danielgranhao:
In the same way a `Payment` can represent a channel closing if the type is `PaymentType::ClosedChannel`, it would be user-friendly if the payment list returned by `list_payments()` could also include transactions created by any of:
* `send_onchain()`
* `sweep()`
* `refund()`

roeierez:
We follow the principle where the payments are those that affect the lighting balance (not onchain balance) so anything that affect the lightning balance is a valid candidate to be part of the list_payment API.
Using this principle it makes sense to add special identifiers to the `send_onchain` & `receive_onchain` 
Since both of them represents today by lightning transactions (in & out) I suggest to add another two optional fields (send_onchain_tx, receive_onchain_tx) to the payment as these fields are useful anyway for the developer to show the relevant transaction to the user. This also makes it unnecessary to introduce another PaymentTypes.
@danielgranhao  What do you think?

Regarding sweep and refund which are pure onchain tx not sure we want to add it to the payments list but we can think of exposing it using a different API.
@kingonly what do you think?

danielgranhao:
Thanks for elaborating on the reasoning. That makes sense. 

Having a different API to access onchain tx history would be cool!

danielgranhao:
@roeierez I'm currently working on implementing this. 

I'm adding an optional `ReverseSwapInfo` field to the payment details (an approach identical to how swaps are handled). The issue I'm seeing is that the db only stores `FullReverseSwapInfo`s, and network requests are required to convert into the user-facing `ReverseSwapInfo`. 

I think we need to cache in the db the `lockup_txid` and `claim_txid` to avoid accessing the network when listing payments, but I wanted to get feedback before working on such a solution.

roeierez:
@danielgranhao I agree no network request should be done when the user asks for list_payments.
I think we can safely cache these values for the list_payment purpose but use the cache only in list payments and keep doing network requests to get the up to date values in the other flows as today (refreshing the cached values as well).
Reason is that we are considering also mempool transaction for these values which an be changed/overriden by the swapper.
Does it sound reasonable to you?

ok300:
A different approach could be as follows:

The `Payment` could include an optional `rev_swap_id: Option<String>`. If set, this invoice was associated with a reverse swap ID.

[...]

PR #728, opened 2024-01-18
Include `ReverseSwapInfo` in `Payment`

danielgranhao:
Addresses #527.

TODO:
- [x] Bindings


danielgranhao:
I am looking for some feedback on the current implementation before fixing the bindings.

libs/sdk-core/src/swap_out/reverseswap.rs
```diff
@@ -541,13 +552,32 @@ impl BTCSendSwap {
...
-    /// Updates the state of given reverse swap in the cache table, if the status has changed
+    /// Updates the cached values of given reverse swap in the cache table, if the status has changed
     async fn refresh_reverse_swap(&self, rsi: FullReverseSwapInfo) -> Result<()> {
         match self.get_status_update_for_monitored(&rsi).await? {
             None => Ok(()),
-            Some(new_status) => Ok(self
-                .persister
-                .update_reverse_swap_status(&rsi.id, &new_status)?),
+            Some(new_status) => {
+                let lockup_txid = self
```
roeierez, line 560: The claim tx is the second transaction (after the lockup) and is constructed and broadcasted in `execute_pending_reverse_swaps` function (which is called right after `refresh_reverse_swap` function call.
With this addition we call the `get_lockup_tx` twice in a row. Why not cache it right after we broadcast it and omit it from here?

roeierez (changes requested):

libs/sdk-core/src/persist/reverseswap.rs
```diff
@@ -25,30 +25,39 @@ impl SqliteStorage {
...
         Ok(())
     }
 
-    pub(crate) fn update_reverse_swap_status(
+    pub(crate) fn update_reverse_swap_cache(
         &self,
         id: &str,
         status: &ReverseSwapStatus,
+        lockup_txid: &str,
+        claim_txid: &str,
```
dangeross, line 47: These could be optional strings instead of passing empty strings.
```suggestion
        lockup_txid: Option<String>,
        claim_txid:  Option<String>,
```

libs/sdk-core/src/swap_out/reverseswap.rs
danielgranhao: Makes sense. Applied in [100efb1](https://github.com/breez/breez-sdk/pull/728/commits/100efb106c287e41c5bbf5248ae84ea4ada85064).

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -415,7 +415,10 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
        "SELECT 1;", // Placeholder statement, to avoid that column is added twice (from sync fn below and here)
        "ALTER TABLE channels ADD COLUMN alias_local TEXT;",
        "ALTER TABLE channels ADD COLUMN alias_remote TEXT;",
-       "ALTER TABLE channels ADD COLUMN closing_txid TEXT;",
+       "ALTER TABLE channels ADD COLUMN closing_txid TEXT;", 
+       "
+       ALTER TABLE reverse_swaps_info ADD COLUMN lockup_txid TEXT;\
```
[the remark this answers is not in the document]
danielgranhao: Yeah, it was introduced automatically by the IDE. Definitely not needed. Removed in [100efb1](https://github.com/breez/breez-sdk/pull/728/commits/100efb106c287e41c5bbf5248ae84ea4ada85064).

libs/sdk-core/src/persist/reverseswap.rs
```diff
@@ -57,17 +66,42 @@ impl SqliteStorage {
 
     pub(crate) fn list_reverse_swaps(&self) -> PersistResult<Vec<FullReverseSwapInfo>> {
         let con = self.get_connection()?;
-        let mut stmt = con.prepare(&self.select_reverse_swap_query())?;
+        let mut stmt = con.prepare(&self.select_reverse_swap_query(""))?;
```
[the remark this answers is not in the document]
danielgranhao: I had a look into `persist/swap.rs` and there a string slice is used in a very similar method. The difference is that instead of an empty string, `"true"` is being passed, which makes the intent of the `where_clause` parameter more clear. I've applied the same approach in [100efb1](https://github.com/breez/breez-sdk/pull/728/commits/100efb106c287e41c5bbf5248ae84ea4ada85064) but I can also change to an optional string in both places if that makes sense.

libs/sdk-core/src/persist/reverseswap.rs
```diff
@@ -25,30 +25,39 @@ impl SqliteStorage {
...
-            "INSERT INTO reverse_swaps_info (id, status)\
-            VALUES (:id, :status)",
+            "INSERT INTO reverse_swaps_info (id, status, lockup_txid, claim_txid)\
+            VALUES (:id, :status, :lockup_txid, :claim_txid)",
             named_params! {
                 ":id": rsi.id,
-                ":status": serde_json::to_value(rsi.cache.status)?
+                ":status": serde_json::to_value(rsi.cache.status)?,
+                ":lockup_txid": "",
+                ":claim_txid": "",
```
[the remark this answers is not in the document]
danielgranhao: No longer applicable after addressing another comment.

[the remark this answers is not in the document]
danielgranhao: Applied in [100efb1](https://github.com/breez/breez-sdk/pull/728/commits/100efb106c287e41c5bbf5248ae84ea4ada85064)

libs/sdk-core/src/persist/reverseswap.rs
danielgranhao: No longer applicable after addressing another comment.

libs/sdk-core/src/swap_out/reverseswap.rs
```diff
@@ -170,9 +170,17 @@ impl BTCSendSwap {
...
+            Ok(_) => {
+                let lockup_txid = self
+                    .get_lockup_tx(&created_rsi)
+                    .await?
+                    .map(|lockup_tx| lockup_tx.txid)
+                    .unwrap_or_default();
+                self.persister
+                    .update_reverse_swap_status(&created_rsi.id, &InProgress)?;
+                self.persister
+                    .update_reverse_swap_lockup_txid(&created_rsi.id, &lockup_txid)?
```
dangeross, line 182: At this point the lockup_txid could still potentially be "" if it's not been seen by the mempool after the reverse swap in created? (tell me if I'm wrong). I think it's preferable if the lockup_txid column is only set if it's a valid lockup tx.

dangeross (commented):
Do we still need to look at BreezServices `send_onchain` and `in_progress_reverse_swaps`? The swapper's `convert_reverse_swap_info` method called on each swap, in turn calling `get_lockup_tx` and `create_claim_tx` (when applicable). Could we simply use the cached txid's here?

libs/sdk-core/src/swap_out/reverseswap.rs
roeierez: @danielgranhao  it actually took me a while to understand where exactly we are doing some network requests to get the claim tx or lockup tx regardless of you new code.
I think the flow wasn't clear enough from the code before your code change and your change exposes that.
The problem IMO is that both `refresh_monitored_reverse_swaps` and `execute_pending_reverse_swaps` intersects in their logic for example in `refresh_monitored_reverse_swaps` we are querying the blockchain for the claim tx and doing the same thing also in `execute_pending_reverse_swaps` which follows right after.
So if we want to add a cach layer on to of it we need to change the code first to have a single point where these transactions are fetched from the blockhain and then persisted.

Given the above I think there is no reason for this two steps process (refresh, execute) and we can use one function (currently most of the logic is inside `get_status_update_for_monitored` ) where we:
1. Always fetch both transactions (lock, claim) and cache them
2. Refresh the status and persist it
3. Broadcast the claim transaction if needed (as we are doing in `execute_pending_reverse_swaps` )

It is a bit of refactoring but I think it is necessary. Let me know what do you think.

roeierez (changes requested):

danielgranhao:
@roeierez Yes, my understanding of this reverse swap logic was also incomplete. I've refactored it in [895fad2](https://github.com/breez/breez-sdk/pull/728/commits/895fad29208a86f31ffa4ad846666893df63fe65). I think now it's more understandable, but please let me know what you think.

libs/sdk-core/src/swap_out/reverseswap.rs
danielgranhao: > Do we still need to look at BreezServices send_onchain and in_progress_reverse_swaps? The swapper's convert_reverse_swap_info method is called on each swap, in turn calling get_lockup_tx and create_claim_tx (when applicable). Could we simply use the cached txid's here?

The idea would be not to change these other flows so that they always get up-to-date information. For the purpose of `list_payments`, it isn't as important that the information is synced. WDYT?

> At this point the lockup_txid could still potentially be "" if it's not been seen by the mempool after the reverse swap in created? (tell me if I'm wrong). I think it's preferable if the lockup_txid column is only set if it's a valid lockup tx.

True. Addressed in [895fad2](https://github.com/breez/breez-sdk/pull/728/commits/895fad29208a86f31ffa4ad846666893df63fe65).

roeierez:
> @roeierez Yes, my understanding of this reverse swap logic was also incomplete. I've refactored it in [895fad2](https://github.com/breez/breez-sdk/pull/728/commits/895fad29208a86f31ffa4ad846666893df63fe65). I think now it's more understandable, but please let me know what you think.

@danielgranhao I like it! I think it is a lot more clearer now.
@ok300 can you review this as well?

ok300 (approved):
Looks good!

A few things were simplified along the way too, like the `TxStatus` conversion, which is good to see.

roeierez (approved):
LGTM! Thanks you @danielgranhao

dangeross (approved):
LGTM! Just need to resolve the conflicts

danielgranhao:
@dangeross Conflicts resolved :)

danielgranhao:
Seems like CI is complaining about `BreezSDKModule.kt`, which I had previously modified by having an older version of `ktlint` installed. I've already updated `ktlint` and rebased it so that my changes to that file were removed, but it continues to fail. Any suggestions regarding how to fix this?

ok300:
The kotlin check is a bit finicky at the moment. I re-triggered it and its green.

We're looking for a more reliable solution, but for now this works.

PR #732, opened 2024-01-19
Add payment metadata functionality to sdk-cli

hydra-yse:
_Note:_ Had to make some changes to check whether or not the payment exists, due to the fact that there is no foreign key relation on the `payments_metadata` table

### DIFF
--- a/tools/sdk-cli/src/command_handlers.rs
+++ b/tools/sdk-cli/src/command_handlers.rs
@@ -222,11 +222,12 @@ pub(crate) async fn handle_command(
                     let mut filters = vec![];
 
                     for filter in raw_filters.iter() {
-                        let v = filter.split(':').collect::<Vec<&str>>();
+                        let (json_path, json_value) =
+                            filter.split_once(':').context("Invalid metadata filter")?;
 
                         filters.push(MetadataFilter {
-                            json_path: v.first().context("Invalid JSON path")?.to_string(),
-                            json_value: v.get(1).context("Invalid JSON value")?.to_string(),
+                            json_path: json_path.to_string(),
+                            json_value: json_value.to_string(),
                         });
                     }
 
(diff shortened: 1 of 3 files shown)


tools/sdk-cli/src/commands.rs
```diff
@@ -144,6 +144,16 @@ pub(crate) enum Commands {
         /// Optional offset in payments
         #[clap(short = 'o', long = "offset")]
         offset: Option<u32>,
+
+        /// Optional metadata filter, in the form of json_path:json_value,...
+        #[clap(short = 'm', long = "metadata", num_args = 1.., value_delimiter = ',')]
```
dangeross, line 149: Can't the value_delimiter `,` and the split token `:`  be a valid components of json_value? 

E.g. `supportsNested:{"value":[1,2]}` or `supportsNested.value:[1,2]`

hydra-yse: True, that'd be the case for objects/arrays. Perhaps a delimiter isn't necessary, and we can use `-m` repeatedly. Concerting the `:`, I'll just join the slices into a single string. That should avoid the issue.

tools/sdk-cli/src/command_handlers.rs
```diff
@@ -215,11 +215,30 @@ pub(crate) async fn handle_command(
...
+                Some(raw_filters) => {
+                    let mut filters = vec![];
+
+                    for filter in raw_filters.iter() {
+                        let i = filter.find(':').context("Invalid metadata filter")?;
+
+                        filters.push(MetadataFilter {
+                            json_path: filter[0..i].to_string(),
+                            json_value: filter[i + 1..].to_string(),
+                        });
```
hydra-yse, line 230: Perhaps the more proper solution @dangeross. Let me know what you think

dangeross: Or maybe `filter.split_once(':')`?

hydra-yse: Guess that's just more Rust magic to know of. Fixed.

dangeross (approved):
LGTM!

PR #735, opened 2024-01-20
Store LNURL-pay domain

### LINKED ISSUE #729: Persist and expose domain of LNURL-pay service
When paying LNURL-pay the domain of LNURL-pay service has to be shown to the user when asking for payment input, as per LUD-06 spec. `breez_sdk_core::input_parser::LnUrlPayRequestData` struct kindly provides this information, but once the payment is complete this information gets lost.

The domain should be persisted with the payment and exposed in `breez_sdk_core::LnPaymentDetails` similar to `ln_address` such that the user may know what domain he payed to.

andrei-21:
Hi, please review my proposal for storing and exposing an LNURL-pay domain. I recommend storing either the lightning address or LNURL-pay domain, as having both doesn't make sense. We can then repurpose the `sync.payments_external_info.ln_address` database field to store one of the values (though the field should be renamed).

If the concept is acceptable, I can proceed with preparing the code for production.

Close #729.

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -709,7 +709,7 @@ pub struct LnPaymentDetails {
     /// the endpoint returns a success action
     pub lnurl_success_action: Option<SuccessActionProcessed>,
 
-    /// Only set for [PaymentType::Sent] payments if it is not a payment to a Lightning Adddress
+    /// Only set for [PaymentType::Sent] payments if it is not a payment to a Lightning Address
     pub lnurl_pay_domain: Option<String>,
 
     /// Only set for [PaymentType::Sent] payments that are sent to a Lightning Address
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -639,6 +639,7 @@ pub struct Payment {
 #[derive(Default)]
 pub struct PaymentExternalInfo {
     pub lnurl_pay_success_action: Option<SuccessActionProcessed>,
+    pub lnurl_pay_domain: Option<String>,
     pub lnurl_metadata: Option<String>,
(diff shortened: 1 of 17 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -640,7 +640,7 @@ pub struct Payment {
 pub struct PaymentExternalInfo {
     pub lnurl_pay_success_action: Option<SuccessActionProcessed>,
     pub lnurl_metadata: Option<String>,
-    pub ln_address: Option<String>,
+    pub ln_address_or_domain: Option<String>,
```
roeierez, line 643: @andrei-21  In LnPaymentDetails we added another field for domain, why not adding here as well instead of reusing the same field for both values?

roeierez:
> @roeierez are we happy with this approach? If yes I proceed with bindings and add some tests.

@andrei-21  Perhaps to be consistent we can add the field to the persist layer as well instead of renaming the column?

roeierez:
> @roeierez are we happy with this approach? If yes I proceed with bindings and add some tests.

Yes, looks good to me.

roeierez (approved):
LGTM

roeierez:
> I added tests and squashed commits into one. Regarding bindings, am I right that it is pretty much not feasible to generate them on Linux because of swiftformat?

That's probably right. We will generate the binding before merging.

ok300:
In the meantime, I generated the bindings here: https://github.com/andrei-21/breez-sdk/pull/1

ok300:
@andrei-21 once you merge that PR into your branch, the changes will appear here, so we can have a final look before merging this.

ok300 (approved):
Looks good!

PR #736, opened 2024-01-20
make short_channel_id optional

JssDWt:
short_channel_id is only set when the channel is confirmed onchain. This commit makes the short_channel_id optional, to match reality.

Should fix the error `short_channel_id is missing`.

The database column was already nullable.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -293,9 +293,9 @@ dictionary LnPaymentDetails {
 };
 
 dictionary ClosedChannelPaymentDetails {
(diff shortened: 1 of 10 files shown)


dangeross:
Is it worth adding a test case where the short_channel_id is None?

JssDWt:
> Is it worth adding a test case where the short_channel_id is None?

Makes sense. I added a test that syncs opened and closed channel. It also updates the short_channel_id later to make sure it is set once the funding tx is confirmed and the short_channel_id is set on the response from the node.

dangeross (approved):
LGTM!

ok300 (approved):
Looks good!

PR #741, opened 2024-01-23
Use Swift bindings via CocoaPods in Flutter package

erdemyerebasmaz:
- #717 cont.

> We already do that for the iOS part: There we pull the binaries from CocoaPods.

This wasn't actually the case for iOS, the built binaries were previously added as an on demand resource.

Swift bindings are now pulled via CocoaPods with this PR.


cnixbtc:
Wondering if we could remove that part now? 

https://github.com/breez/breez-sdk/blob/main/.github/workflows/publish-flutter.yml#L67

erdemyerebasmaz:
> Wondering if we could remove that part now?
> 
> https://github.com/breez/breez-sdk/blob/main/.github/workflows/publish-flutter.yml#L67

Yes, I missed that! Removing.

roeierez (approved):
LGTM

cnixbtc:
Thanks @erdemyerebasmaz!

PR #740, opened 2024-01-23
Export Android bindings dependency of Flutter package

erdemyerebasmaz:
- #717 cont.
 
Exports Android bindings dependency of Flutter package so it can be accessed natively on projects that use this package.

PR #742, opened 2024-01-23
Do not bundle pre-built binaries with Flutter plugin

erdemyerebasmaz:
In light of using Kotlin & Swift bindings on Flutter plugin via Gradle & CocoaPods, respectively, we're removing the obsolete makefile commands and workflow steps.

Related PRs:
- #717 
  - #740
- #741


cnixbtc (approved):
LGTM

PR #743, opened 2024-01-25
Add dart bindings for payment metadata

hydra-yse:


libs/sdk-core/src/binding.rs
```diff
@@ -261,6 +261,12 @@ pub fn payment_by_hash(hash: String) -> Result<Option<Payment>> {
         .map_err(anyhow::Error::new::<SdkError>)
 }
 
+/// See [BreezServices::set_payment_metadata]
```
roeierez, line 264: @hydra-yse we should also generate the dart bindings using the flutter rust bridge

roeierez (changes requested):
LGTM

roeierez (approved):

erdemyerebasmaz:
@hydra-yse We can't use the generated Dart bindings by default on our Flutter plugin as there's additional mandatory steps we need to take for some API's & streams on a secondary wrapper class, `BreezSDK`. To expose this method by the Flutter plugin, it needs to be added to `libs/sdk-flutter/lib/breez_sdk.dart`. 

I suggest adding it after `paymentByHash` API.
```
/// Set the external metadata of a payment as a valid JSON string
Future<void>  setPaymentMetadata({
  required String hash,
  required String metadata,
}) async {
  return await _lnToolkit.setPaymentMetadata(hash: hash, metadata: metadata);
}
```

CC: @roeierez

PR #745, opened 2024-01-25
CLI: Pretty-format dev command results

ok300:
The protobuf structs (dev command results) already derive `Debug`, which means that `println!("{res:#?})` will pretty-print them, with each field and inner struct indented.

However, since we don't print that directly, but return it as a String from `sdk.execute_dev_command()`, we have to un-escape a few characters.

So, as a result of this PR:
- `sdk.execute_dev_command()` returns a debug-like (indented) string representation of the structs
- the SDK CLI pretty-prints these dev command results


ok300:
Unfortunately this makes the dev command output in c-breez worse, so I'm abandoning this.

Thanks @ubbabeck for the tip.

PR #748, opened 2024-01-26
fix empty sent payment list

roeierez:


ubbabeck (approved):
nice catch 🥵

JssDWt (commented):
Approved

PR #749, opened 2024-01-26
Move early-return check to top of method

ok300:
This pattern (having early return checks as early as possible) prevents issues like the one fixed in `0.2.15`.

(PR is on top of `0.2.15`, so marking as draft until `0.2.15` is merged into `main`).


JssDWt (approved):
LGTM

PR #754, opened 2024-01-26
Bump Node version from 16 (EOL) to 20 in CI step

ok300:
Address CI warnings that point to an EOL Nodejs version:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.


dangeross (approved):
LG. Tested CI without publish

PR #755, opened 2024-01-26
remove lnurl pay description hash validation

JssDWt:
The description hash validation is no longer required for lnurl pay. https://github.com/lnurl/luds/pull/234


dangeross (approved):
LGTM

PR #756, opened 2024-01-26
CI: upgrade to macos-13

JssDWt:
Changing the runner to macos-13 in the main CI fixes the react native binding generation check.
I figured it's a good idea to upgrade all runners to macos-13, because we've been running into more issues with the macos-latest version lately.
It's still in beta though, so not sure whether things will break again soon.

### DIFF
--- a/.github/workflows/build-bindings-darwin.yml
+++ b/.github/workflows/build-bindings-darwin.yml
@@ -25,7 +25,7 @@ on:
 jobs:
   build:
     if: ${{ !inputs.use-dummy-binaries }}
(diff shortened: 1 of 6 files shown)


ok300 (approved):

ok300:
This might not be necessary if #759 works.

JssDWt:
I think @ok300's idea is better. I prefer https://github.com/breez/breez-sdk/pull/760 myself.

JssDWt:
preferred https://github.com/breez/breez-sdk/pull/762

PR #757, opened 2024-01-27
Fix reqwest pinning

ok300:
Unfortunately the cargo versioning rules are not very intuitive.

From [the docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio):

> ```toml
> [dependencies]
>time = "0.1.12"
>```
>The string "0.1.12" is a version requirement. Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows [SemVer](https://semver.org/) compatible updates. An update is allowed if the new version number does not modify the left-most non-zero number in the major, minor, patch grouping. In this case, if we ran cargo update time, cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0.

This means that
```toml
reqwest = { version = "0.11.20", features = ["json"] }
```

would allow, for example, version `0.11.22`.

However, we want to limit it to `0.11.20` at the most, which can be done with

```toml
reqwest = { version = "=0.11.20", features = ["json"] }
```

### DIFF
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -567,7 +567,7 @@ dependencies = [
  "breez-sdk-core",
  "camino",
  "flutter_rust_bridge",
- "lightning-invoice 0.23.0",
+ "lightning-invoice 0.26.0",
  "log",
  "once_cell",
  "thiserror",
@@ -1296,7 +1296,7 @@ dependencies = [
  "async-trait",
  "base64 0.21.5",
  "bech32",
- "bitcoin 0.29.2",
+ "bitcoin 0.30.2",
  "bytes",
  "chacha20poly1305",
  "cln-grpc",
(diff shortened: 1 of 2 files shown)


ok300:
Not sure why the CI doesn't like `<=`, as it worked for me locally. I switched it to `=`.

libs/Cargo.lock
```diff
@@ -1296,7 +1296,7 @@ dependencies = [
  "async-trait",
  "base64 0.21.5",
  "bech32",
- "bitcoin 0.29.2",
+ "bitcoin 0.30.2",
```
hydra-yse, line 1299: :thinking: I wonder why this updates the bitcoin and lightning-invoice crate as well (as the only crate changed was reqwest) though there don't seem to be any problems with building, so should be alright.

hydra-yse (approved):
LGTM!

libs/Cargo.lock
ok300: It's possible this selected-in or selected-out other dependencies too, based on how they set their own dependency ranges for `reqwest`.

Issue #703, opened 2024-01-02
Add the payer invoice to the payment in case of JIT channel


roeierez:
This was requested by @andrei-21 in telegram.
It seems that it will also solve https://github.com/breez/breez-sdk/issues/702

hydra-yse:
If my understanding is correct, the invoice should already be present inside the `PaymentDetails`, right? So when a Payment is related to a transaction (not closed channel), you should be able to retrieve it: https://github.com/breez/breez-sdk/blob/82e30c4e9c898748131fc434fc832b010f3b74d0/libs/sdk-core/src/models.rs#L714-L720

JssDWt:
> If my understanding is correct, the invoice should already be present inside the `PaymentDetails`, right? So when a Payment is related to a transaction (not closed channel), you should be able to retrieve it:

The invoice shown there is a different invoice than we sent to the payer. Because the recipient receives a smaller amount than the sender sent in case of a jit channel. So we store a smaller invoice on the node, and modify the invoice to the actual amount to send to the sender.

hydra-yse:
Thanks @JssDWt , I understand now. Storing the actual invoice sent to the payer though does seem to require changes to the lsp's reply, as currently it is just an empty object:
https://github.com/breez/breez-sdk/blob/c60b12264e7f9103283c8fa4c57e2c3423faaeec/libs/sdk-core/src/grpc/proto/breez.proto#L176
Thinking on it again, maybe this is not necessary for now. As pointed out in PR https://github.com/breez/breez-sdk/pull/758, maybe just checking if `payer_amount_msat` is not null would be enough to satisfy #702's requirements, without needing any further changes to the LSP. @roeierez Let me know what you think.

PR #758, opened 2024-01-27
Add JIT payment filter to `list_payments`

### LINKED ISSUE #702: Feature Request: Explicit list_channel_history
If a user wants to calculate the fees they've paid to an LSP, there currently isn't a great way to do so. 

They could use `list_payments` and filter payments where the `amount_msat` is significantly lower than the amount encoded within the `bolt11` invoice... but this doesn't seem like a good solution.

I think it'd be helpful to:  
1. Add field(s) to `list_payments` to distinguish if a payment was involved in a JIT channel creation and include the fee paid to the LSP. (perhaps by specifying `requested_amount`, `lsp_fee`, and/or `received_amount`)
2. Add another call to only show those payments (e.g. `list_channel_history`).

hydra-yse:
Closes #702 and #703.


hydra-yse:
As a side note, there seems to be an issue in RN CI where the `ktlint` version used fluctuates between 1.0.1 and 1.1.1 depending on the instance, resulting in the same code passing sometimes and failing others.

ok300:
> Let me know if you believe the filter should be part of the `ListPaymentsRequest`, or if it's better this way.

My main critique to the new `ListPaymentsRequest` field is that it basically overrides the `filters`, which can confuse the user.

Since this is a separate category of payments, IMO the cleanest way would be to add a new `PaymentType`:
```rust
pub enum PaymentType {
    Sent,
    Received,
    ClosedChannel,
    OpenChannelFee // or JitPayment, etc
}
```

and a corresponding `PaymentTypeFilter`.

Then `ListPaymentsRequest` wouldn't need a new field.

What do you think?

hydra-yse:
@ok300 It [previously](https://github.com/breez/breez-sdk/pull/758/commits/fd842312f33484bb48f254aa1b119d5ecf0f0833) worked that way, but since the filters are supposed to match multiple cases (e.g. when a user looks for Sent and ReceivedJIT, they are expecting payments which match both, not just one), I opted for splitting it up. That way the user knows that, when `is_jit` is specified, the only results they _can_ expect are only received payments.

ok300:
I see, thanks, I wasn't aware of this previous version.

> when a user looks for Sent and ReceivedJIT, they are expecting payments which match both, not just one

Match both = match either of them, right? Isn't that how the filter list is supposed to work? I may be missing something, but I don't see where the problem with that is.

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -445,6 +447,13 @@ fn filter_to_where_clause(
         });
     }
 
+    if let Some(true) = is_jit {
+        where_clause.push(format!(
+            "payment_type = '{}' AND payer_amount_msat IS NOT NULL",
+            PaymentType::Received
+        ));
+    }
+
```
hydra-yse, line 456: @ok300 Yes, the user would expect to match either `Sent` and `ReceivedJIT`, though due to the above code snippet, the only results returned will be `ReceivedJIT` (since `payer_amount_msat` can only be not-null in received payments). 
If we really want to opt for this approach, we could change the payment-type filter clause from `payment_type in (<type1> <type2>)`  to an `OR` sequence: 
```sql
payment_type = <type1> OR payment_type = <type2> ... OR payment_type = ReceivedJIT and payer_amount_msat IS NOT NULL
```
But although I considered it I'm not sure whether it's the best approach.

ok300: Oh I see, thanks.

Yes I think the `OR` sequence is a good idea, I'll explain below.

From what I can tell, the current version can lead to
```sql
WHERE
  # payment filters
  payment_type IN (A, B, C) AND
  # jit clauses
  payment_type = 'Received' AND payer_amount_msat IS NOT NULL
```

If the user doesn't include `Received` as a filter, but sets `is_jit = true`, the query will fail, because all these checks are chained via `AND`.

This wouldn't be a problem with your idea above, of chaining the `where` clauses in an `OR` sequence.

Besides, it even allows for paranthesis, which makes the query clearer

```sql
payment_type = <type1> OR
  payment_type = <type2> OR
  (payment_type = ReceivedJIT and payer_amount_msat IS NOT NULL)
```

I'd suggest going for this. Its clearer for the user (jit is just another filter in the list) and for us (better query structure).

roeierez: There is one thing that is a bit weird with this approach. We have the payment_type field on the Payment structure and so if you query for `ReceivedJIT` you will get payment with type `Received`...

hydra-yse: @roeierez Shouldn't that be intended? A received-jit is just a subset of a received payment, so it should be intuitive that it be of type received

hydra-yse: @ok300 I have implemented the `OR` approach, there shouldn't be any issues there. The code actually turned out pretty readable, and performance wise it may only be slightly worse.

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -40,24 +40,28 @@ impl SqliteStorage {
...
-           description,
-           details
+           description
         )
-         VALUES (?1,?2,?3,?4,?5,?6,?7,?8)
+         VALUES (?1,?2,?3,?4,?5,?6,?7)
         ",
         )?;
 
+        let mut update_details = con.prepare(
```
ok300, line 49: Why do `details` have to be handled separately? Please add a short comment.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2129,6 +2130,40 @@ impl Receiver for PaymentReceiver {
             info!("Signed invoice with hint = {}", signed_invoice_with_hint);
 
             parsed_invoice = parse_invoice(&signed_invoice_with_hint)?;
+
+            self.persister.insert_or_update_payments(
```
roeierez, line 2134: We already have a point where we use `insert_open_channel_payment_info` to add the payer invoice amount.
I think this method call can be changed to contain the payer invoice instead of calling `insert_or_update_payments`

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -196,6 +196,7 @@ dictionary CheckMessageResponse {
 enum PaymentTypeFilter {
     "Sent",
     "Received",
+    "ReceivedJIT",
```
roeierez, line 199: Now that we persist the payer_bolt11 perhaps we don't need the new `ReceivedJIT` type?

roeierez (changes requested):

hydra-yse:
After some thought I decided to opt for the alternative proposed by @roeierez, leaving the `payer_bolt11` as part of the payment details and letting the user handle the filtering of such payments themselves.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -292,6 +292,7 @@ dictionary LnPaymentDetails {
     string payment_preimage;
     boolean keysend;
     string bolt11;
+    string? payer_bolt11;
```
roeierez, line 295: NIT: Since we have decided to expose the implementation details here perhaps the name "open_channel_bolt11" better fits here? I am thinking also of this field as the one that distinguish jit payments from the other.

roeierez (changes requested):
LGTM! I only added a small NIT, other than that ready to go IMO.

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -555,5 +555,6 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
          ) STRICT;
         ",
        "ALTER TABLE payments_external_info ADD COLUMN lnurl_pay_domain TEXT;",
+       "ALTER TABLE open_channel_payment_info ADD COLUMN payer_bolt11 TEXT;",
```
roeierez, line 558: Better to rename here also the `payer_bolt11` to `open_channel_bolt11`

roeierez (changes requested):

roeierez (approved):
LGTM

ok300 (approved):

PR #759, opened 2024-01-27
Pin ktlint version to avoid random fluctuations in GH CI runners

ok300:
As described [here](https://github.com/breez/breez-sdk/pull/758#issuecomment-1912930109), the `ktlint` version used by the CI randomly alternates between 1.0.1 and 1.1.1. This causes valid formatting to appear invalid and vice-versa.

This PR proposes to explicitly set the version when `brew install`ing it.

Another solution tried in https://github.com/breez/breez-sdk/pull/756 is to use a different macos base image, where this doesn't appear to happen. I tested it by running the CI 7x, and the `ktlint` check was successful every time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -203,7 +203,7 @@ jobs:
         working-directory: libs/sdk-react-native 
         run: |
           yarn global add tslint typescript
-          brew install kotlin ktlint swiftformat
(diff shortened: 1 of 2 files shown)


ok300:
Looks like it's more complicated: https://gist.github.com/zmts/b5f3e98984f9cd3ed9852b1f34874c22

IMO not worth the extra steps and complexity.

Closing this in favor of #756.

PR #760, opened 2024-01-27
pin ktlint to version 1.1.1

JssDWt:
As an alternative to https://github.com/breez/breez-sdk/pull/756, pin the ktlint to version 1.1.1

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -203,7 +203,8 @@ jobs:
         working-directory: libs/sdk-react-native 
         run: |
           yarn global add tslint typescript
-          brew install kotlin ktlint swiftformat
+          brew install kotlin swiftformat
+          curl -sSLO https://github.com/pinterest/ktlint/releases/download/1.1.1/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/local/bin/
           make react-native-codegen
 
       - name: Check git status


dangeross (approved):
LGTM

ok300 (approved):

.github/workflows/main.yml
```diff
@@ -203,7 +203,8 @@ jobs:
         working-directory: libs/sdk-react-native 
         run: |
           yarn global add tslint typescript
-          brew install kotlin ktlint swiftformat
+          brew install kotlin swiftformat
+          curl -sSLO https://github.com/pinterest/ktlint/releases/download/1.1.1/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/local/bin/
```
roeierez, line 207: Perhaps we can use homebrew to install the specific version?

`brew install ktlint@1.1.1 `

roeierez (changes requested):

.github/workflows/main.yml
ok300: I tried that in #759 but it didn't work.

ok300: Looks like that only works if the homebrew formula has multiple versions published. For example: https://formulae.brew.sh/formula/postgresql@12

For `ktlint` the formula has a single version: https://formulae.brew.sh/formula/ktlint , so that syntax won't work for it.

roeierez: I missed that.

roeierez (approved):

ok300:
@JssDWt how about #762 as an alternative to this?

JssDWt:
@ok300 yeah let's do https://github.com/breez/breez-sdk/pull/762

PR #761, opened 2024-01-27
Split `Payment` struct depending on method 

### LINKED ISSUE #563: Seperate between PaymentListItem and Payment
`send_payment` and `send_spontaneous_payment` use the same return type `Payment` as `list_payments`. They mean very different things, however. `list_payments` is meant to give you a list of all the balance mutations: receives, sends, closed channels. 

Suggestion is to rename the current `Payment` to `PaymentListItem`. It will be returned by `list_payments`.
And use a different model that is more suited for the `send_payment` and `send_spontaneous_payment` functions.

hydra-yse:
Fixes #563.
_Note:_ I understand that `PendingPayment` could cause some confusion given that a `PaymentListItem` can also have "pending" as status (meaning that the two structs are identical/repetitive), though there are instances in which the distinction between the two is clear enough to allow for this kind of naming. I'm open to suggestions, though.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -216,7 +216,7 @@ enum PaymentType {
     "ClosedChannel"
 };
 
-dictionary Payment {
+dictionary PaymentListItem {
     string id;
     PaymentType payment_type;    
     i64 payment_time;    
@@ -227,6 +227,15 @@ dictionary Payment {
     string? description;
     PaymentDetails details;
     string? metadata;
+};
+
+dictionary PendingPayment {
+    string id;
+    i64 payment_time;    
+    u64 amount_msat;
+    u64 fee_msat;
+    string? description;
+    PaymentDetails details;
 };
 
 dictionary ListPaymentsRequest {
@@ -360,7 +369,7 @@ dictionary LogEntry {
 dictionary InvoicePaidDetails {
     string payment_hash;
     string bolt11;
-    Payment? payment;
+    PaymentListItem? payment;
 };
 
 dictionary PaymentFailedData {
@@ -378,7 +387,7 @@ interface BreezEvent {
     NewBlock(u32 block);
     InvoicePaid(InvoicePaidDetails details);
     Synced();
-    PaymentSucceed(Payment details);
+    PaymentSucceed(PendingPayment details);
     PaymentFailed(PaymentFailedData details);
     BackupStarted();
     BackupSucceeded(); 
@@ -697,7 +706,7 @@ dictionary SendSpontaneousPaymentRequest {
 };
 
 dictionary SendPaymentResponse {
-    Payment payment;
+    PaymentListItem payment;
 };
 
 dictionary MaxReverseSwapAmountResponse {
@@ -781,10 +790,10 @@ interface BlockingBreezServices {
    void backup();
 
    [Throws=SdkError]
-   sequence<Payment> list_payments(ListPaymentsRequest req);
+   sequence<PaymentListItem> list_payments(ListPaymentsRequest req);
 
    [Throws=SdkError]
-   Payment? payment_by_hash(string hash);
+   PaymentListItem? payment_by_hash(string hash);
 
    [Throws=SdkError]
    void set_payment_metadata(string hash, string metadata);
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -621,7 +621,7 @@ pub struct NodeState {
 /// Internal response to a [NodeAPI::pull_changed] call
 pub struct SyncResponse {
     pub node_state: NodeState,
-    pub payments: Vec<crate::models::Payment>,
+    pub payments: Vec<crate::models::PaymentListItem>,
     pub channels: Vec<crate::models::Channel>,
 }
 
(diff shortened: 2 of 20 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -649,6 +649,19 @@ pub struct Payment {
     pub metadata: Option<String>,
 }
 
+/// Represents a sent and pending payment, not yet persisted into the storage,
+/// including its [PaymentType] and [PaymentDetails]
+#[derive(PartialEq, Eq, Debug, Clone, Serialize, Deserialize)]
+pub struct PendingPayment {
```
roeierez, line 655: I am not sure we need this struct as what we really want to return from send_payment & send_spontaneous_payment is something that describes the lightning payment right?
So why not just return the LnPaymentDetails from these methods?
cc @JssDWt as the issue opener.

roeierez (changes requested):

libs/sdk-core/src/models.rs
JssDWt: The main goal is to make a cleaner api, where the fields returned from an api call describe exactly what you did, and contain all the relevant information. The main issue with the send_payment and send_spontaneous_payment is the `details` field right now. That contains either ClosedChannelPaymentDetails or LnPaymentDetails. It would improve, by returning only LnPaymentDetails, but still it would contain many fields that you don't need, like the lnurl and swap fields. The reason for creating the issue, is to get rid of the ClosedChannelPaymentDetails in the response to a payment.

I do like the idea of having internal structs that represent exactly the information available at any point in the code. That reduces the amount of things you have to keep in mind when developing on this stuff. When you see a struct, you should know what information it contains and not have to scavenge through the codebase looking for potential other uses of the same struct. So I do like the idea of introducing a new struct for this case. LnPaymentDetails can be used as an aggregate. Either for persistence purposes or to include in the payment list. But ideally every subcall has its own response type.

Looking at the 'PendingPayment' struct, it still contains the PaymentDetails, with the LnPaymentDetails/ClosedChannelPaymentDetails. I would like this struct to represent a 'lightning payment' alone, without any wrappers for lnurl/swaps/closed channels whatsoever. So when there has been a lightning payment somewhere in the code, it's clear that at that point we're looking at a lightning payment, and nothing else.

Another comment about the PendingPayment struct is that the name implies it is pending. But it might not be pending. So we'd need to make clear whether the payment is pending or not.

As for the struct exposed to the client, maybe we can fill fields of the SendPaymentResponse itself, or indeed return the PendingPayment struct.

Generally it's a good idea in terms of code isolation to at least use different structs for 
- persistence
- the api you expose
- the api you call
That helps with being able to feely modify the code for specific purposes in different parts of the codebase.

(sorry about the long reply)

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -360,7 +360,7 @@ dictionary LogEntry {
 dictionary InvoicePaidDetails {
     string payment_hash;
     string bolt11;
-    Payment? payment;
+    PaymentListItem? payment;
```
JssDWt, line 363: An InvoicePaid event maybe shouldn't include A PaymentListItem, but only the data relevant to the paid invoice.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -378,7 +378,7 @@ interface BreezEvent {
     NewBlock(u32 block);
     InvoicePaid(InvoicePaidDetails details);
     Synced();
-    PaymentSucceed(Payment details);
+    PaymentSucceed(PaymentListItem details);
```
JssDWt, line 381: Maybe this can be LNPaymentDetails

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -697,7 +697,7 @@ dictionary SendSpontaneousPaymentRequest {
 };
 
 dictionary SendPaymentResponse {
-    Payment payment;
+    PaymentListItem payment;
```
JssDWt, line 700: This should probably contain the PendingPayment

JssDWt (commented):
Added some comments in the UDL as an addition to the comment I left above.

What I do really like about renaming `Payment` to `PaymentListItem`, it is immediately clear where the `Payment` struct was 'abused' before, in the sense that this specific struct is not the right one for every usecase.

libs/sdk-core/src/models.rs
hydra-yse: Not at all @JssDWt, now I have a clearer idea of what you had in mind. I'll try to split everything as best as I can, so let the reviews continue!

libs/sdk-bindings/src/breez_sdk.udl
roeierez: I think both here and in PaymentSucceed event we should contain the same struct/data.
@JssDWt @hydra-yse  how about using the LNPaymentDetails here and there? This way we don't need to introduce another struct (PendingPayment).

JssDWt: Question: How would you feel about extending the PaymentDetails enum with variants for lnurl_pay, ln_address, lnurl_withdraw, swap and reverse swap? And make LnPaymentDetails basically the thing we mean with PendingPayment here?

PR #762, opened 2024-01-28
Ensure homebrew updates its internal list of formulas before installing

ok300:
This PR is a different approach for for #760

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -203,6 +203,7 @@ jobs:
         working-directory: libs/sdk-react-native 
         run: |
           yarn global add tslint typescript
+          brew update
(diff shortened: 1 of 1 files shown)


ok300:
I ran the CI 3 times, the correct (latest) `ktlint` was used every time.

roeierez (approved):
I prefer this approach

dangeross (approved):
Preferred, if it keeps the version across runs consistent. We can always fallback to #760

JssDWt:
Good stuff, I tried this too, but it took more than 6 minutes to complete. It must have been a flake, since this run completed in 1m.

JssDWt (approved):

Issue #726, opened 2024-01-17
Swap indefinitely pending, as long as initial tx is not confirmed


ok300:
As long as the initial swap transaction remains unconfirmed, the SDK
- cannot refund the swap
- cannot complete the swap

This prevents the SDK from performing further swaps.

There is no perfect solution, because as long as the initial tx is in *some* mempools, it can end up being mined at some point in the future. Thus, the SDK has to be ready for that possibility, so it has to keep track of this pending swap.

Some possible solutions for the sender are:
- to bump the fees with RBF
- to use CPFP in order to get the swap tx to be mined sooner
- ... (TBD)

Posting this here to collect feedback and ideas.

kingonly:
https://breez-tech.slack.com/archives/C06330E0S4X/p1705525064790539

roeierez:
Following our discussion:
1. We need to consider swaps as redeemable as long as there are unpaid confirmed funds regardless of the mempool status.
2. We need to add API to rescan all swaps (not only those that are monitored).

PR #764, opened 2024-01-30
Rescan swaps

roeierez:
Two changes here:
1. We no longer requires redeemable swaps to have no unconfirmed funds associated with swap address. If the swap address as confirmed sats and no lightning payment yet then it is redeemable. It allows users, in case of long waiting unconfirmed transaction, initiate another transaction (if the wallet doesn't support RBF) with higher fees.
2. New API added to rescan all swaps.

fixes https://github.com/breez/breez-sdk/issues/726

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -737,6 +737,12 @@ impl BreezServices {
         Ok(None)
     }
 
+    pub async fn rescan_swaps(&self) -> SdkResult<()> {
+        let tip = self.chain_service.current_tip().await?;
+        self.btc_receive_swapper.rescan_swaps(tip).await?;
+        Ok(())
+    }
+
     /// Lookup the reverse swap fees (see [ReverseSwapServiceAPI::fetch_reverse_swap_fees]).
     ///
     /// To get the total estimated fees for a specific amount, specify the amount to be sent in
--- a/libs/sdk-core/src/swap_in/swap.rs
+++ b/libs/sdk-core/src/swap_in/swap.rs
@@ -268,9 +268,16 @@ impl BTCReceiveSwap {
             .ok_or_else(|| anyhow!(format!("Swap address {} was not found", address)))
     }
 
+    pub(crate) async fn rescan_swaps(&self, tip: u32) -> Result<()> {
+        _ = self
+            .refresh_swaps(self.persister.list_swaps()?, tip)
+            .await?;
+        Ok(())
+    }
+
     pub(crate) async fn execute_pending_swaps(&self, tip: u32) -> Result<()> {
         // first refresh all swaps we monitor
-        _ = self.refresh_monitored_swaps(tip).await?;
+        _ = self.refresh_swaps(self.list_monitored()?, tip).await?;
 
         // redeem swaps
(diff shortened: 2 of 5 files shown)


libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -299,9 +306,8 @@ impl BTCReceiveSwap {
         Ok(())
     }
 
-    async fn refresh_monitored_swaps(&self, tip: u32) -> Result<Vec<SwapInfo>> {
-        let to_check = self.list_monitored()?;
-        for s in to_check {
+    async fn refresh_swaps(&self, swaps: Vec<SwapInfo>, tip: u32) -> Result<Vec<SwapInfo>> {
+        for s in swaps {
```
ok300, line 310: Since it's not refreshing the monitored swaps, it should also not return them in L232 below.

Instead it should collect the successfully refreshed swaps (returned in L312) and return them as a vector.

If the list is not used anywhere, it can also just return nothing.

libs/sdk-core/src/breez_services.rs
```diff
@@ -737,6 +737,12 @@ impl BreezServices {
         Ok(None)
     }
 
+    pub async fn rescan_swaps(&self) -> SdkResult<()> {
```
ok300, line 740: This could return the list of updated swaps, to spare the caller another `in_progress_swap`.

Also: does this mean `in_progress_swap` can now return a list? If I understand this correctly, it can now be that any number of previously-closed swaps can become in-progress if they receive sats on those addresses.

roeierez: If we want to return something here then we need also to include the refundables (which may also be updated)
I think since the user has the API to retrieve them we can return nothing as you suggested

Regarding in progress if they get funds after the redeem they become refundables so it should not be a problem

dangeross (commented):
Looks good, still needs UDL changes?

libs/sdk-core/src/swap_in/swap.rs
roeierez: Done

dangeross (commented):
Flutter needs running too

roeierez:
> Flutter needs running too

I felt the commit chunk is too small... :)
Done.

libs/sdk-core/src/breez_services.rs
dangeross, line 740: Maybe good to add a service method description here

dangeross (changes requested):

ok300 (approved):

dangeross (approved):
LGTM

libs/sdk-flutter/lib/bridge_generated.dart
```diff
@@ -2661,6 +2666,22 @@ class BreezSdkCoreImpl implements BreezSdkCore {
         argNames: ["req"],
       );
 
+  Future<void> rescanSwaps({dynamic hint}) {
```
erdemyerebasmaz, line 2669: @roeierez This is not exposed through the Flutter plugin, needs to be added to `BreezSDK`(_sdk-flutter/lib/breez_sdk.dart_).

PR #765, opened 2024-01-30
Allow setting of close_to_addr

### LINKED ISSUE #711: Allow set close_to address
This address is the channel closure address where the funds are swept to.
GL supports this: https://github.com/Blockstream/greenlight/blob/6ab8f3205639ddf611f99e71691f7515fc0d24e6/libs/proto/greenlight.proto#L55C16-L55C24

dangeross:
This PR adds a `set_node_config` SDK method to allow the `close_to_addr` to be configured in Greenlight. This address is then available Greenlight while the client is not present.

https://github.com/Blockstream/blockstream_breez_issues/issues/16#issuecomment-1878702022

Fixes #711

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -176,6 +176,10 @@ dictionary NodeState {
     u64 inbound_liquidity_msats;    
 };
 
+dictionary SetNodeConfigRequest {
+    string? close_to_address;
+};
+
 dictionary SignMessageRequest {
     string message;
 };
@@ -741,6 +745,9 @@ interface BlockingBreezServices {
 
    [Throws=SdkError]
    void disconnect();
+
+   [Throws=SdkError]
+   void set_node_config(SetNodeConfigRequest req);
 
    [Throws=SendPaymentError]
    SendPaymentResponse send_payment(SendPaymentRequest req);
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -238,6 +238,13 @@ impl BreezServices {
         Ok(())
     }
 
+    /// Set the node config
+    ///
+    /// This calls [NodeAPI::set_node_config] to make changes to the active node's configuration.
+    pub async fn set_node_config(&self, req: SetNodeConfigRequest) -> SdkResult<()> {
+        Ok(self.node_api.set_node_config(req.close_to_address).await?)
+    }
+
     /// Pay a bolt11 invoice
     ///
     /// Calling `send_payment` ensures that the payment is not already completed, if so it will result in an error.
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -651,6 +651,19 @@ impl NodeAPI for Greenlight {
         .map(|credentials| NodeCredentials::Greenlight { credentials }))
     }
 
+    async fn set_node_config(&self, close_to_address: Option<String>) -> NodeResult<()> {
+        let mut client = self.get_client().await?;
+        if let Some(close_to_addr) = close_to_address {
+            client
+                .configure(gl_client::pb::GlConfig { close_to_addr })
+                .await
+                .map_err(|e| {
+                    NodeError::Generic(anyhow!("Unable to set close to address: {}", e))
+                })?;
+        }
+        Ok(())
+    }
+
     async fn create_invoice(
         &self,
         amount_msat: u64,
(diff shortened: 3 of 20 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -651,6 +651,19 @@ impl NodeAPI for Greenlight {
         .map(|credentials| NodeCredentials::Greenlight { credentials }))
     }
 
+    async fn set_node_config(&self, close_to_address: Option<String>) -> NodeResult<()> {
+        let mut client = self.get_client().await?;
+        if let Some(close_to_addr) = close_to_address {
+            client
+                .configure(gl_client::pb::GlConfig { close_to_addr })
```
ok300, line 658: Is this where validation happens?

If a user were to make a typo or give an invalid address, is this where it would fail?

ok300: If not, would it make sense to add our own
```rust
ensure_sdk!(
    bitcoin::util::address::Address::from_str(&close_to_addr).is_ok(),
    // SdkError
);
```

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -742,6 +746,9 @@ interface BlockingBreezServices {
    [Throws=SdkError]
    void disconnect();
 
+   [Throws=SdkError]
+   void set_node_config(SetNodeConfigRequest req);
```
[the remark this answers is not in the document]
dangeross: Getting might be possible using `datastore` and unwrapping/serialising the stores request data. Maybe simpler to change the naming

libs/sdk-core/src/greenlight/node_api.rs
dangeross: I don't think the address is validated in the gl-plugin. Probably a good idea to add validation.

dangeross: Ok I checked, there is validation: https://github.com/Blockstream/greenlight/blob/6ab8f3205639ddf611f99e71691f7515fc0d24e6/libs/gl-plugin/src/node/mod.rs#L465-L509

libs/sdk-bindings/src/breez_sdk.udl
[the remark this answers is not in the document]
dangeross: Done

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -651,6 +651,19 @@ impl NodeAPI for Greenlight {
         .map(|credentials| NodeCredentials::Greenlight { credentials }))
     }
 
+    async fn set_node_config(&self, close_to_address: Option<String>) -> NodeResult<()> {
+        let mut client = self.get_client().await?;
+        if let Some(close_to_addr) = close_to_address {
```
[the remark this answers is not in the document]
dangeross: Done

ok300 (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -238,6 +238,13 @@ impl BreezServices {
         Ok(())
     }
 
+    /// Configure the node
+    ///
+    /// This calls [NodeAPI::configure_node] to make changes to the active node's configuration.
+    pub async fn configure_node(&self, req: ConfigureNodeRequest) -> SdkResult<()> {
```
JssDWt, line 244: I'm thinking about the case where there's multiple apps integrating the sdk. And each app has a different opinion on what the close address should be. They might pingpong to set this close address to a different value each time. 

Perhaps a comment saying this method should ideally just be called once in the lifetime of the node? You can do it on creation of the node. But ideally not after each `connect`.

tools/sdk-cli/src/commands.rs
```diff
@@ -205,6 +205,13 @@ pub(crate) enum Commands {
     /// The up to date node information
     NodeInfo {},
 
+    /// Configurate the node
+    ConfigureNode {
+        // Optional address to send funds to during a channel close
```
JssDWt, line 210: I think it's good to identify this is for mutual closes. At the moment most closes are probably force closes, and the address will not be used there.

```suggestion
        // Optional address to send funds to during a mutual channel close
```

JssDWt (commented):
Just two comments about comments.

libs/sdk-core/src/breez_services.rs
dangeross: Done [fb7e777](https://github.com/breez/breez-sdk/pull/765/commits/fb7e777d8aaa9b7783d038baf759200c44256a81)

tools/sdk-cli/src/commands.rs
dangeross: Done [fb7e777](https://github.com/breez/breez-sdk/pull/765/commits/fb7e777d8aaa9b7783d038baf759200c44256a81)

roeierez (approved):
LGTM

PR #766, opened 2024-01-31
Rename SwapInfo paid_sats to paid_msat

### LINKED ISSUE #746: Bug: `SwapInfo` includes msat amount in field named `paid_sats`
As the title describes, I found that the swap info struct being returned from `list_payments` provides the paid msat amount instead of the sat amount. 

After a quick inspection of the source code, indeed, the msat amount is being persisted in the db into a column named `paid_sats`. 

It should be a trivial fix. It's just a matter of deciding if it makes sense to fix it by persisting the sat amount or changing the column name and variables to `paid_msats`.

dangeross:
Rename the column to fix millisatoshi payment amount stored to sats column.

Fixes #746

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -1238,7 +1238,7 @@ pub struct SwapInfo {
 
     /// bolt11 invoice to claim the sent funds.
     pub bolt11: Option<String>,
-    /// Amount of millisatoshis paid to the claim address.
+    /// Amount of millisatoshis claimed from sent funds and paid for via bolt11 invoice.
     pub paid_msat: u64,
     /// Confirmed onchain sats to be claim with an bolt11 invoice or refunded if swap fails.
     pub confirmed_sats: u64,
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -1238,8 +1238,8 @@ pub struct SwapInfo {
 
     /// bolt11 invoice to claim the sent funds.
     pub bolt11: Option<String>,
-    /// Amount of sats payed to the claim address.
-    pub paid_sats: u64,
+    /// Amount of millisatoshis paid to the claim address.
+    pub paid_msat: u64,
     /// Confirmed onchain sats to be claim with an bolt11 invoice or refunded if swap fails.
     pub confirmed_sats: u64,
     /// Unconfirmed sats waiting to be confirmed onchain.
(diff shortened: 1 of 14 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -1238,8 +1238,8 @@ pub struct SwapInfo {
 
     /// bolt11 invoice to claim the sent funds.
     pub bolt11: Option<String>,
-    /// Amount of sats payed to the claim address.
-    pub paid_sats: u64,
+    /// Amount of millisatoshis paid to the claim address.
+    pub paid_msat: u64,
```
ok300, line 1242: SDK requests / responses refering to LN amounts are using `msat`.

Those refering to onchain amounts are now using `sat`, for example the other `SwapInfo` fields

https://github.com/breez/breez-sdk/blob/1cffacb3e855360ad0e87fc93397c9f6c52f0585/libs/sdk-core/src/models.rs#L1243-L1246

or `ReverseSwapInfo`

https://github.com/breez/breez-sdk/blob/1cffacb3e855360ad0e87fc93397c9f6c52f0585/libs/sdk-core/src/models.rs#L160-L163

Wouldn't it then make more sense to keep using `sat` here instead of `msat`?

dangeross: In the issue it was suggested that the paid amount (over lightning) should be stored as millisatoshis

ok300: You're right.

> Amount of millisatoshis paid to the claim address
> pub paid_msat: u64

Its not clear if this means
- "amount which left the LN wallet" (LN, should be msat) or
- "amount which arrived in the claim address" (onchain, should be sat)

Turns out its the former. The swap field is written in `update_swap_paid_amount` as the LN invoice msat amount:

https://github.com/breez/breez-sdk/blob/be1915da8bf066042f5adc686112cf229f55f0fe/libs/sdk-core/src/swap_in/swap.rs#L118-L126

roeierez (approved):
LGTM

libs/sdk-core/src/models.rs
dangeross: >    Amount of millisatoshis claimed from sent funds and paid for via bolt11 invoice.
>    pub paid_msat: u64,

Improved the field description [9ba9dcf](https://github.com/breez/breez-sdk/pull/766/commits/9ba9dcf5eb4a1e8f2ce580afc74ea070f01fb2fc)

ok300: Thanks, that's indeed clearer.

ok300 (approved):

PR #771, opened 2024-02-02
Fix rustdoc warnings

ok300:
Fix warnings that prevent links from being parsed as such.

For example, the rustdoc description of [SendPaymentError](https://breez.github.io/breez-sdk/breez_sdk_core/error/enum.SendPaymentError.html) is

> Error returned by [BreezServices::send_payment] and [BreezServices::send_spontaneous_payment]

where none of these references are clickable links.


dangeross (approved):
LGTM

PR #772, opened 2024-02-05
Add SDK support for swap tx notifications

ok300:
This PR adds support for swap tx notifications.

The workflow is as follows:

- when the caller wishes to create a new swap
  - the caller calls `swap_info = sdk.receive_onchain()`
  - the caller builds `webhook_url` with the right NDS, token, platform, etc.
  - the caller calls `sdk.register_swap_tx_webhook(webhook_url, swap_info.bitcoin_address)`
  - the SDK registers a callback on the chain notifier, for the specific bitcoin address
- when a tx to the registered bitcoin address is confirmed, the chain notifier triggers a callback to the `webhook_url`
- the NDS processes the received callback and triggers a notification


roeierez:
@ok300 writing here for us:

As me and Carlos has discusses we decided to:
1. Reuse the same register_webhook for onchain notifications as well by storing the webhook in the db.
2. On every swap address creation the SDK registers automatically.

libs/sdk-core/src/swap_in/notification.rs
```diff
@@ -0,0 +1,33 @@
+use reqwest::Body;
+use serde_json::json;
+use tonic::codegen::http::header::CONTENT_TYPE;
+
+use crate::error::{SdkError, SdkResult};
+use crate::input_parser::get_reqwest_client;
+
+/// Registers for a swap tx notification. When a new transaction to the specified `swap_address`
+/// is confirmed, a callback will be triggered to the specified `webhook_url`.
+pub(crate) async fn register_swap_tx_webhook(
```
roeierez, line 10: Can we make this function only accessible to the swapper?

roeierez (changes requested):

libs/sdk-core/src/swap_in/notification.rs
ok300: Done: 91e169eaed1e727d4e7e5796ee9fa978b719fce4

libs/sdk-core/src/breez_services.rs
```diff
@@ -730,6 +730,18 @@ impl BreezServices {
             .btc_receive_swapper
             .create_swap_address(channel_opening_fees)
             .await?;
+
+        if let Some(webhook_url) = self.persister.get_webhook_url()? {
```
roeierez, line 734: If this is part of the btc_receive_swapper and always executed as part of creating swap address perhaps this logic should be inside the btc_receive_swapper?

ok300: Good point, I moved it in `create_swap_address`: c35fb43b6b168fa76b7091465eefee2fae289b7f

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -493,6 +502,21 @@ impl BTCReceiveSwap {
             .await?;
         get_utxos(address, transactions)
     }
+
+    /// Registers for a swap tx notification. When a new transaction to the specified `swap_address`
+    /// is confirmed, a callback will be triggered to the specified `webhook_url`.
+    pub(crate) async fn register_swap_tx_notification(
```
roeierez, line 508: This still has create visibility. Perhaps we don't need this function at all and can invoke the super::`register_swap_tx_notification` directly from the `create_swap_address` function?

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -206,6 +211,10 @@ impl BTCReceiveSwap {
 
         // persist the swap info
         self.persister.insert_swap(swap_info.clone())?;
+
+        self.register_swap_tx_notification(chainnotifier_url, &swap_info.bitcoin_address)
+            .await?;
```
JssDWt, line 216: I would put this registration in breezservices.
Keep the swapper code clean. create_swap_address means creating a swap address.
The breezservices layer is the application layer, where I think this kind of registration is most suitable.

roeierez: Sounds good to me. My main feedback was actually about the visibility of the register_swap_tx_notification which if we move the code to breezservices can be private there.

ok300:
I moved the call to `BreezServices` and made the method private there.

For this, the actual implementation in `swap_in/mod.rs` had to get `pub(super)` visibility.

Is this what you had in mind?

roeierez (changes requested):

libs/sdk-core/src/swap_in/mod.rs
```diff
@@ -1,2 +1,34 @@
...
 pub(crate) mod swap;
+
+use reqwest::Body;
+use serde_json::json;
+use tonic::codegen::http::header::CONTENT_TYPE;
+
+use crate::error::{SdkError, SdkResult};
+use crate::input_parser::get_reqwest_client;
+
+pub(super) async fn register_swap_tx_notification(
```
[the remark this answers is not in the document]
ok300: Done: a65e7cebbeffb00669d7679b8277bc858ac9d8ed

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -133,6 +133,9 @@ impl BTCReceiveSwap {
     }
 
     /// Create a [SwapInfo] that represents the details of an on-going swap.
+    ///
```
roeierez, line 136: This comment seems redundant now.

roeierez (changes requested):

JssDWt:
Looks good.
Let's squash those commits a little so the history remains clean.

libs/sdk-core/src/swap_in/swap.rs
ok300: Done, removed it: 861d688edd62745e45be90b0106dd2d481a8b988

PR #776, opened 2024-02-07
Update docs for swap channel fees

ok300:
Added clarification that `SwapInfo::channel_opening_fees` are always set.


dangeross (approved):
LG, but needs the flutter gen running.

PR #779, opened 2024-02-08
Re-use bitcoin crate exported by gl-client

ok300:
This PR removes the explicit dependency on `bitcoin` in our `Cargo.toml`. Instead of it, we use the version re-exported by GL at `gl_client::bitcoin`.

This helps us avoid having to track and re-align the `bitcoin` version with each `gl_client` update.

### DIFF
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -518,7 +518,6 @@ dependencies = [
  "anyhow",
  "base64 0.13.1",
  "bip21",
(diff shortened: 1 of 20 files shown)


JssDWt:
When is this an issue? When gl_client updates their bitcoin version? Only on minor version upgrades, or any upgrade?

JssDWt (approved):
LGTM, but madness that this is necessary ofcourse.

PR #780, opened 2024-02-08
remove deprecated fields from lsp info

JssDWt:
The channel_capacity and target_conf have no meaning in LspInformation. This PR removes those fields. They will be removed from lspd as well.

Also deprecated/unused fields in LSPInformation are now removed from the breez.proto file.


ok300 (approved):

roeierez (approved):
LGTM

PR #782, opened 2024-02-08
Ensure there is at least one routing hint when creating invoice

roeierez:
This PR do a bit of refactor of the part adding the routing hints to the invoice.
The motivation for this was a real case when a user had private channels with a node that is not the lsp (same case as changing lsp after some channels were opened).
Creating an invoice for that user resulted without routing hints. The invoice returned from greenlight without routing hints and we didn't add routing hints because our previous logic only added hints for the current lsp and the user still didn't have a channel with the current lsp...

The logic here has been simplified to this:
1. If we need to open a channel the routing hints are only the lsp with the zero channel id
2. If we don't need to open a channel we: 
- use all private channels of the node to create the routing hints array (one channel per peer).
- We add to the invoice only the routing hints for peers that don't exist already in the hints returned from gl.

In addition, before using the node api to create the invoice we check if the user has routing hint candidates (private channels) and if not we return an early error.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -2095,7 +2095,8 @@ impl Receiver for PaymentReceiver {
         // check if the lsp hint already exists
         info!("Existing routing hints {:?}", parsed_invoice.routing_hints);
 
-        // We only create a new invoice if we need to add the lsp hint or change the amount
+        // We only create a new invoice if we need to add routing hints because the invoice
+        // has not routing hints or change the amount due to openning a channel request.
         if parsed_invoice.routing_hints.is_empty() || open_channel_needed {
             // create the large amount invoice
             let raw_invoice_with_hint = add_routing_hints(
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -2090,7 +2090,7 @@ impl Receiver for PaymentReceiver {
             optional_lsp_hint,
             parsed_invoice.routing_hints.is_empty(),
         ) {
-            // If we need to open a channel we only need to set the decicated lsp hint.
+            // If we need to open a channel we only need to set the dedicated lsp hint.
             (true, _, _) => {
                 let open_channel_hint = RouteHint {
                     hops: vec![RouteHintHop {
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -2051,30 +2051,16 @@ impl Receiver for PaymentReceiver {
                 // remove the fees from the amount to get the small amount on the current node invoice.
                 destination_invoice_amount_msat = req.amount_msat - channel_fees_msat;
             }
-
-            let lsp_hop = RouteHint {
-                hops: vec![RouteHintHop {
-                    src_node_id: lsp_info.pubkey,
-                    short_channel_id: parse_short_channel_id("1x0x0")?,
-                    fees_base_msat: lsp_info.base_fee_msat as u32,
-                    fees_proportional_millionths: (lsp_info.fee_rate * 1000000.0) as u32,
-                    cltv_expiry_delta: lsp_info.time_lock_delta as u64,
-                    htlc_minimum_msat: Some(lsp_info.min_htlc_msat as u64),
-                    htlc_maximum_msat: None,
-                }],
-            };
-            info!("Adding LSP hop as routing hint: {:?}", lsp_hop);
-            routing_hints.push(lsp_hop)
         } else {
             // not opening a channel so we need to get all private channels into the routing hints
-            info!("Finding channel ID for routing hint");
-            routing_hints = self.node_api.get_routing_hints().await?;
-        }
-
-        if routing_hints.is_empty() {
-            return Err(ReceivePaymentError::InvoiceNoRoutingHints {
-                err: "Must be connected to at list one peer".into(),
-            });
+            info!("Getting routing hints from node");
+            let (hints, has_public_channel) = self.node_api.get_routing_hints().await?;
+            if !has_public_channel && hints.is_empty() {
+                return Err(ReceivePaymentError::InvoiceNoRoutingHints {
+                    err: "Must have at least one active channel".into(),
+                });
+            }
+            routing_hints = hints;
         }
 
         info!("Creating invoice on NodeAPI");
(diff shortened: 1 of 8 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -2052,33 +2051,30 @@ impl Receiver for PaymentReceiver {
...
-                    };
+            routing_hints = self.node_api.get_routing_hints().await?;
+        }
 
-                    break;
-                }
-            }
+        if routing_hints.is_empty() {
+            return Err(ReceivePaymentError::InvoiceNoRoutingHints {
+                err: "Must be connected to at list one peer".into(),
```
ok300, line 2076:
```suggestion
                err: "Must be connected to at least one peer".into(),
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1435,6 +1435,67 @@ impl NodeAPI for Greenlight {
...
+        let mut hints: Vec<RouteHint> = vec![];
+        let mut node_client = self.get_node_client().await?;
+        let peers = node_client
+            .list_peers(cln::ListpeersRequest::default())
+            .await?
+            .into_inner();
+
+        // For each peer look for a private channel and create a routing hint from it.
+        for peer in peers.peers {
+            if !peer.channels.is_empty() {
```
ok300, line 1450: IMO no need to check for this, because if the list is empty, the iterator below will just not iterate.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2052,33 +2051,30 @@ impl Receiver for PaymentReceiver {
...
-
-                    short_channel_id = match hint {
-                        None => None,
-                        Some(hint) => {
-                            let scid = parse_short_channel_id(&hint)?;
-                            info!("Found channel ID: {scid:?} {active_channel:?}");
-                            Some(scid)
-                        }
-                    };
+            routing_hints = self.node_api.get_routing_hints().await?;
```
ok300, line 2071: The log statement above is not relevant anymore, as this is not finding the channel ID anymore.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1435,6 +1435,67 @@ impl NodeAPI for Greenlight {
         debug!("send_custom_message returned status {:?}", resp.status);
         Ok(())
     }
+
+    // Gets the routing hints related to all private channels that the node has
+    async fn get_routing_hints(&self) -> NodeResult<Vec<RouteHint>> {
```
JssDWt, line 1440: This can potentially return many routing hints. I think we'd want to select a maximum of 3 hints. Probably ordered by the liquidity available (if possible). 

Note that a route hint is (usually) not useful if the peer only has private channels as well. So you might want to select nodes with active public channels in the lightning graph.

JssDWt (commented):
Adding one comment now.
There's probably some nits, which I'll ignore for now.
I still have to check the `add_routing_hints` logic, which has become a bit hard to follow. 
Otherwise looks good.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Do you see any real issue by leaving the logic trivial?  I don't want to complicate it if not really necessary.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1435,6 +1435,67 @@ impl NodeAPI for Greenlight {
...
+        let mut hints: Vec<RouteHint> = vec![];
+        let mut node_client = self.get_node_client().await?;
+        let peers = node_client
+            .list_peers(cln::ListpeersRequest::default())
```
JssDWt, line 1444: This should be `list_peer_channels`. `list_peers` might also returns channels, but that functionality is deprecated and will be removed soon. (is already removed if deprecated apis are not enabled)

libs/sdk-core/src/invoice.rs
```diff
@@ -145,15 +145,20 @@ impl RouteHint {
     }
 }
 
-pub fn add_lsp_routing_hints(
+pub fn add_routing_hints(
     invoice: String,
-    include_route_hints: bool,
-    lsp_hint: Option<RouteHint>,
+    merge_with_existing: bool,
```
JssDWt, line 150: If there are route hints in the invoice already, and we don't have to replace the route hints because there is no channel needed, do we have to merge in more route hints? Can we keep the ones selected by greenlight and be done with it? Only add route hints if there are none in that case?

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: Not necessarily. Chances are high greenlight will already have hop hints in the invoice. Just a limit of 2 or 3 hints would be good.

JssDWt (commented):
2 more comments

EDIT: that don't show up here...

libs/sdk-core/src/greenlight/node_api.rs
dangeross: I think the issue is GL is not reliably adding the hop hints, especially when creating an invoice very soon after connecting

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1435,6 +1435,67 @@ impl NodeAPI for Greenlight {
...
+            .await?
+            .into_inner();
+
+        // For each peer look for a private channel and create a routing hint from it.
+        for peer in peers.peers {
+            if !peer.channels.is_empty() {
+                let optional_opened_channel = peer
+                    .channels
+                    .iter()
+                    .find(|&c| c.state == cln::ChannelState::ChanneldNormal as i32);
```
dangeross, line 1454: This should also check the channel is private?

libs/sdk-core/src/invoice.rs
```diff
@@ -166,28 +171,35 @@ pub fn add_lsp_routing_hints(
     // We make sure the hint we add does not conflict with other hints.
     // The lsp hint takes priority so in case the lsp hop is already in one of the existing hints
     // We make sure not to include them in the new hints.
```
dangeross, line 173:
```suggestion
    // We make sure the route hints we want to add do not conflict with invoice route hints.
    // When merging route hints, only route hints are added that go through different nodes than ones in the invoice route hints.
    // Otherwise when not merging route hints, the invoice route hints are replaced by the provided route hints.
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -2096,57 +2092,23 @@ impl Receiver for PaymentReceiver {
...
-            info!("Routing hint added");
-            let signed_invoice_with_hint = self.node_api.sign_invoice(raw_invoice_with_hint)?;
-            info!("Signed invoice with hint = {}", signed_invoice_with_hint);
+        info!("Routing hint added");
+        let signed_invoice_with_hint = self.node_api.sign_invoice(raw_invoice_with_hint)?;
+        info!("Signed invoice with hint = {}", signed_invoice_with_hint);
 
-            parsed_invoice = parse_invoice(&signed_invoice_with_hint)?;
-        }
+        parsed_invoice = parse_invoice(&signed_invoice_with_hint)?;
```
dangeross, line 2111: I would suggest trying to ascertain if the routing hints have changed and only signing a new invoice if its the case. Else we're always creating two invoices, even if the route hints are no different to the GL ones.

ok300:
Fixed conflicts with `main`.

libs/sdk-core/src/invoice.rs
```diff
@@ -167,28 +172,35 @@ pub fn add_lsp_routing_hints(
...
-        None => invoice.route_hints(),
-        Some(lsp_hint) => match include_route_hints {
+    let unique_hop_hints: Vec<lightning::routing::router::RouteHint> = match route_hints.len() {
+        0 => invoice.route_hints(),
+        _ => match merge_with_existing {
             true => {
-                let mut all_hints: Vec<lightning::routing::router::RouteHint> = invoice
-                    .route_hints()
-                    .into_iter()
+                let unique_to_add: Vec<&RouteHint> = route_hints
```
ok300, line 179: Maybe a better way to find `unique_to_add` is

```rust
let invoice_hints_hop_src_node_ids: Vec<String> = invoice
    .route_hints()
    .into_iter()
    .flat_map(|hint| hint.0)
    .map(|hop| hop.src_node_id.serialize().encode_hex::<String>())
    .collect();
let unique_to_add: Vec<&RouteHint> = route_hints
    .iter()
    .filter(|hint| {
        hint.hops
            .iter()
            .all(|hop| !invoice_hints_hop_src_node_ids.contains(&hop.src_node_id))
    })
    .collect();
```

Avoids the 4-level-deep nesting in the filter, which is error-prone.

libs/sdk-core/src/breez_services.rs
roeierez: fixed

libs/sdk-core/src/breez_services.rs
roeierez: fixed

libs/sdk-core/src/breez_services.rs
roeierez: done

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Added limit

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Addressed

libs/sdk-core/src/invoice.rs
roeierez: changed int the caller scope to only include lsp hint if exists.

libs/sdk-core/src/invoice.rs
roeierez: Done, thank you.

JssDWt (approved):
Looks good.

ok300 (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -2096,56 +2078,68 @@ impl Receiver for PaymentReceiver {
...
+        // limit the hints to max 3 and extract the lsp one.
+        let optional_lsp_hint = Self::limit_and_extract_lsp_hint(&mut routing_hints, &lsp_info);
+
+        // We here check if we need to modify the invoice.
+        let optional_modified_invoice = match (
+            open_channel_needed,
+            optional_lsp_hint,
+            parsed_invoice.routing_hints.is_empty(),
+        ) {
+            // If we need to open a channel we only need to set the decicated lsp hint.
```
dangeross, line 2093:
```suggestion
            // If we need to open a channel we only need to set the dedicated lsp hint.
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -2096,56 +2078,68 @@ impl Receiver for PaymentReceiver {
...
-                req.amount_msat,
-            )?;
-
-            info!("Routing hint added");
-            let signed_invoice_with_hint = self.node_api.sign_invoice(raw_invoice_with_hint)?;
-            info!("Signed invoice with hint = {}", signed_invoice_with_hint);
+            // In case we don't need to open a channel and we have a channel with our lsp then we only ensure it
+            // exists as part of the invoice routing hints (merging).
+            (false, Some(h), _) => {
+                // In case we need to open a channel and the invoice has no routing hints we replace them with ours.
```
dangeross, line 2117:
```suggestion
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -2096,56 +2078,68 @@ impl Receiver for PaymentReceiver {
...
+                )?)
+            }
+            (_, _, _) => None,
+        };
 
-            parsed_invoice = parse_invoice(&signed_invoice_with_hint)?;
+        if let Some(raw_invoice) = optional_modified_invoice {
+            let signed_invoice = self.node_api.sign_invoice(raw_invoice)?;
+            info!("Signed invoice with hint = {}", signed_invoice);
+            parsed_invoice = parse_invoice(&signed_invoice)?;
```
dangeross, line 2142: I think for the normal case (not opening channel, have channel with lsp, have GL lsp route hint), this will still result in 2 invoices every time

roeierez: Good catch. Fixed

libs/sdk-core/src/breez_services.rs

libs/sdk-core/src/breez_services.rs

dangeross (approved):
LGTM!

PR #783, opened 2024-02-08
Link bindings dynamically in iOS Flutter plugin

cnixbtc:
Due to an [issue with `flutter_rust_bridge`](https://github.com/fzyzcjy/flutter_rust_bridge/issues/1712), we need to link our bindings dynamically in the iOS Flutter plugin. [Inspired by BDK,](https://github.com/LtbLightning/bdk-flutter/blob/main/ios/bdk_flutter.podspec#L24) @roeierez found [a way to make that work](https://github.com/breez/breez-sdk-flutter/compare/main...dloader#diff-2a95d4193310f4a1bc04106bce8d8c6b3e712ce52a0864b10d0e80ee21335bbbR33). This PR adds this change.

The idea is to:

1. Download `breez_sdkFFI.xcframework` on `pod install` (we cannot bundle it due to size constraints with pub.dev).
2. Deliver `BreezSDK.swift` together with the plugin the same way we do it for local development

That way we achieve:

- The ability to link the bindings dynamically and use native Swift code as a Flutter plugin user the same way we do it in c-breez
- Have an easy way to switch between development and production in c-breez

## Testing

I tried to simulate locally what the CI would do and pushed that to: [`breez-sdk-flutter/tree/dynamic-linking-test`](https://github.com/breez/breez-sdk-flutter/tree/dynamic-linking-test). Running c-breez and pulling the breez-sdk from this branch works on my machine.

Would appreciate if anyone could double check this.

### DIFF
--- a/.github/workflows/publish-flutter.yml
+++ b/.github/workflows/publish-flutter.yml
@@ -35,7 +35,6 @@ jobs:
           ssh-key: ${{ secrets.REPO_SSH_KEY }}
           fetch-depth: 0
           path: flutter
-          ref: test-ios-dynamic-linking
 
       - name: Checkout breez-sdk repo
         uses: actions/checkout@v3
@@ -89,11 +88,10 @@ jobs:
         working-directory: flutter
         if: ${{ inputs.publish }}
         run: |
-          [[ $(git branch --show-current) == "test-ios-dynamic-linking" ]] || return 1
           git config --global user.email github-actions@github.com
           git config --global user.name github-actions
           git add .
           git commit -m "Update Breez SDK Flutter package to version v${{ inputs.package-version }}"
           git push
-          # git tag v${{ inputs.package-version }} -m "v${{ inputs.package-version }}"
-          # git push --tags
+          git tag v${{ inputs.package-version }} -m "v${{ inputs.package-version }}"
+          git push --tags
--- a/.github/workflows/publish-flutter.yml
+++ b/.github/workflows/publish-flutter.yml
@@ -35,6 +35,7 @@ jobs:
           ssh-key: ${{ secrets.REPO_SSH_KEY }}
           fetch-depth: 0
           path: flutter
+          ref: test-ios-dynamic-linking
 
       - name: Checkout breez-sdk repo
         uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


roeierez:
@cnixbtc this looks good.
I think we are missing the copy the BreezSDK.swift file to the flutter plugin don't we?

cnixbtc:
> @cnixbtc this looks good.
> 
> I think we are missing the copy the BreezSDK.swift file to the flutter plugin don't we?

My plan was to download the file in CI (tagged you in an inline comment at the right place). If you agree with the approach I'd do one more round of testing where you run the CI but change it so that it pushes to a fork of the `breez-sdk-flutter` repo. Then we can test integrating into c-breez one more time.

.github/workflows/publish-flutter.yml
```diff
@@ -61,12 +61,17 @@ jobs:
           cp ../breez-sdk/libs/sdk-flutter/pubspec.yaml .
           cp ../breez-sdk/libs/sdk-flutter/pubspec.lock .
 
+      - uses: actions/download-artifact@v3
```
cnixbtc, line 64: @roeierez Here's where the source file is downloaded

roeierez: Ah.. I see, makes sense.
I agree with this approach.

roeierez:
@cnixbtc Thanks for the update.
I do want to release tomorrow a new SDK version and I think that the last problem you mentioned is in c-breez not in your SDK changes.
How about merging this so we can release and fix c-breez (in mode production) later on?

cnixbtc:
> @cnixbtc Thanks for the update.
> 
> I do want to release tomorrow a new SDK version and I think that the last problem you mentioned is in c-breez not in your SDK changes.
> 
> How about merging this so we can release and fix c-breez (in mode production) later on?

Yeah that probably works. Just note that I only tested in c-breez so far though. But I think it should be fine.

Marked as ready for review. Feel free to merge!

libs/sdk-flutter/lib/native_toolkit.dart
```diff
@@ -16,8 +17,11 @@ BreezSdkCore getNativeToolkit() {
       // On Linux the lib needs to be in LD_LIBRARY_PATH or working directory
       _breezSDK = BreezSdkCoreImpl(DynamicLibrary.open(_libName));
     } else if (Platform.isIOS || Platform.isMacOS) {
-      // iOS and macOS are statically linked
-      _breezSDK = BreezSdkCoreImpl(DynamicLibrary.process());
+      try {
+        _breezSDK = BreezSdkCoreImpl(DynamicLibrary.open("$_iosLibName.framework/$_iosLibName"));
```
erdemyerebasmaz, line 21: I may have overlooked something but this does not look right to me. Where does `"breez_sdk.framework/breez_sdk"` point to?

erdemyerebasmaz (commented):
Have a question regarding library path.

libs/sdk-flutter/lib/native_toolkit.dart
roeierez: For production we moved to dynamic loading where the pod download the framework on install.
iOS then expand this dynamic linked framework like this.

roeierez (approved):

PR #784, opened 2024-02-08
Handle `redeem_swap` error in case the invoice is known

### LINKED ISSUE #710: Better handle swap `last_redeem_error`: "preimage already used"
A customer reported seeing "preimage already used" in a swap's `last_redeem_error`.

This error message comes from GL, and it's thrown when we try to create an invoice with an already known preimage: https://github.com/ElementsProject/lightning/blob/a59dbbdae5b8f939d378edd18214a591bef79a74/lightningd/invoice.c#L875-L881

If the previous redeem failed, it's unlikely this was the cause. The cause must have happened before, causing us to try to create the invoice again in the next `execute_pending_swaps` attempt.

We should therefore better identify the root cause of this failure and put that in `last_redeem_error`, before we try to re-create the invoice.

ok300:
This PR attempts to fix the issue in the following scenario:

- user starts a swap on device A
- user pays onchain amount and it confirms, but for one reason or another the invoice cannot be paid (for example, the amount was below `min_allowed_deposit`)
- user restores on device B
- there, the ongoing swap sees a confirmed tx and attempts to create an invoice, but fails because GL cannot create another invoice with the same preimage

The main change brought by the PR is in `redeem_swap`:
- if no invoice is known, try to create it on GL
- if invoice creation fails with `InvoicePreimageAlreadyExists`, try to retrieve it from DB

Fixes #710

---

~The issue is this invoice doesn't exist on device B. It exists on GL, but we only `sync()` settled invoices.~

Update: as per @dangeross 's suggestion, I added a `NodeAPI` method to fetch the invoice from GL directly.

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -708,6 +708,25 @@ impl NodeAPI for Greenlight {
         Ok(res.bolt11)
     }
 
+    async fn lookup_bolt11(&self, payment_hash: Vec<u8>) -> NodeResult<Option<String>> {
+        let request = cln::ListinvoicesRequest {
+            payment_hash: Some(payment_hash),
+            ..Default::default()
+        };
+
+        let found_bolt11 = self
+            .get_node_client()
+            .await?
+            .list_invoices(request)
+            .await?
+            .into_inner()
+            .invoices
+            .first()
+            .cloned()
+            .and_then(|res| res.bolt11);
+        Ok(found_bolt11)
+    }
+
     // implement pull changes from greenlight
     async fn pull_changed(
         &self,
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -708,7 +708,7 @@ impl NodeAPI for Greenlight {
         Ok(res.bolt11)
     }
 
-    async fn lookup_bolt11(&self, payment_hash: Vec<u8>) -> NodeResult<Option<String>> {
+    async fn fetch_bolt11(&self, payment_hash: Vec<u8>) -> NodeResult<Option<String>> {
         let request = cln::ListinvoicesRequest {
             payment_hash: Some(payment_hash),
             ..Default::default()
--- a/libs/sdk-core/src/swap_in/swap.rs
+++ b/libs/sdk-core/src/swap_in/swap.rs
@@ -276,24 +276,16 @@ impl BTCReceiveSwap {
         // redeem swaps
         let redeemable_swaps = self.list_redeemables()?;
         for s in redeemable_swaps {
-            let redeem_res = self.redeem_swap(s.bitcoin_address.clone()).await;
+            let swap_address = s.bitcoin_address;
+            let bolt11 = s.bolt11.unwrap_or_default();
 
-            if redeem_res.is_err() {
-                let err = redeem_res.as_ref().err().unwrap();
-                error!(
-                    "failed to redeem swap {:?}: {} {}",
-                    err,
-                    s.bitcoin_address,
-                    s.bolt11.unwrap_or_default(),
-                );
-                self.persister
-                    .update_swap_redeem_error(s.bitcoin_address, err.to_string())?;
-            } else {
-                info!(
-                    "succeed to redeem swap {:?}: {}",
-                    s.bitcoin_address,
-                    s.bolt11.unwrap_or_default()
-                )
+            match self.redeem_swap(swap_address.clone()).await {
+                Ok(_) => info!("succeed to redeem swap {swap_address}: {bolt11}"),
+                Err(err) => {
+                    error!("failed to redeem swap {err:?}: {swap_address} {bolt11}");
+                    self.persister
+                        .update_swap_redeem_error(swap_address, err.to_string())?;
+                }
             }
         }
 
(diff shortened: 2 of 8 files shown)


dangeross:
One option would be to just call list_invoices with a payment_hash filter to retrieve it from GL.

ok300:
The PR is ready for review.

I added a few other fixes and smaller refactorings. I separated them in individual commits, for easier review.

I tested both the success and failure cases, both were handled correctly.

JssDWt (commented):
just a little nit

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -393,49 +394,68 @@ impl BTCReceiveSwap {
...
+
+                        // Making sure the created invoice amount matches the on-chain amount minus opening channel opening fees
+                        let expected_invoice_amount_msat = swap_info.confirmed_sats * 1_000
+                            - create_invoice_response.opening_fee_msat.unwrap_or_default();
+                        let invoice_amount_msat = invoice.amount_msat.unwrap_or_default();
+
+                        ensure_sdk!(
+                            expected_invoice_amount_msat == invoice_amount_msat,
+                            anyhow!("Invoice amount {invoice_amount_msat} msat doesn't match expected {expected_invoice_amount_msat} msat")
+                        );
```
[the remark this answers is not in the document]
ok300: Good point. I don't think it's really necessary, as the swapper already checks for this:

https://github.com/breez/breez-sdk/blob/fd9861ea4e745c500fd4a718b7187d2ba50d1abe/libs/sdk-core/src/grpc/proto/breez.proto#L263

I will remove it.

JssDWt (approved):

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -708,6 +708,25 @@ impl NodeAPI for Greenlight {
         Ok(res.bolt11)
     }
 
+    async fn lookup_bolt11(&self, payment_hash: Vec<u8>) -> NodeResult<Option<String>> {
```
dangeross, line 711: I think `lookup_bolt11` implies retrieving the bolt11 from local persistence. Maybe `fetch_bolt11` describes it better and seems to be more inline with other functions here.

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -393,49 +394,57 @@ impl BTCReceiveSwap {
...
-        debug!("swap info confirmed = {}", swap_info.confirmed_sats);
-        if ln_invoice.amount_msat.unwrap() != (swap_info.confirmed_sats * 1000) {
-            warn!(
-                "invoice amount doesn't match confirmed sats {:?}",
-                ln_invoice.amount_msat.unwrap()
-            );
-            return Err(anyhow!("Does not match confirmed sats"));
-        }
+                    // If the swap was crated on a different device and the invoice payment failed,
+                    // the invoice already exists. In this case, lookup the invoice from the node.
```
dangeross, line 432: For me this makes sense, but I'm not sure if you mean: "...invoice payment failed..." on the other device
```suggestion
                    // If the swap was created on a different device and the invoice creation failed,
                    // then the invoice already exists. In this case, lookup the invoice from the node.
```

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -61,19 +62,25 @@ impl SwapperAPI for BreezServer {
...
+            .get_fund_manager_client()
             .await?
             .get_swap_payment(req)
             .await?
             .into_inner();
-        Ok(())
+
+        match resp.swap_error() {
+            crate::grpc::get_swap_payment_reply::SwapError::NoError => Ok(()),
+            err => Err(anyhow!("Failed to complete swap: {}", err.as_str_name())),
```
roeierez, line 74: Good catch! Probably that means that we didn't catch these errors before your fix.

roeierez (changes requested):
Looks good! One comment about saving the invoice. Other than that LGTM.

libs/sdk-core/src/greenlight/node_api.rs
ok300: Done: 31e56b12870421103bde7354783d7c4365b18f94

libs/sdk-core/src/swap_in/swap.rs
ok300: Good point, it's ambiguous. I rewrote it to clarify: ec09de9b6da3e61ff09badb4b679db875557a3d9

libs/sdk-core/src/swap_in/swap.rs
ok300: Thanks. Yes, i's now propagating the swapper error, if any, all the way to `last_redeem_error`. Before they were silent errors.

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -393,49 +394,57 @@ impl BTCReceiveSwap {
...
-                "invoice amount doesn't match confirmed sats {:?}",
-                ln_invoice.amount_msat.unwrap()
-            );
-            return Err(anyhow!("Does not match confirmed sats"));
-        }
+                    // If the swap was crated on a different device and the invoice payment failed,
+                    // the invoice already exists. In this case, lookup the invoice from the node.
+                    Err(ReceivePaymentError::InvoicePreimageAlreadyExists { .. }) => {
+                        match self.node_api.lookup_bolt11(swap_info.payment_hash).await? {
+                            Some(bolt11) => Ok(bolt11),
```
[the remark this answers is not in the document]
ok300: Done: ccd7471f746b4a426e792971d3b2c26a0f2827d6

roeierez (approved):
LGTM

dangeross (approved):
LGTM!

Issue #730, opened 2024-01-19
Cannot remove a registered webhook URL


jose31canizar:
Currently there is no way to remove a registered webhook URL. So after changing a query param in the url, for example the language param, it will register an additional webhook and thus the device will receive two notifications for every payment. It would be great to have the capability of removing a previously registered webhook so users can re-register with a different language.

JssDWt:
This needs some infrastructure updates and sdk support. Reopening.

PR #786, opened 2024-02-11
cache: remove webhook

rustaceanrob:
This would a small step in removing the webhook (#730) in the case the user changes languages or would no longer like notifications. Also makes a change in the delete function for cache to use a `&str` instead of `String` for consistency.


roeierez (approved):
LGTM thank you.

ok300 (approved):

PR #789, opened 2024-02-14
Check for ongoing swaps during register_webhook

ok300:
If there are any ongoing swaps when `register_webhook` is called, then re-register for swap notifications.

This covers two scenarios:
- if the app is running and the push token is renewed, `register_webhook()` will propagate the new `webhook_url` for both the payment notification and the swap tx one (on the chainnotifier)
- if the app was stopped for a while and is just starting, but in the meantime the push token was changed, this also ensures both notification types use the new `webhook_url`

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -732,8 +732,11 @@ impl BreezServices {
             .btc_receive_swapper
             .create_swap_address(channel_opening_fees)
             .await?;
-        self.register_swap_tx_notification(&swap_info.bitcoin_address)
-            .await?;
+        if let Some(webhook_url) = self.persister.get_webhook_url()? {
+            info!("Webhook URL found in local cache, registering for swap tx notification");
+            self.register_swap_tx_notification(&swap_info.bitcoin_address, &webhook_url)
+                .await?;
+        }
         Ok(swap_info)
     }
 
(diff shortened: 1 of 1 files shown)


erdemyerebasmaz (approved):
LGTM

libs/sdk-core/src/breez_services.rs
```diff
@@ -1570,9 +1570,14 @@ impl BreezServices {
         })
     }
 
-    /// Register for webhook callbacks at the given `webhook_url` whenever a new payment is received.
+    /// Register for webhook callbacks at the given `webhook_url`.
     ///
-    /// More webhook types may be supported in the future.
+    /// Current supported callback types: when a new payment is received, when a swap tx is confirmed.
```
dangeross, line 1575:
```suggestion
    /// Currently supported callback types:
    /// - a payment is received
    /// - an LNURL pay info or invoice request
    /// - a swap tx is confirmed
```

dangeross (changes requested):
Just needs the LNURL-pay callback types including in the comment

dangeross (approved):
LGTM!

libs/sdk-core/src/breez_services.rs
```diff
@@ -1591,11 +1596,17 @@ impl BreezServices {
             )
             .await?;
 
+        if let Some(swap) = self.in_progress_swap().await? {
```
roeierez, line 1602: It will only refresh the hook for the in progress (address that already has a transaction associated).
Aren't we messing the unused addresses?
Perhaps we should register for all addresses that return from `list_monitored` ?

Also perhaps we should only do this if the webhook url has changes?

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: > perhaps we should only do this if the webhook url has changes?

Done: ca19d3d8a578262aa76846b767451fb8c458849a . Applies to all other callback types too.

ok300: Yes, using `list_monitored` makes more sense.

Done in d6c2fe4e4af98c8fd62c21c9afd583d75ed640ee

libs/sdk-core/src/breez_services.rs
```diff
@@ -1591,11 +1603,19 @@ impl BreezServices {
             )
             .await?;
 
+        for swap in self.btc_receive_swapper.list_monitored()?.iter()
+            .filter(|swap| !swap.refundable())
+        {
+            let swap_address = &swap.bitcoin_address;
+            info!("Found non-refundable monitored swap with address {swap_address}, registering for swap tx notifications");
+            self.register_swap_tx_notification(swap_address).await?;
```
roeierez, line 1611: If this call fails we may not try again next time since we already saved the new webhook url. I think we should move the `set_webhook_url` cal to the end of this function.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
```diff
@@ -1586,16 +1597,31 @@ impl BreezServices {
...
 
+        for swap in self
+            .btc_receive_swapper
+            .list_monitored()?
+            .iter()
+            .filter(|swap| !swap.refundable())
+        {
+            let swap_address = &swap.bitcoin_address;
+            info!("Found non-refundable monitored swap with address {swap_address}, registering for swap tx notifications");
+            self.register_swap_tx_notification(swap_address).await?;
```
roeierez, line 1613: This will register now with the old url as it is not saved in the db.

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: Fixed: 6eb54b06f4ddefbec9066085d7ee457095b97431

roeierez (approved):

ok300:
I've updated the docs in e173e862e693ec13f5f5c0b64880abffa6880976

@roeierez is this an accurate description?

roeierez:
> I've updated the docs in [e173e86](https://github.com/breez/breez-sdk/commit/e173e862e693ec13f5f5c0b64880abffa6880976)
> 
> @roeierez is this an accurate description?

Let's remove the LNURL pay, it is not part of the sdk ATM. It is only part of c-breez.

PR #790, opened 2024-02-14
Connect with restore_only option

### LINKED ISSUE #788: Allow restore only mode for connect
We want to allow user to restore a node only if exists.
There are some use cases where a discovery is needed for a specific seed without creating a node.

dangeross:
Adds a `restore_only` config parameter to only restore an existing node on connect

Generator PR https://github.com/breez/breez-sdk-rn-generator/pull/20
Fixes #788


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -14,6 +14,13 @@ enum SdkError {
     "ServiceConnectivity",
 };
 
+[Error]
+enum BreezServicesError {
```
roeierez, line 18: NIT: `ConnectError` perhaps?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -132,6 +139,7 @@ dictionary Config {
     u32 payment_timeout_sec;
     string? default_lsp_id;
     string? api_key;
+    boolean restore_only;
```
roeierez, line 142: It seems as the `restore_only` flag is better fit as a parameter to the connect function rather than adding it to the config. Nevertheless I see your concern about adding another parameter to the connect.

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
dangeross: Yes, I wanted to avoid adding a restore_only param or options struct, which seems excessive.

The other option would be a ConnectRequest:
```
pub struct ConnectRequest {
    pub config: Config,
    pub seed: Vec<u8>,
    pub restore_only: Option<bool>,
}

pub async fn connect(
    req: ConnectRequest,
    event_listener: Box<dyn EventListener>,
) -> Result<Arc<BreezServices>, ConnectError>
```

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -92,22 +92,29 @@ impl Greenlight {
         // Query for the existing credentials
         let mut parsed_credentials =
             Self::get_node_credentials(config.network, &signer, persister.clone())?
-                .ok_or(anyhow!("No credentials found"));
+                .ok_or(NodeError::Generic(anyhow!("No credentials found")));
         if parsed_credentials.is_err() {
             info!("No credentials found, trying to recover existing node");
             parsed_credentials = match Self::recover(config.network, seed.clone()).await {
                 Ok(creds) => Ok(creds),
```
ok300, line 99: `recover` also works, so perhaps `restore_only` is a bit misleading?

Maybe `deny_register` or `deny_new_nodes` is better?

dangeross: `recover` only recovers the device cert/key from GL if the node exists, so it's a valid way to find out if a node exists for the seed.

libs/sdk-bindings/src/breez_sdk.udl
roeierez: I am in favor of ConnectRequest. Also more consistent with other methods.
Why not inserting the event listener inside the ConnectRequest like other parameters?

dangeross: I can test, but I think it might break the RN code generation

roeierez: In that case I think we can go with the listener outside the struct.

roeierez (approved):
LGTM. Let's hold merging until the swap notifications fixes are in so we can release before this change.

PR #791, opened 2024-02-15
Re-use lightning-invoice crate exported by gl-client

ok300:
This PR replaces our explicit `lightning-invoice` dependency with the `gl-client` one, similar to how #779 did the same for the `bitcoin` crate.

The benefits we get are:
- we reduce the dependency tree and binary size
  - no need to bundle both our `lightning-invoice` version and the one from `gl-client`
- we help our SDK users easily choose the right `lightning-invoice` version
  - they can simply access it with `breez_sdk_core::lightning_invoice`

### DIFF
--- a/libs/sdk-core/src/lib.rs
+++ b/libs/sdk-core/src/lib.rs
@@ -191,9 +191,8 @@ mod swap_in;
 mod swap_out;
 
 // Re-use crates from gl_client for consistency
-// Re-export them so that others using the SDK have easy access to the same versions we use
-pub use gl_client::bitcoin;
-pub use gl_client::lightning_invoice;
+use gl_client::bitcoin;
+use gl_client::lightning_invoice;
 
 pub use breez_services::{
     mnemonic_to_seed, BackupFailedData, BreezEvent, BreezServices, CheckMessageRequest,


libs/sdk-core/Cargo.toml
```diff
@@ -48,9 +48,8 @@ tonic = { version = "^0.8", features = [
     "tls-webpki-roots",
 ] }
 lazy_static = "^1.4.0"
-lightning = "0.0.118"
-# Keep in sync with version used in gl-client
-lightning-invoice = "0.26.0"
+# Keep in sync with the version used by gl-client
+lightning = "0.0.116"
```
JssDWt, line 52: This is changed from 0.0.118 to 0.0.116. Why is that?
Can we also do the `lightning` crate in the same construct?

ok300: > This is changed from 0.0.118 to 0.0.116. Why is that?

This is because GL recently consolidated two `lightning-invoice` versions, who each were compatible with different `lightning` versions. The winning pair was `lightning-invoice` 0.24.0 / `lightning` 0.0.116, as used in VLS ([link](https://gitlab.com/lightning-signer/validating-lightning-signer/-/blob/v0.11.0-rc.1/vls-core/Cargo.toml?ref_type=tags#L128) / [link](https://gitlab.com/lightning-signer/validating-lightning-signer/-/blob/v0.11.0-rc.1/vls-core/Cargo.toml?ref_type=tags#L110)).

We had both 0.0.116 and 0.0.118 before, but now only one of them is compatible with the GL code, which is 0.0.116.

So we're now using the same version as GL and VLS.

> Can we also do the lightning crate in the same construct?

Would first need to ask GL to re-export the VLS `lightning` dependency. Added a PR for that: https://github.com/Blockstream/greenlight/pull/359

Once that's reviewed and merged, I will add a PR to the SDK.

ok300:
As discussed with Roei, I removed the re-exports, as we don't use those crates in our public APIs.

roeierez (approved):
LGTM

PR #794, opened 2024-02-16
Add confirmed_at to swap info

ademar111190:


roeierez (approved):
LGTM

dangeross (commented):
I think the PR title is misleading. I see confirmed_at added to the swaps_info table, but created_at not removed from the swaps table. Is that the intent?

libs/sdk-core/src/models.rs
```diff
@@ -1288,6 +1288,8 @@ pub struct SwapInfo {
     ///
     /// Swaps created after dynamic fees were introduced always have this field set.
     pub channel_opening_fees: Option<OpeningFeeParams>,
+    /// The block height when the swap was confirmed.
+    pub confirmed_at: u32,
```
dangeross, line 1292: Would it be better for it to be an Option<u32> and nullable in the table?

roeierez:
> I think the PR title is misleading. I see confirmed_at added to the swaps_info table, but created_at not removed from the swaps table. Is that the intent?

Yes the intent is only add a field as the created_at is still valid (need some fix in a different PR).

libs/sdk-core/src/models.rs
roeierez: Yes, makes sense. Donw

dangeross (approved):
LGTM

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -381,14 +383,16 @@ impl BTCReceiveSwap {
...
-            utxos.confirmed_sats(),
-            utxos.confirmed_tx_ids(),
-            swap_status,
-        )?)
+        let chain_info = SwapChainInfo {
+            unconfirmed_sats: utxos.unconfirmed_sats(),
+            unconfirmed_tx_ids: utxos.unconfirmed_tx_ids(),
+            confirmed_sats: utxos.confirmed_sats(),
+            confirmed_tx_ids: utxos.confirmed_tx_ids(),
+            confirmed_at: confirmed_block,
```
roeierez, line 391: I see a bug here where the confirmed_at is not optional and may be zero.

dangeross: Good catch, I thought at this point the known

dangeross (approved):

ok300 (approved):

PR #795, opened 2024-02-19
Expose `redeem_swap`

ok300:
This PR exposes the `redeem_swap` as an SDK method, to allow redeeming an individual swap.

Necessary for https://github.com/breez/c-breez/pull/783

### DIFF
--- a/libs/sdk-flutter/lib/bridge_generated.dart
+++ b/libs/sdk-flutter/lib/bridge_generated.dart
@@ -258,6 +258,11 @@ abstract class BreezSdkCore {
 
   FlutterRustBridgeTaskConstMeta get kRescanSwapsConstMeta;
 
+  /// See [BreezServices::redeem_swap]
+  Future<void> redeemSwap({required String swapAddress, dynamic hint});
+
+  FlutterRustBridgeTaskConstMeta get kRedeemSwapConstMeta;
+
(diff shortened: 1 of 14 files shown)


roeierez (approved):
LGTM

libs/sdk-flutter/lib/bridge_generated.dart
```diff
@@ -258,6 +258,11 @@ abstract class BreezSdkCore {
 
   FlutterRustBridgeTaskConstMeta get kRescanSwapsConstMeta;
 
+  /// See [BreezServices::redeem_swap]
+  Future<void> redeemSwap({required String swapAddress, dynamic hint});
```
erdemyerebasmaz, line 262: This function needs to be exposed through `BreezSDK`(_sdk-flutter/lib/sdk.dart_) to be accessible from the plugin.

erdemyerebasmaz (changes requested):
One comment:

libs/sdk-flutter/lib/bridge_generated.dart
ok300: Good catch.

Is this the right way to do it? 67c89ee0132c554ddf7553d39f09f5d590aff1b4

erdemyerebasmaz (approved):

PR #799, opened 2024-02-19
Re-use `lightning` crate exported by `gl-client`

ok300:
This PR is the equivalent of #791 for the `lightning` dependency.

It includes:
- bumping GL by one commit, such that it now re-exports its `lightning` dependency
- removing the direct dependency from our `Cargo.toml`
- updating our internal references to re-use the imported one from `gl_client`

### DIFF
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -528,7 +528,6 @@ dependencies = [
  "gl-client",
  "hex",
  "lazy_static",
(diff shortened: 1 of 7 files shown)


JssDWt (approved):
(It bumps GL by 3 commits)
ACK!

roeierez (approved):
LGTM

Issue #774, opened 2024-02-06
`SwapInfo` field `created_at` is always zero


danielgranhao:
I don't know if this field is supposed to be a timestamp or a blockheight. While trying to figure that out from the source code, I realized this isn't ever modified from the initial moment a swap is created (`created_at` set to zero).

I also confirmed from a confirmed successful swap of mine that the value of `created_at` was indeed zero.

roeierez:
Thanks @danielgranhao for bringing to our attention.
I actually look at the field and the comment looks wrong (Relative time lock start) The relative time lock starts at confirmation and not at creation so the creation time is not really needed IMO.
Let me know if I am missing something in my assumption.

danielgranhao:
@roeierez Yes, it would make sense if the field was named `confirmed_at`.

roeierez:
Alright then. Let's rename it to `confirmed_at` and use the block height of the confirmed transaction.

PR #800, opened 2024-02-20
Set created at using the current time

ademar111190:
- Set the current time on swap info's `created_at` field at the creation moment.
- ~I saw we are getting the current time in some places already duplicating the logic of getting the time and converting it to i64 from u64; so I am thinking about adding a central point to provide the time already converted~
- ~Added my proposal to this PR; with this proposal, we are not only centralizing the logic but also providing a way for help testing~

~Let me know your thoughts; if this time utils proposal works for you I can convert the areas we are getting the time directly to use it; if not I can delete it and get the time directly as we are doing already in other places.~

Fixes https://github.com/breez/breez-sdk/issues/774


libs/sdk-core/src/time_utils.rs
```diff
@@ -0,0 +1,26 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+pub(crate) struct BreezTimeUtils;
```
roeierez, line 3: I think making this a trait only to be able to test created_at is overkill and we can test that without the trait.
IMO it is fine to get the current time directly from rust API.

roeierez (changes requested):

roeierez (approved):
LGTM

PR #802, opened 2024-02-20
Allow custom claim tx feerates for reverse swap fee estimates

### LINKED ISSUE #801: `fetch_reverse_swap_fees`: Allow optional user-chosen feerate
If the user sets a feerate for the claim tx, we can precisely calculate the reverse swap fees.

If no such feerate is given, `fetch_reverse_swap_fees` can return a fee estimate, as it does currently.

ok300:
Fixes #801


roeierez (approved):
LGTM

PR #806, opened 2024-02-21
Bump openssl dependency

ok300:
This addresses the dependabot finding: https://github.com/breez/breez-sdk/security/dependabot/136


JssDWt (approved):
I'm not aware of any issues with openssl, so ACK

PR #810, opened 2024-02-22
Add send_payment validation: liquidity

ok300:
Before attempting to send the payment, check a few node conditions that would obviously result in a failure.

This allows `send_payment` to fail instantly, instead of after about a minute with "ran out of routes".

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -91,6 +91,8 @@ enum SendPaymentError {
     "InvalidInvoice",
     "InvoiceExpired",
     "InvalidNetwork",
+    "NoConnectedPeers",
+    "NotEnoughLiquidity",
     "PaymentFailed",
     "PaymentTimeout",
     "RouteNotFound",
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -91,7 +91,6 @@ enum SendPaymentError {
     "InvalidInvoice",
     "InvoiceExpired",
     "InvalidNetwork",
-    "NoConnectedPeers",
     "NotEnoughLiquidity",
     "PaymentFailed",
     "PaymentTimeout",
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -294,7 +294,6 @@ impl BreezServices {
         {
             Some(_) => Err(SendPaymentError::AlreadyPaid),
             None => {
-                self.validate_peers_connected()?;
                 self.validate_liquidity(amount_msat)?;
 
                 self.persist_pending_payment(&parsed_invoice, amount_msat)?;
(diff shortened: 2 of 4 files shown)


libs/sdk-core/src/error.rs
```diff
@@ -682,6 +698,22 @@ pub enum SendPaymentError {
     ServiceConnectivity { err: String },
 }
 
+impl SendPaymentError {
+    pub(crate) fn not_enough_liquidity(
+        amount_msat: u64,
+        channels_balance_msat: u64,
+        onchain_balance_msat: u64,
+        pending_onchain_balance_msat: u64,
```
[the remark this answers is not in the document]
ok300: Yes.

Just saying "not enough liquidity" might scare people into thinking they lost funds.

Mentioning the onchain amounts in the error will hint to the app developer to either suggest to the end-user an action based on the onchain funds, or (if onchain = 0) to show a friendly message like "no balance available, please fund it first",

libs/sdk-core/src/breez_services.rs
```diff
@@ -294,6 +294,9 @@ impl BreezServices {
         {
             Some(_) => Err(SendPaymentError::AlreadyPaid),
             None => {
+                self.validate_peers_connected()?;
```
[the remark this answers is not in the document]
ok300: Good point about persisted state. Yes, the info could be stale.

Unfortunately the only error we get from GL is "ran out of routes to try" after it keeps trying to pay for about a minute.

On the other hand, GL is managing the peers, so it's trying to connect to one if unconnected. So it could be that calling `send_payment` with no peers will still work, as it could connect to the peer while trying to pay.

Then overall, IMO it's better if I remove this.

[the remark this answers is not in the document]
ok300: Done: ef80e5f891c695314c761af9412f93cdaed39111

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -91,6 +91,7 @@ enum SendPaymentError {
     "InvalidInvoice",
     "InvoiceExpired",
     "InvalidNetwork",
+    "NotEnoughLiquidity",
```
JssDWt, line 94: Perhaps `InsufficientFunds` is more understandable for non-lightning experts.

libs/sdk-core/src/breez_services.rs
```diff
@@ -294,6 +294,8 @@ impl BreezServices {
         {
             Some(_) => Err(SendPaymentError::AlreadyPaid),
             None => {
+                self.validate_liquidity(amount_msat)?;
```
roeierez, line 297: We don't force sync before paying so the channels balance might not be synced as well with the node backend.
I think the reasonable way to solve it and return a proper timely mannered response is by adding this optimization to c-lightning.

dangeross: I raised an issue in Greenlight last week related to this: https://github.com/Blockstream/greenlight/issues/361

ok300: I agree, the right place to solve it is GL.

I'll close this, as it will be addressed by the GL issue above.

PR #811, opened 2024-02-22
swap events

roeierez:


libs/sdk-core/src/models.rs
```diff
@@ -1213,10 +1214,18 @@ pub enum SwapStatus {
...
+
+    Redeemed = 3,
+
     /// The swap address has confirmed transactions associated with it and the lock timeout has passed since
     /// the earliest confirmed transaction. This means the only way to spend the funds from this address is by
     /// broadcasting a refund transaction.
-    Expired = 1,
+    Refundable = 4,
+
+    Completed = 5,
```
dangeross, line 1228: Is there any need to migrate the swap_info table to update the swap status from Expired to Refundable/Completed else it will become WaitingConfirmation

roeierez: Sure, I am waiting with the migration until I'll finish the implementation.
My plan is to reset all of them to Initial which will include them in the next rescan by the swapper. It will update the right state for all of them.

libs/sdk-core/src/breez_services.rs
```diff
@@ -81,21 +81,35 @@ pub trait EventListener: Send + Sync {
...
     /// Indicates that the backup process has just started
     BackupStarted,
     /// Indicates that the backup process has just finished successfully
     BackupSucceeded,
     /// Indicates that the backup process has just failed
-    BackupFailed { details: BackupFailedData },
+    BackupFailed {
+        details: BackupFailedData,
+    },
+    // Indicates that we have just updated the swap associated information
```
ok300, line 109:
```suggestion
    /// Indicates that this swap's status was just updated
```

This more strongly indicates that mainly the status has changed, and not only just some of the fields.

libs/sdk-core/src/chain.rs
```diff
@@ -75,13 +75,21 @@ impl AddressUtxos {
...
 /// Gets unspent tx outputs. Specifically filters out inbound utxos that have been spent.
-pub(crate) fn get_utxos(address: String, transactions: Vec<OnchainTx>) -> Result<AddressUtxos> {
+/// If include_unconfirmed_spends is true, then the result won't include utxos that were spent
+/// in unconfirmed transactions.
+pub(crate) fn get_utxos(
+    address: String,
+    transactions: Vec<OnchainTx>,
+    include_unconfirmed_spends: bool,
+) -> Result<AddressUtxos> {
     // Calculate confirmed amount associated with this address
```
ok300, line 85:
```suggestion
```

IMO better to remove this comment, as the method doesn't only deal with confirmed utxos anymore.

libs/sdk-core/src/models.rs
```diff
@@ -1297,29 +1310,85 @@ pub struct SwapInfo {
...
+        self.status == SwapStatus::Refundable
     }
 
     pub(crate) fn monitored(&self) -> bool {
-        self.unused() || self.in_progress() || self.refundable()
+        self.status != SwapStatus::Completed
+    }
+
+    fn calculate_status(&self, tip: u32) -> SwapStatus {
+        // The Refundable status never
```
ok300, line 1362: Incomplete comment?

libs/sdk-core/src/persist/swap.rs
```diff
@@ -94,16 +95,17 @@ impl SqliteStorage {
...
+            "UPDATE swaps_info SET paid_msat=:paid_msat, status=:status where bitcoin_address=:bitcoin_address",
             named_params! {
              ":paid_msat": paid_msat,
              ":bitcoin_address": bitcoin_address,
+             ":status": status as u32,
             },
         )?;
-
-        Ok(())
+        Ok(self.get_swap_info_by_address(bitcoin_address)?.unwrap())
```
ok300, line 108: Unwrap is bad, as it can panic on failure.

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -87,6 +89,8 @@ pub(crate) struct BTCReceiveSwap {
     persister: Arc<crate::persist::db::SqliteStorage>,
     chain_service: Arc<dyn ChainService>,
     payment_receiver: Arc<dyn Receiver>,
+    current_tip: Mutex<u32>,
```
ok300, line 92: Since we already have the persister, it might be simpler to use
```rust
self.persister.node_info()?.block_height
```

like in https://github.com/breez/breez-sdk/pull/812/files#diff-dfcf1c8c08fb243da9a4e02f631600d2e5a5c827fbbed56c8e98daadd4347483R253

roeierez: Good point but when tested I have noticed the greenlight block sometimes lag behind the mempool space block.
Almost always there is a 1 block lag. Also when the node is not connected for a while it takes some time to catchup and gl shows the tip it is aware of which lags a lot behind for a while.
Perhaps if gl will expose a listener to block change it will make more sense as even if lags a bit behind it will be consistent. What do you think?

libs/sdk-core/src/chain.rs
```diff
@@ -75,13 +75,21 @@ impl AddressUtxos {
 }
 
 /// Gets unspent tx outputs. Specifically filters out inbound utxos that have been spent.
-pub(crate) fn get_utxos(address: String, transactions: Vec<OnchainTx>) -> Result<AddressUtxos> {
+/// If include_unconfirmed_spends is true, then the result won't include utxos that were spent
+/// in unconfirmed transactions.
```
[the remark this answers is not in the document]
roeierez: Actually in case where this flag is set to true,  the result will not include confirmed utxos if they were spend by unconfirmed transactions.
This is because we consider these spends even though they are not confirmed

libs/sdk-core/src/breez_services.rs
roeierez: I combined your suggestion in the comment as we we emit events not only when the status has changed.

libs/sdk-core/src/chain.rs
roeierez: removed

libs/sdk-core/src/persist/swap.rs
roeierez: True, made it in by mistake, we don't need it.

libs/sdk-core/src/models.rs
roeierez: Fixed

libs/sdk-core/src/chain.rs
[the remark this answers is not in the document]
dangeross: Apologies, following the code further down it makes sense

libs/sdk-core/src/swap_in/swap.rs
ok300: I wasn't aware the GL tip was lagging. Yes, then it makes sense the way you did it.

I'll abandon my other PR using `self.persister.node_info()?.block_height` as it would suffer from the same problem.

libs/sdk-core/src/breez_services.rs
ok300: I see, then I misunderstood it. I thought the event is only emitted on status change.

ok300: May I ask what's the use case for that? Would a user want to get an event for something other than a status change?

roeierez: Sure, among other things we update the swap chain information such as refund tx id (after refund).
We also update the swap open channel fees from time to time.

ok300: Makes sense, thanks.

ok300 (approved):

dangeross (approved):
LGTM, pending migrations

erdemyerebasmaz:
@roeierez
Documentation for SwapInfo.status should be updated to reflect these changes. [L1275-R1291](https://github.com/breez/breez-sdk/pull/811/files#diff-78b2d03eb877d309a4fe8572e1599cc8918d7879dfdbb88ddae440a556fce9fbL1275-R1291)

PR #812, opened 2024-02-23
Minimize the use of mempool.space /tip endpoint

ok300:
This is an attempt to minimize the amount of mempool.space calls we make.

It replaces, where possible, the use of `chain_service.current_tip()` with our internal `node_info()?.block_height`.


roeierez:
@ok300  I think the call that is more likely causes the rate limit is inside track_new_blocks where we poll for the block changes.

ok300:
I'm not sure about that, as `track_new_blocks` is ticking only once every 30s.

For now I'll abadon this PR, as the proposed tip would lag behind the real tip (https://github.com/breez/breez-sdk/pull/811#discussion_r1504709187).

PR #814, opened 2024-02-23
DRAFT: add min/max payment size to opening fee params

JssDWt:
The fields `min_payment_size_msat` and `max_payment_size_msat` will be added to lspd in https://github.com/breez/lspd/pull/193

Currently, the only purpose to add these in the SDK is to allow LSPD to create a `promise` in the opening_fee_params that includes these fields. Later, these fields can actually be used to enforce the min and max payment size.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -426,6 +426,8 @@ dictionary OpeningFeeParams {
     string valid_until;
     u32 max_idle_time;
     u32 max_client_to_self_delay;
+    u64 min_payment_size_msat;
+    u64 max_payment_size_msat;
     string promise;
 };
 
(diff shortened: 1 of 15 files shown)


roeierez:
@JssDWt so the payment size differs between different fee params?
Does it mean we need to change our logic when we lookup for the best matched fee params? Should we not only consider cheapest one but also this that supports the requested amount?

JssDWt:
> Does it mean we need to change our logic when we lookup for the best matched fee params? Should we not only consider cheapest one but also this that supports the requested amount?

@roeierez Eventually that would be the idea, yes. Unfortunately we cannot enforce the min/max payment size inside lspd. We create a promise based on a signature in the fields of the opening fee params. If some clients don't return these fields, however, we will never be able to check that signature if the signature invludes these fields. So these fields have to be excluded from the signature. But that means they cannot be trusted when the client sends params back to the server.

So for now these fields are only here so they will be passed back to the server. That way the server can eventually enforce them.

Another option could be to only add these fields in LSPS2 and not bother adding them in our grpc protocol.

JssDWt:
Closing, because these fields will only be included in LSPS2.

PR #816, opened 2024-02-25
Add next-iteration reverse swap

ok300:
This PR adds a new version of reverse swap methods, to support all following scenarios:
- user wishes to set `receive_amount`, we derive the matching `send_amount`
- user wishes to set `send_amount`, we calculate the matching `receive_amount`

In all cases we return both `send_amount` and `receive_amount`, as well as intermediary fees that add up to the difference. We also validate that `send_amount` fits within the Boltz `min`/ `max`. If it's outside the range, the `send_amount` and `receive_amount` returned are empty.


libs/sdk-core/src/breez_services.rs
```diff
@@ -907,6 +907,63 @@ impl BreezServices {
...
+        Ok(PrepareOnchainPaymentResponse {
+            min: fee_info.min,
+            max: fee_info.max,
+            fees_hash: fee_info.fees_hash,
+            fees_percentage: fee_info.fees_percentage,
+            fees_lockup,
+            fees_claim,
+            send_amount_sat: res_send_amt,
+            receive_amount_sat: res_recv_amt,
+            // TODO We might not need this field anymore, if we return both send_amt and recv_amt
```
roeierez, line 956: I agree we can omit this.

libs/sdk-core/src/models.rs
```diff
@@ -982,6 +982,37 @@ pub struct SendOnchainResponse {
     pub reverse_swap_info: ReverseSwapInfo,
 }
 
+/// See [ReverseSwapFeesRequest]
+pub struct PrepareOnchainPaymentRequest {
+    pub amount_sat: u64,
+
+    /// If true, amount is send_amount. If false, amount is receive_amount.
+    pub is_send_amount: bool,
```
roeierez, line 990: Did you try different ways to express that? Perhaps enum? Or two optional fields (send_amount, received_amount)?
If enum with two variants translates simply to foreign languages then it could be a better option.

roeierez (commented):
Overall looks good.

libs/sdk-core/src/breez_services.rs
ok300: Done: de2a67fcf4a44d6a1d559e9d63d402f74dd382d9

libs/sdk-core/src/models.rs
ok300: Something like this? 0f926038f03839474c3f95e3006f31fab8b3b545

I initially avoided enums, because AFAIK foreign languages need named inner fields, and that would become verbose. But using a plain enum instead of the boolean is a good solution IMO.

ok300:
Other than the bindings for RN and docs, which I'll add at the end, it's ready for review.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -733,10 +764,21 @@ dictionary SendOnchainRequest {
     u32 sat_per_vbyte;
 };
 
+dictionary PayOnchainRequest {
+    u64 send_amount_sat;
+    u64 receive_amount_sat;
```
roeierez, line 769: How do we validate that the receive_amount matches the send_amount in a way that they were generated in a valid prepare request? Do we leave that to the user responsibility? Even if this is the case we still need to ensure we won't create eventually illegal claim transaction.

roeierez (changes requested):
Looks good! Left some comments

libs/sdk-core/src/models.rs
```diff
@@ -982,6 +987,52 @@ pub struct SendOnchainResponse {
...
+    pub fees_claim: u64,
+
+    // Fields set only if the resulting send amount is within range
+    pub send_amount_sat: Option<u64>,
+    pub receive_amount_sat: Option<u64>,
+}
+
+pub struct PayOnchainRequest {
+    pub send_amount_sat: u64,
+    pub receive_amount_sat: u64,
```
erdemyerebasmaz, line 1027: Shouldn't these be optional?

ok300: No because they are both returned in `PrepareOnchainPaymentResponse`.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -434,6 +442,17 @@ dictionary ReverseSwapFeesRequest {
     u32? claim_tx_feerate = null;
 };
 
+enum DesiredSwapAmountType {
```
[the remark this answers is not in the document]
ok300: Fixed 88d7cd1cb8f4371daed8d58b0080ee6e39469855

libs/sdk-core/src/models.rs
erdemyerebasmaz: It's a bit confusing on the app side, let's say I'd like to prepare onchain payment for sending 50_000 sats, why would I also pass a value to receive amount sat?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -548,6 +567,18 @@ dictionary ReverseSwapPairInfo {
...
+dictionary PrepareOnchainPaymentResponse {
+    u64 min;
+    u64 max;
+    string fees_hash;
+    f64 fees_percentage;
+    u64 fees_lockup;
+    u64 fees_claim;
+
+    u64? send_amount_sat;
+    u64? receive_amount_sat;
```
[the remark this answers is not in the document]
ok300: Done: b5b537c62ab3ee84d09be087d3d3f4d2ae7a0ba7

libs/sdk-bindings/src/breez_sdk.udl
ok300: How about this: we could check if `receive_amount_sat > onchain_amount_sat`. Even if it's a 1 sat difference, it's technically a valid fee. Or we could enforce a minmum claim feerate of 1 sat/vB.

We can validate this in `create_reverse_swap`.

What do you think?

roeierez: I think going with at least 1 sat/vB for the claim is good.

ok300: Done: c5c9e9208492181760ba83a16cccd757bffdd222

libs/sdk-core/src/models.rs
ok300: The order would be
- `prepare_onchain_payment` with `amount_sat = 50 000` and `amount_type = Send`
- this returns the intermediary fees, `send_amount_sat` and `receive_amount_sat`
- the app can then call `pay_onchain` with the send and receive amounts

Does this address your question?

libs/sdk-core/src/swap_out/reverseswap.rs
```diff
@@ -140,17 +177,24 @@ impl BTCSendSwap {
...
-            .await?;
+            .await
+            .map(hex::encode)?;
         let created_rsi = self
-            .create_and_validate_rev_swap_on_remote(req, hex::encode(reverse_routing_node))
+            .create_and_validate_rev_swap_on_remote(req.clone(), routing_node)
             .await?;
+
+        if let CreateReverseSwapArg::V2(req) = req {
+            let claim_fee = created_rsi.onchain_amount_sat - req.receive_amount_sat;
```
roeierez, line 194: possible negative value here?

roeierez (changes requested):

libs/sdk-core/src/swap_out/reverseswap.rs
ok300: This should take care of it: b8cb40df284c05b61b7c3ea67d6e8013ca6ddb5f

libs/sdk-core/src/models.rs
roeierez: @ok300 we had a discussion today and Erdem has a point. The relation between the prepareOnchainResponse and the payOnchain input is not concise and has some edge cases that are not so clear to a new user.
The suggestion was:
1. The prepareOnchain will return PayOnchainRequest
2. Then the user should call payOnchain with the above request so it is clear how to use the whole flow.

@erdemyerebasmaz this is what you had in mind?

ok300:
I made as many of the changes we discussed as possible. More specifically:
- [x] Three methods `fetch_onchain_limits`, `prepare_onchain_payment` and `pay_onchain`
- [x] Return only the `min_sat` `max_sat` with `fetch_onchain_limits`.
  - The `fees_hash` had to also be included if we want to be able to check if the range changed in `prepare`, see below.
- [x] When valid, `prepare_onchain_payment` returns a response to be used to call `pay_onchain`. This includes all fees, fees_hash, sender/recipient amounts.
- [x] If the limits changed when calling `prepare_onchain_payment`, throw `OnchainLimitsChanged` error.
  - `prepare` doesn't know the range used by the user in `fetch_limits`. It only knows the `fees_hash`. If that changed, that could mean the fee promises (service fee, lockup fee) have changed and / or the range. Therefore the new error type here is `FeePromiseChanged`.
- [x] Document what `OnchainLimitsChanged` means and what to do (call `fetch_onchain_limits` again).
  - See 3d3df4379c814c18f29493e5fc53c5b1dd55a93a for how the new error(s) were documented
- [x] `pay_chain` validates the `prepare_onchain_payment` response used in the request
  - Since the validation can only be done against a full list of fee details, and `pay` only has the `fees_hash` as arg, this means `pay` has to fetch the most recent fees details from Boltz and compare the `fees_hash`. If they are equal, validation can proceed on the fee values. This validation is done in `validate_against_fresh_fees_info` and is re-used by `prepare`.
- [ ] Throw OnchainLimitsChanged error if the fees_hash is invalid calling Boltz.
  - This would require us to parse the error message from Boltz, to identify this error case as separate from any others. This is probably not a good idea.
- [x] Throw an error if the request data cannot be used in `pay_onchain`.
- [x] `pay_onchain` function to receive the entire response object from `prepare_onchain` as a parameter

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -877,6 +925,12 @@ interface BlockingBreezServices {
    [Throws=SdkError]
    ReverseSwapPairInfo fetch_reverse_swap_fees(ReverseSwapFeesRequest req);
 
+   [Throws=SdkError]
+   FetchOnchainLimitsResponse fetch_onchain_limits();
```
roeierez, line 929: `fetch_onchain_limits ` doesn't imply paying onchain. Perhaps `onchain_payment_limits` ?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -435,6 +445,24 @@ dictionary ReverseSwapFeesRequest {
     u32? claim_tx_feerate = null;
 };
 
+dictionary FetchOnchainLimitsResponse {
+    u64 min_sat;
+    u64 max_sat;
+    string fees_hash;
```
roeierez, line 451: I think it complicates the flow if we return here fee hash and require it to be passed to the PrepareOnchain.
Seems simpler to detach the limits call from the prepare call.
Given the fact they are being called almost together the probability for limits change that affect the user amount seems very small.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -435,6 +445,24 @@ dictionary ReverseSwapFeesRequest {
...
+enum SwapAmountType {
+    "Send",
+    "Receive",
+};
+
+dictionary PrepareOnchainPaymentRequest {
+    u64 amount_sat;
+    SwapAmountType amount_type;
+    u32 claim_tx_feerate;
+    string fees_hash;
```
roeierez, line 463: Same here, I think we can avoid passing the fee hash here. After all the prepareOnchain use the swapper API to retrieve a new fees_hash.

libs/sdk-core/src/breez_services.rs
```diff
@@ -930,6 +920,109 @@ impl BreezServices {
...
+    }
+
+    /// Creates a reverse swap and attempts to pay the HODL invoice
+    ///
+    /// Supersedes [BreezServices::send_onchain]
+    pub async fn pay_onchain(
+        &self,
+        req: PayOnchainRequest,
+    ) -> Result<PayOnchainResponse, PayOnchainError> {
+        let fee_info = self.btc_send_swapper.fetch_reverse_swap_fees().await?;
```
roeierez, line 998: Why do we need to fetch the fees again if we have the fee_hash? Isn't it better to create the swap using this fee hash which is promised to return the same fees the user got from the prepare call? We can then validate against the returned swap.

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
ok300: Done

libs/sdk-bindings/src/breez_sdk.udl
ok300: This strongly depends on how the app handles this:
- call `onchain_payment_limits` once on app startup, then always use same limits
- call `onchain_payment_limits` when needed, but cache results
- call `onchain_payment_limits` for each new swap
- call `onchain_payment_limits` for each new swap and when `FeePromiseChanged` error is thrown

IMO the first two are much more likely than the last two.

As a consequence, they're more likely to face the changed limits situation. Having the `fees_hash` in the `onchain_payment_limits` result prevents this.

I can remove it, but because of the above, doing so will IMO open the door to `OutOfRange` errors while the user input appears to be in range.

Should I still remove it?

libs/sdk-bindings/src/breez_sdk.udl
ok300: Yes, if `fees_hash` is removed from the `onchain_payment_limits` result, this will go as well. Depends on decision above.

libs/sdk-bindings/src/breez_sdk.udl
roeierez: @ok300  Even if we leave it we still create a new fee_hash in prepareOnchain so we can't use it anyway...

libs/sdk-core/src/breez_services.rs
ok300: Done: 6c876a6d99d3160d0a85f688e0e74eca734cef4f

libs/sdk-bindings/src/breez_sdk.udl
ok300: Actually, the fresh `fee_hash` in `prepare` allows us to check if the `onchain_payment_limits` `fee_hash` is expired or not.

So, by leaving it, we can do this check. By removing it, we expose the caller to the risks above.

roeierez: That last commit that compares the input fee_hash to the new fee_hash takes into consideration some non validated assumptions on the swapper internal logic. We don't know how they generate these hashes and if they decide to return new hash for every request based on time or some other factors (even though the limit hasn't change).
So this has risk IMO and I strongly think that for simplicity we should remove the fee_hash from the onchain limits.

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
ok300: > We don't know how they generate these hashes

We do know:

> `hash`: SHA256 hash of the JSON encoded data in the pair object
> https://docs.boltz.exchange/v/api/api-v1#supported-pairs

The pair object contains only the fees and the limits, see https://api.boltz.exchange/getpairs , no nonce or timestamp-related value.

> some non validated assumptions on the swapper internal logic

In addition to the docs, this has been validated by the current v1 reverse swap. The fact that that `pay` in v1 asks for the `fees_hash` and uses it as arg in the create request, means this `fees_hash` must be valid. Otherwise, `pay` would fail. If however `getpairs` shows a different hash, and we now call `pay` with a slightly older hash, it will also fail. So the hash remains the same for as long as the fee promise is valid.

Despite all this, should I
- remove this validation from `pay`
- remove the `fees_hash` field from `onchain_payment_limits` response + from the `prepare` arg + remove this validation from `prepare`
?

dangeross: > This strongly depends on how the app handles this:
> 
> * call `onchain_payment_limits` once on app startup, then always use same limits
> * call `onchain_payment_limits` when needed, but cache results
> * call `onchain_payment_limits` for each new swap
> * call `onchain_payment_limits` for each new swap and when `FeePromiseChanged` error is thrown

I think the 3rd / 4th is more likely, there is no point caching it as it's rarely used and can be an initialisation function of a Swap component

> Despite all this, should I
> 
> * remove this validation from `pay`
> * remove the `fees_hash` field from `onchain_payment_limits` response + from the `prepare` arg + remove this validation from `prepare`
>   ?

I would say the 2nd. For me it doesn't make sense to keep hold of the hash as it's not a promise (or did I get this wrong?). The SDK could also check the amount is within the current limits without the hash, then return the `prepare` response with fees and hash. Having the hash here (and passed to `pay`) is a fair indicator the app has shown the fees to the user and to use them when calling `pay`

roeierez: I agree with Ross. I think we should:
1. remove the fee_hash from the onchain_payment_limits
2. in pay_onchain, create the swap (no need to call fetch_fees again) and then validate the amounts with the returned data. No need to validate that the hash is the same, boltz already does it when we create the swap.

libs/sdk-flutter/lib/breez_sdk.dart
```diff
@@ -322,12 +322,23 @@ class BreezSDK {
   /* On-Chain Swap API's */
 
   /// Creates a reverse swap and attempts to pay the HODL invoice
+  @Deprecated(
+    'Use payOnchain instead. '
+    'This method was deprecated after v0.3.1',
```
dangeross, line 327:
```suggestion
    'This method was deprecated after v0.3.2',
```

libs/sdk-flutter/lib/breez_sdk.dart
```diff
@@ -413,12 +424,23 @@ class BreezSDK {
   }
 
   /// Lookup the most recent reverse swap pair info using the Boltz API
+  @Deprecated(
+    'Use prepareOnchainPayment instead. '
+    'This method was deprecated after v0.3.1',
```
dangeross, line 429:
```suggestion
    'This method was deprecated after v0.3.2',
```

libs/sdk-bindings/src/breez_sdk.udl
ok300: Done: b6fb822f8f735c74136b767219d7b825dae6cd30

libs/sdk-core/src/models.rs
```diff
@@ -983,6 +997,55 @@ pub struct SendOnchainResponse {
...
+/// the resulting send and receive amounts.
+#[derive(Clone, Debug, Serialize)]
+pub struct PrepareOnchainPaymentResponse {
+    pub fees_hash: String,
+    pub fees_percentage: f64,
+    pub fees_lockup: u64,
+    pub fees_claim: u64,
+
+    pub send_amount_sat: u64,
+    pub receive_amount_sat: u64,
```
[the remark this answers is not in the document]
ok300: I don't think that's necessary anymore, as the entire `PrepareOnchainPaymentResponse` is used as arg in `pay_onchain`. The user doesn't deal with these fields anymore.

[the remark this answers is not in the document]
roeierez: But they are still visible, don't they? The app shows them in the UI.

libs/sdk-core/src/error.rs
```diff
@@ -621,6 +621,109 @@ impl From<SendPaymentError> for SdkError {
...
+
+    #[error("Payment failed: {err}")]
+    PaymentFailed { err: String },
+
+    #[error("Payment timeout: {err}")]
+    PaymentTimeout { err: String },
+
+    #[error("Service connectivity: {err}")]
+    ServiceConnectivity { err: String },
+}
```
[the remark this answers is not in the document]
ok300: Done: 56dd5bf50e6f9fdc37028c096881fed8ee866b20

I re-used `SendOnchainError` and added rustdocs for the new `OutOfRange` error type.

libs/sdk-core/src/models.rs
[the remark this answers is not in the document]
dangeross: I think from a developer perspective it's more understandable also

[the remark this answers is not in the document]
ok300: Done a310a6fc3d271db8bc21ec4018b128d1e6941d3e

ok300:
Ready for final review.

libs/sdk-core/src/breez_services.rs
```diff
@@ -930,6 +919,106 @@ impl BreezServices {
         Ok(self.btc_receive_swapper.refund_swap(req).await?)
     }
 
+    pub async fn onchain_payment_limits(&self) -> SdkResult<OnchainPaymentLimitsResponse> {
```
erdemyerebasmaz, line 922: Can we add `fetch` prefix to this function name?

roeierez (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -930,6 +919,106 @@ impl BreezServices {
...
+        })
+    }
+
+    /// Supersedes [BreezServices::fetch_reverse_swap_fees]
+    ///
+    /// ### Errors
+    ///
+    /// - `OutOfRange`: This indicates the send amount is outside the range of minimum and maximum
+    /// values returned by `onchain_payment_limits`. When you get this error, please first call
+    /// `onchain_payment_limits` to get the new limits, before calling this method again.
```
dangeross, line 936:
```suggestion
    /// values returned by [BreezServices::onchain_payment_limits]. When you get this error, please first call
    /// [BreezServices::onchain_payment_limits] to get the new limits, before calling this method again.
```

dangeross (commented):
Looks good. Just one small rustdocs fix

libs/sdk-core/src/breez_services.rs
erdemyerebasmaz: @roeierez I agree it's a bit long, 5 letter longer then the function it supersedes, but as far as I'm aware we have `fetch` prefixes on almost all functions that makes a network request.

dangeross (approved):
LGTM!

libs/sdk-core/src/breez_services.rs
roeierez: I don't think it is the case right now. A few examples: `open_channel_fee`, `lsp_info`, `recommended_fees`

erdemyerebasmaz (approved):
LGTM!

Thanks for all the hard work :)

PR #817, opened 2024-02-25
Lookup open channel bolt11 before fetching invoice

roeierez:
There is a bug in the following scenario:
1. User initiate a swap
2. Swap confirmed, sdk creates invoice that opens a channel, tries to redeem and fails for some reason.
3. User restores, swap redeem always fails, user has to refund.

The reason for the failure in step 3 is because when the sdk tries to create the invoice it fails on preimage already exists.
Recently we added a code that fallbacks to the node invoice but this invoice is the decreased amount in case of opening a channel so the swapper returns error for amount doesn't match.
The fix is to attempt first fetching the increased amount invoice from the persistent storage before falling back to the node.

### DIFF
--- a/libs/sdk-core/src/persist/transactions.rs
+++ b/libs/sdk-core/src/persist/transactions.rs
@@ -334,6 +334,26 @@ impl SqliteStorage {
             .optional()?)
     }
 
+    /// Look up a modified open channel bolt11 by hash.
+    pub(crate) fn get_open_channel_bolt11_by_hash(
+        &self,
+        hash: &str,
+    ) -> PersistResult<Option<String>> {
+        Ok(self
+            .get_connection()?
+            .query_row(
+                "
+          SELECT
+           o.open_channel_bolt11           
+          FROM sync.open_channel_payment_info o        
+          WHERE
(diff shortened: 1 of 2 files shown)


libs/sdk-core/src/persist/transactions.rs
```diff
@@ -960,6 +980,15 @@ fn test_ln_transactions() -> PersistResult<(), Box<dyn std::error::Error>> {
         150,
         "original_invoice",
     )?;
+
+    let open_channel_bolt11 = storage
+        .get_open_channel_bolt11_by_hash(payment_hash_with_lnurl_withdraw)?
+        .unwrap();
+    assert_eq!(open_channel_bolt11, "original_invoice");
+
+    let open_channel_bolt11 = storage.get_open_channel_bolt11_by_hash("non existing hadh")?;
```
dangeross, line 989:
```suggestion
    let open_channel_bolt11 = storage.get_open_channel_bolt11_by_hash("non existing hash")?;
```

dangeross (approved):
Looks good, just a spelling nit in the test

ok300 (approved):

PR #819, opened 2024-02-26
SDK Feature: Add Liquid Swaps

hydra-yse:
The following PR aims at adding support for liquid swaps. Mainly, the goal is to support these basic use-cases:

- [ ] Swapping L-BTC to fund our node (swap-in/normal submarine swap), either to pay invoices or open channels
- [ ] Swapping our node's funds onto the Liquid chain (swap-out/reverse submarine swap), both partially and totally (drain)

**Note:** As the repository is currently private, remember to change the import in `Cargo.toml` under sdk-core. The code is at the moment purely representative of what the structure should look like.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -9,6 +9,7 @@ use bip39::*;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::hashes::{sha256, Hash};
 use bitcoin::util::bip32::ChildNumber;
+use breez_sdk_liquid::{BreezWollet, WolletOptions};
 use chrono::Local;
 use futures::TryFutureExt;
 use log::{LevelFilter, Metadata, Record};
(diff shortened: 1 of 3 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -907,6 +909,40 @@ impl BreezServices {
...
+        let mut lock = self.liquid_wallet.lock().await;
+
+        let wollet = lock
+            .as_mut()
+            .ok_or(SdkError::Generic { err: "Liquid wallet not initialized".to_string() })?;
+
+        let invoice = self
+            .node_api
+            .create_invoice(amount_sat * 1000, "liquid swap".to_string(), None, None, None, None)
+            .await?;
```
hydra-yse, line 938: Of course, this will need to be expanded. Ideally, we want to persist the invoice created (and probably even mark it as a swap), and poll for the funds to be received/use a webhook (like we currently do in `receive_onchain`)

PR #823, opened 2024-02-28
Add an optional list arg to rescan_swaps()

ok300:
This PR adds the ability to rescan and persist the latest status of a single swap.


dangeross:
Does it need adding to libs/sdk-flutter/lib/breez_sdk.dart also?

ok300:
Superseded by the much simpler #824

PR #826, opened 2024-02-28
Automatic channel setup fee limit - Android extension

erdemyerebasmaz:
This PR makes automatic channel setup fee limit configurable via ExtensionSettings.


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
```diff
@@ -92,6 +97,10 @@ abstract class ForegroundService : SdkForegroundService, Service() {
      *  a ConnectRequest to be used to call the Breez SDK connect function. */
     abstract fun getConnectRequest(): ConnectRequest?
 
+    /** To be implemented by the application foreground service.
+     * Allows the user to override the default ExtensionSettings. */
+    abstract fun getExtensionSettings(): ExtensionSettings?
```
roeierez, line 99: NIT: I think the term extension is a bit confusing. Perhaps `getServiceConfig` or `getServiceSettings` ?

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
```diff
@@ -41,7 +42,12 @@ class LnurlPayInvoiceJob(
         try {
             request = Json.decodeFromString<LnurlInvoiceRequest>(LnurlInvoiceRequest.serializer(), payload)
             val nodeState = breezSDK.nodeInfo()
-            if (request.amount < 1000UL || request.amount > nodeState.inboundLiquidityMsats) {
+            val maxSendableMsats: ULong =
+                maxOf(
+                    ExtensionSettings.instance.autoChannelSetupFeeLimitMsats,
```
roeierez, line 47: Using a global singleton here is problematic. It makes the getServiceSettings abstraction redundant.
Also IMO it makes the input flow to the jobs complicated. A unified flow where all the job input is defined and passed in one place is better, this is why I think passing it as an argument on the intent makes sense.

roeierez (changes requested):

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -43,6 +44,11 @@ class LnurlPayInfoJob(
         try {
             request = Json.decodeFromString<LnurlInfoRequest>(LnurlInfoRequest.serializer(), payload)
             val nodeState = breezSDK.nodeInfo()
+            val maxSendableMsats: ULong =
+                maxOf(
+                    ExtensionSettings.instance.autoChannelSetupFeeLimitMsats,
+                    nodeState.inboundLiquidityMsats
```
dangeross, line 50: I think this confuses the user acceptable fee limit with the max amount that can be received. 

If a fee limit is set, the max sendable can be nodeState.maxReceivableMsat (minus channelsBalanceMsat?). Then before creating the invoice in the next step, if the amount is over the channelsBalanceMsat check the openChannelFee result to see if the feeMsat is within the user fee limit.

erdemyerebasmaz:
Closed in favor of #833

PR #830, opened 2024-02-29
Register webhook on manual LSP change

dangeross:
When manually switching to a new LSP using `connect_lsp(lsp_id)`, register the cached webhook with them


roeierez (approved):
LGTM

PR #833, opened 2024-02-29
LNURL-Payment: Channel setup fee limit

erdemyerebasmaz:
This PR makes automatic channel setup fee limit configurable via ServiceConfig.

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
@@ -67,9 +67,12 @@ class LnurlPayInfoJob(
                 nodeInfo.inboundLiquidityMsats,
                 maxReceivableMsatsThatFallsWithinFeeLimits,
             )
-            // Get the minimum sendable amount(in millisatoshis), can not be less than 1 sats
+            // Get the minimum sendable amount(in millisatoshis), can not be less than 1 or more than maxSendable
             val minSendable: ULong =
                 if (nodeInfo.inboundLiquidityMsats < 1000UL) ofp.minMsat else 1000UL
+            if (minSendable > maxSendable) {
+                throw Exception("Minimum sendable amount can't be greater than maximum sendable amount.")
+            }
             val plainTextMetadata = getString(
                 context,
                 LNURL_PAY_METADATA_PLAIN_TEXT,
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
@@ -69,7 +69,7 @@ class LnurlPayInfoJob(
             )
             // Get the minimum sendable amount(in millisatoshis), can not be less than 1 sats
             val minSendable: ULong =
-                if (nodeInfo.inboundLiquidityMsats == 0UL) ofp.minMsat else 1000UL
+                if (nodeInfo.inboundLiquidityMsats < 1000UL) ofp.minMsat else 1000UL
             val plainTextMetadata = getString(
                 context,
                 LNURL_PAY_METADATA_PLAIN_TEXT,
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
@@ -2,6 +2,7 @@ package breez_sdk_notification.job
 
 import android.content.Context
 import breez_sdk.BlockingBreezServices
+import breez_sdk.OpenChannelFeeRequest
 import breez_sdk_notification.Constants.DEFAULT_LNURL_PAY_INFO_NOTIFICATION_TITLE
 import breez_sdk_notification.Constants.DEFAULT_LNURL_PAY_METADATA_PLAIN_TEXT
 import breez_sdk_notification.Constants.DEFAULT_LNURL_PAY_NOTIFICATION_FAILURE_TITLE
@@ -12,6 +13,7 @@ import breez_sdk_notification.Constants.NOTIFICATION_CHANNEL_LNURL_PAY
 import breez_sdk_notification.NotificationHelper.Companion.notifyChannel
 import breez_sdk_notification.ResourceHelper.Companion.getString
 import breez_sdk_notification.SdkForegroundService
+import breez_sdk_notification.ServiceConfig
 import breez_sdk_notification.ServiceLogger
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
@@ -40,6 +42,7 @@ class LnurlPayInfoJob(
     private val fgService: SdkForegroundService,
     private val payload: String,
     private val logger: ServiceLogger,
+    private val config: ServiceConfig,
 ) : LnurlPayJob {
     companion object {
         private const val TAG = "LnurlPayInfoJob"
@@ -49,7 +52,24 @@ class LnurlPayInfoJob(
         var request: LnurlInfoRequest? = null
         try {
             request = Json.decodeFromString(LnurlInfoRequest.serializer(), payload)
-            val nodeState = breezSDK.nodeInfo()
+            // Get the fee parameters offered by the LSP for opening a new channel
+            val ofp =
+                breezSDK.openChannelFee(OpenChannelFeeRequest(amountMsat = null)).feeParams
+            // Calculate maximum receivable amount that falls within fee limits(in millisatoshis)
+            val feeLimitMsats: ULong = config.autoChannelSetupFeeLimitMsats
+            val nodeInfo = breezSDK.nodeInfo()
+            val maxReceivableMsatsThatFallsWithinFeeLimits = minOf(
+                nodeInfo.maxReceivableMsat,
+                feeLimitMsats / (ofp.proportional / 1000000u),
+            )
+            // Calculate maximum sendable amount(in millisatoshis)
+            val maxSendable = maxOf(
+                nodeInfo.inboundLiquidityMsats,
+                maxReceivableMsatsThatFallsWithinFeeLimits,
+            )
+            // Get the minimum sendable amount(in millisatoshis), can not be less than 1 sats
+            val minSendable: ULong =
+                if (nodeInfo.inboundLiquidityMsats == 0UL) ofp.minMsat else 1000UL
             val plainTextMetadata = getString(
                 context,
                 LNURL_PAY_METADATA_PLAIN_TEXT,
@@ -58,8 +78,8 @@ class LnurlPayInfoJob(
             val response =
                 LnurlPayInfoResponse(
                     request.callbackURL,
-                    nodeState.inboundLiquidityMsats,
-                    1000UL,
+                    maxSendable,
+                    minSendable,
                     "[[\"text/plain\",\"$plainTextMetadata\"]]",
                     "payRequest",
                 )
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
@@ -55,19 +55,19 @@ class LnurlPayInfoJob(
             // Get the fee parameters offered by the LSP for opening a new channel
             val ofp =
                 breezSDK.openChannelFee(OpenChannelFeeRequest(amountMsat = null)).feeParams
-            // Calculate maximum receivable amount that falls within fee limits(in millisatoshis)
-            val feeLimitMsats: ULong = config.autoChannelSetupFeeLimitMsats
+            // Calculate the maximum receivable amount within the fee limit (in millisatoshis)
+            val feeLimitMsat: ULong = config.channelFeeLimitMsat
             val nodeInfo = breezSDK.nodeInfo()
-            val maxReceivableMsatsThatFallsWithinFeeLimits = minOf(
+            val maxReceivableMsatFeeLimit = minOf(
                 nodeInfo.maxReceivableMsat,
-                feeLimitMsats / (ofp.proportional / 1000000u),
+                feeLimitMsat / (ofp.proportional / 1000000u),
             )
-            // Calculate maximum sendable amount(in millisatoshis)
+            // Calculate the maximum sendable amount (in millisatoshis)
             val maxSendable = maxOf(
                 nodeInfo.inboundLiquidityMsats,
-                maxReceivableMsatsThatFallsWithinFeeLimits,
+                maxReceivableMsatFeeLimit,
             )
-            // Get the minimum sendable amount(in millisatoshis), can not be less than 1 or more than maxSendable
+            // Get the minimum sendable amount (in millisatoshis), can not be less than 1 or more than maxSendable
             val minSendable: ULong =
                 if (nodeInfo.inboundLiquidityMsats < 1000UL) ofp.minMsat else 1000UL
             if (minSendable > maxSendable) {
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
@@ -2,6 +2,7 @@ package breez_sdk_notification.job
 
 import android.content.Context
 import breez_sdk.BlockingBreezServices
+import breez_sdk.OpenChannelFeeRequest
 import breez_sdk.ReceivePaymentRequest
 import breez_sdk_notification.Constants.DEFAULT_LNURL_PAY_INVOICE_NOTIFICATION_TITLE
 import breez_sdk_notification.Constants.DEFAULT_LNURL_PAY_METADATA_PLAIN_TEXT
@@ -13,6 +14,7 @@ import breez_sdk_notification.Constants.NOTIFICATION_CHANNEL_LNURL_PAY
 import breez_sdk_notification.NotificationHelper.Companion.notifyChannel
 import breez_sdk_notification.ResourceHelper.Companion.getString
 import breez_sdk_notification.SdkForegroundService
+import breez_sdk_notification.ServiceConfig
 import breez_sdk_notification.ServiceLogger
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
@@ -38,6 +40,7 @@ class LnurlPayInvoiceJob(
     private val fgService: SdkForegroundService,
     private val payload: String,
     private val logger: ServiceLogger,
+    private val config: ServiceConfig,
 ) : LnurlPayJob {
     companion object {
         private const val TAG = "LnurlPayInvoiceJob"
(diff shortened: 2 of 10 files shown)


erdemyerebasmaz:
Merged the reviewed Android & iOS PR's for SDK notification handling package & automatic channel setup fee limit into one to reduce cognitive load & ease of development.

erdemyerebasmaz:
Moving @dangeross's comment regarding:
```
maxSendableMsats = max(serviceConfig.autoChannelSetupFeeLimitMsats, nodeInfo.inboundLiquidityMsats)
```
here as it applies to both implementations.

> I think this confuses the user acceptable fee limit with the max amount that can be received.
> 
> If a fee limit is set, the max sendable can be nodeState.maxReceivableMsat (minus channelsBalanceMsat?). Then before creating the invoice in the next step, if the amount is over the channelsBalanceMsat check the openChannelFee result to see if the feeMsat is within the user fee limit.

Yes, the implementation is currently faulty, it'll only accept payments within fee limit as it is.

Perhaps, we should only apply the fee limit if inbound liquidity is zero after making these calculations.
Do we need to factor in reserve amount as well? @roeierez love to hear your input on calculating the correct max sendable amount.

roeierez:
> Moving @dangeross's comment regarding:
> 
> ```
> maxSendableMsats = max(serviceConfig.autoChannelSetupFeeLimitMsats, nodeInfo.inboundLiquidityMsats)
> ```
> 
> here as it applies to both implementations.
> 
> > I think this confuses the user acceptable fee limit with the max amount that can be received.
> > If a fee limit is set, the max sendable can be nodeState.maxReceivableMsat (minus channelsBalanceMsat?). Then before creating the invoice in the next step, if the amount is over the channelsBalanceMsat check the openChannelFee result to see if the feeMsat is within the user fee limit.
> 
> Yes, the implementation is currently faulty, it'll only accept payments within fee limit as it is.
> 
> Perhaps, we should only apply the fee limit if inbound liquidity is zero after making these calculations. Do we need to factor in reserve amount as well? @roeierez love to hear your input on calculating the correct max sendable amount.

@erdemyerebasmaz  Yes we will need to calculate here the maximum amount (limited by `maxRecievableMsat`) for which the open channel fees is calculated no more than the fee limit. 
In order to calculate that we need to fetch the `open_channel_fee` let's call it `ocp`.
Then given an amount the fee calculation is:
`max(ocp.min_msat, amount * ocp.proportional / 1000000)`

So if `ocp.min_msat > fee_limit` then we return the inbound liquidity (because we can't open a channel).
Otherwise the max amount is calculated:
`min (maxRecievableMsat, fee_limit * 1000000 / ocp.proportional)`

erdemyerebasmaz:
> fee_limit * 1000000 / ocp.proportional

@roeierez Applied fee calculation changes: 37925dd67bf400c235b692757b7c7eee2ecfb135

We don't seem to have invoice amount when constructing `LnurlInfoResponse`, should we always return inbound liquidity as `maxSendable` in that case 🤔

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
```diff
@@ -48,7 +52,9 @@ class LnurlPayInvoiceJob(
         try {
             request = Json.decodeFromString(LnurlInvoiceRequest.serializer(), payload)
             val nodeState = breezSDK.nodeInfo()
-            if (request.amount < 1000UL || request.amount > nodeState.inboundLiquidityMsats) {
+            val ofp = breezSDK.lspInfo().openingFeeParamsList.values.first()
+            val maxSendableMsats: ULong = calculateMaxSendableMsats(nodeState, request, ofp)
+            if (request.amount < 1000UL || request.amount > maxSendableMsats) {
```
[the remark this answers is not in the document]
erdemyerebasmaz: Done eeebe66a8c5ebf051de2fc7c1ffec15562c8fcf3

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
```diff
@@ -106,4 +112,25 @@ class LnurlPayInvoiceJob(
...
+
+    private fun calculateMaxSendableMsats(
+        nodeInfo: NodeState,
+        lnurlInvoiceRequest: LnurlInvoiceRequest,
+        ofp: OpeningFeeParams?,
+    ): ULong {
+        var maxSendableMsats: ULong = nodeInfo.inboundLiquidityMsats
+
+        if (ofp != null) {
+            // Invoice amount is in Sats
```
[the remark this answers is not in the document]
erdemyerebasmaz: Removed this method on: eeebe66a8c5ebf051de2fc7c1ffec15562c8fcf3

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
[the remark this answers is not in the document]
erdemyerebasmaz: We're now using openChannelFee API to retrieve opening fee params & check whether the fees fall within fee limits.

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -50,6 +53,14 @@ class LnurlPayInfoJob(
         try {
             request = Json.decodeFromString(LnurlInfoRequest.serializer(), payload)
             val nodeState = breezSDK.nodeInfo()
+            val ofpResp =
+                breezSDK.openChannelFee(OpenChannelFeeRequest(amountMsat = null))
+            val feeLimitMsats: ULong = config.autoChannelSetupFeeLimitMsats
+            val maxSendable = minOf(
+                nodeState.maxReceivableMsat,
+                feeLimitMsats / ofpResp.feeParams.proportional * 100u
```
dangeross, line 61: Is this correct? If

feeLimitMsats: 5000 msat
proportional: 4000
```
feeLimitMsats / ofpResp.feeParams.proportional * 100u
```
5000 / 4000 * 100 = 125 msat

Shouldn't we get 1250000 with a 5000 fee limit?

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -58,7 +69,7 @@ class LnurlPayInfoJob(
             val response =
                 LnurlPayInfoResponse(
                     request.callbackURL,
-                    nodeState.inboundLiquidityMsats,
+                    maxSendable,
```
dangeross, line 72: The maxSendable should also be the max of feeLimitAmountMsat and the nodeState.inboundLiquidityMsats, because maybe it has more inbound already available than the calculated feeLimitAmountMsat

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -58,7 +69,7 @@ class LnurlPayInfoJob(
...
                     1000UL,
```
dangeross, line 73: Maybe the minSendable should be the ofpResp.minMsat if there is no inbound available, as it has to cover the min channel opening fee. What do you think?

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
erdemyerebasmaz: Proportional unit is in parts per million, for example: 0.50

feeLimitMsats: 5_000_000 / 0.50 * 100
maxSendable that falls within fee limits: 1_000_000_000

In reverse: 1_000_000_000 * 0.50 / 100 = 5_000_000 (feeLimitMsats)

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
erdemyerebasmaz: Yes, it makes sense.

erdemyerebasmaz: Done.
[Use OpeningFeeParams.minMsat for the minSendable](https://github.com/breez/breez-sdk/pull/833/commits/e6c29df8576b265dfe7ecbea1a7d0e52fba89368)

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
erdemyerebasmaz: So, we should calculate maxSendable as:
`max( min(maxReceivable, maxReceivableThatFallsWithinFeeLimit), inboundLiquidity)`
?

erdemyerebasmaz: Done 85db9cf6bd2e25487bf7fb6d1b86f78cf00962f0

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
dangeross:
```
sdk> open_channel_fee
{
  "fee_msat": null,
  "fee_params": {
    "min_msat": 2500000,
    "proportional": 4000,
    "valid_until": "2024-03-12T21:28:51.626Z",
    "max_idle_time": 4320,
    "max_client_to_self_delay": 432,
    "promise": "1f98ad865164d21005ee0828ac08f6300a7202c94b9712917486c372c444bf52353e0c06c8d86d61be465575295cbab63f89f7e266501194920253e02df4b87cf8"
  }
}
```

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
dangeross: But only if there is no inbound (no channels) and a channel is needed, else the minSendable is always something like 2_500_000, whereas if we have inbound minSendable should be 1_000

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
erdemyerebasmaz: Good catch.. should've checked it myself, don't know why I recalled proportional returning per
million value.

Corrected 5cfdc79415f2a58c4ab6ff91c1fef34240127d5f

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
erdemyerebasmaz: Ah, I see. Sorry, been jumping between a lot of tasks today.

Considering `fee_msat` is 0 if no channel is needed:
```rust
/// Opening fee for receiving the amount set in the [OpenChannelFeeRequest], in case it was set.
/// It may be zero if no new channel needs to be opened.
pub fee_msat: Option<u64>,
```
It should be
```kotlin
val minMsat = ofpResp.feeMsat?.let { if (it == 0UL) 1000UL else ofpResp.feeParams.minMsat }!!
if(request.amount < minMsat ...
```

erdemyerebasmaz: Done f9f6ab7b791ff66b7046d9e7a6b8b8fcd1972ad2

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -50,6 +53,19 @@ class LnurlPayInfoJob(
         try {
             request = Json.decodeFromString(LnurlInfoRequest.serializer(), payload)
             val nodeState = breezSDK.nodeInfo()
+            val ofpResp =
+                breezSDK.openChannelFee(OpenChannelFeeRequest(amountMsat = null))
+            val feeLimitMsats: ULong = config.autoChannelSetupFeeLimitMsats
+            val maxReceivableMsatsThatFallsWithinFeeLimits = minOf(
+                nodeState.maxReceivableMsat,
+                feeLimitMsats / (ofpResp.feeParams.proportional / 1000000u) / 100u,
```
dangeross, line 61: I still think this is wrong 🙂

feeLimitMsats: 10_000
proportional: 4_000
10_000 / (4_000 / 1_000_000) / 100 = 25_000

It should be 2_500_000

```suggestion
                feeLimitMsats * 1000000u / ofpResp.feeParams.proportional,
```
or
```suggestion
                feeLimitMsats / (ofpResp.feeParams.proportional / 1000000u),
```

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
```diff
@@ -47,8 +50,13 @@ class LnurlPayInvoiceJob(
         var request: LnurlInvoiceRequest? = null
         try {
             request = Json.decodeFromString(LnurlInvoiceRequest.serializer(), payload)
-            val nodeState = breezSDK.nodeInfo()
-            if (request.amount < 1000UL || request.amount > nodeState.inboundLiquidityMsats) {
+            val ofpResp =
+                breezSDK.openChannelFee(OpenChannelFeeRequest(amountMsat = request.amount))
+            val feeLimitMsats: ULong = config.autoChannelSetupFeeLimitMsats
+            val isFeeWithinLimits = ofpResp.feeMsat?.let { it != 0UL && it > feeLimitMsats }
```
dangeross, line 56: Is within limits if the ofp fee equals 0 or ofp fee is less or equal to the fee limit
```suggestion
            val isFeeWithinLimits = ofpResp.feeMsat?.let { it == 0UL || it <= feeLimitMsats }
```

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/Task/LnurlPayInvoice.swift
```diff
@@ -38,8 +38,11 @@ class LnurlPayInvoiceTask : LnurlPayTask {
         do {
             let plainTextMetadata = ResourceHelper.shared.getString(key: Constants.LNURL_PAY_METADATA_PLAIN_TEXT, fallback: Constants.DEFAULT_LNURL_PAY_METADATA_PLAIN_TEXT)
             let metadata = "[[\"text/plain\",\"\(plainTextMetadata)\"]]"
-            let nodeInfo = try breezSDK.nodeInfo()
-            if lnurlInvoiceRequest!.amount < 1000 || lnurlInvoiceRequest!.amount > nodeInfo.inboundLiquidityMsats {
+            let ofpResp = try breezSDK.openChannelFee(req: OpenChannelFeeRequest(amountMsat: lnurlInvoiceRequest!.amount))
+            let feeLimitMsats: UInt64 = config.autoChannelSetupFeeLimitMsats
+            let isFeeWithinLimits = ofpResp.feeMsat! != 0 && ofpResp.feeMsat! > feeLimitMsats
+            let minMsat: UInt64 = ofpResp.feeMsat! == 0 ? UInt64(1000) : ofpResp.feeParams.minMsat
+            if lnurlInvoiceRequest!.amount < minMsat || isFeeWithinLimits {
```
dangeross, line 45:
```suggestion
            if lnurlInvoiceRequest!.amount < minMsat || !isFeeWithinLimits {
```

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
erdemyerebasmaz: Using the second suggestion as it reads
`feeLimitMsats / proportional%`

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInvoice.kt
erdemyerebasmaz: Addressed both suggestions: [Correct isFeeWithinLimits condition](https://github.com/breez/breez-sdk/pull/833/commits/19bab0c0c38ad863bb81db040cb22ed69910fa1b)

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -49,7 +52,24 @@ class LnurlPayInfoJob(
...
+                feeLimitMsats / (ofp.proportional / 1000000u),
+            )
+            // Calculate maximum sendable amount(in millisatoshis)
+            val maxSendable = maxOf(
+                nodeInfo.inboundLiquidityMsats,
+                maxReceivableMsatsThatFallsWithinFeeLimits,
+            )
+            // Get the minimum sendable amount(in millisatoshis), can not be less than 1 sats
+            val minSendable: ULong =
+                if (nodeInfo.inboundLiquidityMsats == 0UL) ofp.minMsat else 1000UL
```
dangeross, line 72: Sorry for the back-and-forth (just thinking it through in my head), I think it makes more sense if the inbound is less than 1_000 then use the ofp.minMsat. 

My reasoning is if the available inbound is say 500, then the minSendable is set to 1_000. Receiving 1_000 would fail (not enough inbound / new channel needed) because it doesn't meet the minMsat requirement of the fee params. What do you think?
```suggestion
                if (nodeInfo.inboundLiquidityMsats < 1000UL) ofp.minMsat else 1000UL
```

erdemyerebasmaz: Makes sense👍
I still feel like we're still missing a couple edge cases😅

dangeross: Yeah... 
maybe if the maxSendable (calculated from the feeLimit) is less than the minSendable (fee param minMsat), could that happen? I guess if the provided feeLimit is too small it could. Then throw an error

erdemyerebasmaz: Good points.

Use minMsat from fee params if inbound liquidity is less than 1000 msats 0458995402d3ed884a8225c8b0fbb9ccee980fb2

Throw an exceptions is minSendable is greater than maxSendable a9e4f06bc1c5d2f053b1f505a6fbee9a474420b1

dangeross (commented):
My only minor nit is the use of `Msats`. `Msat` would be more consistent with the SDK. Otherwise looks good

erdemyerebasmaz:
> My only minor nit is the use of `Msats`. `Msat` would be more consistent with the SDK. Otherwise looks good

Applied [naming changes](https://github.com/breez/breez-sdk/pull/833/commits/cb80cdca5a2e6f000a2a9f6ac434809bba0019df) to increase readability & consistency

dangeross (approved):
Looks great!

PR #834, opened 2024-02-29
SDK notification handling packages

erdemyerebasmaz:
This PR refactors the notification handling code for
- Android into a package bundled with the breez_sdk maven dependency
- iOS into a package bundled with the cocoapods dependency

#### Android:
The `breez_sdk_notification` package is used to define a foreground service and jobs needed to be executed for each notification type. The package has a NotificationHelper to define notification channels and trigger notifications. A ResourceHelper to load resources via resource ID or meta-data, falling back to a default resource when one is not defined by the application.

---

#### iOS:
The package is used to define a notification service and tasks needed to be executed for each notification type. The package has a ResourceHelper to load resources via resource ID in the bundle Info.plist, falling back to a default resource when one is not defined by the application. A KeychainHelper to read secured values from the Keychain.

---

#### Flutter:
On `sdk-flutter` we add example usages on how to retrieve values from `Keychain`/`Keystore` using `flutter_secure_storage` package on `FlutterSecuredStorageHelper` to read and decode secured values native on Android side.
- **Todo:** `This is another PR's scope but we can` allow user to set `ConnectRequest` directly from Flutter.

---


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/LogHelper.kt
```diff
@@ -0,0 +1,28 @@
+package breez_sdk_notification
+
+import org.tinylog.kotlin.Logger
+import java.io.File
+
+@Suppress("unused")
+class LogHelper {
```
roeierez, line 7: @erdemyerebasmaz  @dangeross seems like it aim to abstract the use of tinylog from the user as implementation details but using this API will override the users logging dir if they use tinylog in their app since tinylog API is static.

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/SDKNotificationService.swift
```diff
@@ -0,0 +1,97 @@
...
+import XCGLogger
+import os.log
+
+open class SDKNotificationService: UNNotificationServiceExtension {
+    var breezSDK: BlockingBreezServices?
+    var contentHandler: ((UNNotificationContent) -> Void)?
+    var bestAttemptContent: UNMutableNotificationContent?
+    var currentTask: TaskProtocol?
+    
+    var logger = {
```
roeierez, line 11: How does the user configure the logging directory?

libs/sdk-flutter/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ dependencies:
   ### flutter_rust_bridge_codegen dependencies
   ffi: ^2.1.2 # requires Dart >=3.3.0(Flutter 3.19)
   flutter_rust_bridge: ^1.82.6 # breez-sdk requirement
+  flutter_secure_storage: ^9.0.0
```
roeierez, line 16: Don't we need to remove the flutter_secure_storage from the sdk?

roeierez (changes requested):

libs/sdk-flutter/pubspec.yaml
dangeross: This and the code in libs/sdk-flutter/android/src/main/kotlin/flutter_secured_storage need moving back to c-breez @erdemyerebasmaz

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/LogHelper.kt
dangeross: The application calls configureLogger(logDir) in onMessageReceived() to configure the logging directory
https://github.com/breez/c-breez/blob/68a6cdd61b60c5ae000b4d4a5b9d581fd2dbb0ae/android/app/src/main/kotlin/com/cBreez/client/BreezFcmService.kt#L28-L50

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/SDKNotificationService.swift
dangeross: By overriding the init() then calling logger.setup with the extension log file path
https://github.com/breez/c-breez/blob/68a6cdd61b60c5ae000b4d4a5b9d581fd2dbb0ae/ios/Breez%20Notification%20Service%20Extension/NotificationService.swift#L10-L17

Is there a better approach @roeierez?

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/LogHelper.kt
roeierez: I am thinking on an app that already use tiny log and have that configured with some configuration properties.  Seems like we overriding these properties here, don't we?

dangeross: Only if the application calls configureLogger, no?

dangeross: What happens if the application did not do any configuration? I'm thinking of RN applications also, where nothing would be already configured.

If you think its not needed in the SDK and the application should handle this, we can move it back to c-breez.

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/SDKNotificationService.swift
roeierez: What bothers me a bit is That apart from us being very opinionated about the logging frameworks the logging approach here and in android is different then the logging approach at the rust part of the sdk.
In the rust part the user provides a logger implementation (listener) and the app processes them the way it wants.
I wonder if we want to keep this approach also for kotlin and swift here.
Do you think passing a Logger interface implementation by the user will work? Perhaps another function that the user can override in the service/extension? We can also use the same SDKLogListener that we already have eliminate the need to introduce another interface.
We can still keep these implementations for users that want to use them out of the box.
I think If we do that we can allow ourselves to be more opinionated about the default implementation since the user has a way to override.

dangeross: I think its a pretty good idea. I would actually lean towards removing the implementations from the plugins, have c-breez as a reference logging implementation. It simplifies the plugins/dependencies and allows use to be less opinionated.

erdemyerebasmaz:
Feedbacks on this PR are addressed on #846 due to their scope.

libs/sdk-bindings/bindings-swift/Package.swift
```diff
@@ -7,16 +7,15 @@ let package = Package(
     name: "bindings-swift",
     platforms: [
         .macOS(.v12),
-        .iOS(.v11),
+        .iOS(.v14),
```
dangeross, line 10: Sorry I forgot this. Can you change this also back to v11?

roeierez (approved):
LGTM

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/BreezSDKConnector.swift
```diff
@@ -0,0 +1,53 @@
...
+    static func unregister() {
+        BreezSDKConnector.queue.sync { [] in
+            BreezSDKConnector.sdkListener = nil
+        }
+    }
+    
+    static func connectSDK(connectRequest: ConnectRequest, logger: LogStream?) throws -> BlockingBreezServices? {
+        if let logger = logger {
+            try setLogStream(logStream: logger)
+        }
```
dangeross, line 37: I wonder if we can let the application call `setLogStream`? 

I notice in Kotlin the application can use `SdkLogInitializer.initializeNodeLogStream()` if they want to setup SDK logging. I assume that is optional and they can initialise it (or not) how they want. Should we do the same in Swift?

libs/sdk-react-native/breez_sdk.podspec
```diff
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => "11.0" }
+  s.platforms    = { :ios => "14.0" }
```
dangeross, line 13: This can be lower back to 11 also

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/BreezSDKConnector.swift
erdemyerebasmaz: We can, but `initializeNodeLogStream` was implemented due to a necessity on Kotlin.

**Context:**
Apps crash if `setLogStream` is called when `LogStream` is already set on `Kotlin`, but this is not an issue with `Swift` as this statement will fail and continue.
`try setLogStream(logStream: logger)`

Still, if users call `setLogger`, we're setting node log stream to that `LogStream` which may not be what all our users want.

erdemyerebasmaz: Removed c838429e5e47912d3700c5f130ff5f045501baaf

libs/sdk-bindings/bindings-swift/Package.swift
erdemyerebasmaz: b60d13ad9375f7e451bed5290a66c28ff2f29014

libs/sdk-react-native/breez_sdk.podspec
erdemyerebasmaz: Lowered b60d13ad9375f7e451bed5290a66c28ff2f29014

dangeross (approved):
LGTM!

PR #839, opened 2024-02-29
Add Error rustdocs descriptions

### LINKED ISSUE #837: Improve error descriptions in rustdocs
Add an error description and optional example for each error variant.

https://breez.github.io/breez-sdk/breez_sdk_core/error/index.html

dangeross:
Add a short description of each public enum error variant.

Fixes #837

### DIFF
--- a/libs/sdk-core/src/error.rs
+++ b/libs/sdk-core/src/error.rs
@@ -13,12 +13,17 @@ pub type SdkResult<T, E = SdkError> = Result<T, E>;
 /// Error returned by [crate::breez_services::BreezServices::connect]
 #[derive(Debug, Error)]
 pub enum ConnectError {
+    /// This error is raised when a general error occurs not specific to other error variants
+    /// in this enum.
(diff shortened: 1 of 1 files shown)


ok300 (commented):
Maybe it's a bit too verbose, in that the name of the error is obvious from the enum value? If so, the first 2 lines of every new rustdoc could be removed.

What do you think?

dangeross:
> Maybe it's a bit too verbose, in that the name of the error is obvious from the enum value? If so, the first 2 lines of every new rustdoc could be removed.
> 
> What do you think?

Maybe it's a bit verbose, I added the titles because Relai were asking for the error meanings

ok300:
From the chat, it looks like the description is what they mainly wanted.

dangeross:
> From the chat, it looks like the description is what they mainly wanted.

Yep, I'll make the change

ok300 (approved):

PR #846, opened 2024-03-05
Remove log library dependencies from Notification handling package

erdemyerebasmaz:
This PR removes log library dependencies from notification handling packag.

We no longer enforce a logging library decision with notification handling package with this change and users now can call `setLogger` with a `LogStream` of their choice that utilizes their own logging library to listen in on notification extension logs.

#### SDK
- Add optional tag field to LogEntry

#### Android
- [x] Removed tinylog library
- [x] Added `setLogger(logger: LogStream)` which will be used log notification extension logs
- Updated Kotlin, gradle dependencies

#### iOS
- [x] Removed XCGLogger & KeychainAccess dependency
  - Removed KeychainHelper
- [x] Added `setLogger(logger: LogStream)` which will be used log notification extension logs
- Added BreezSdkNotifications to product targets

#### Flutter
- [x] Removed `flutter_secure_storage` & it's helper classes from Flutter plugin
- Updated Flutter dependencies to latest


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/LogHelper.kt
```diff
@@ -1,28 +1,28 @@
...
-import org.tinylog.kotlin.Logger
-import java.io.File
+import breez_sdk.setLogStream
+import kotlinx.coroutines.CoroutineScope
 
 @Suppress("unused")
-class LogHelper {
-    companion object {
-        private const val TAG = "LogHelper"
+object LogHelper {
```
roeierez, line 7: Would be great to get rid of the `LogHelper` and the `SdkLogListener` as they are static and also implement subscribe/unsubscribe for multiple listeners which I don' think belong to the sdk level.
How about removing these two classes and Implement a function on the `ForegroundService` `setLogListener` which the user can use.

roeierez (changes requested):

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/LogHelper.kt
dangeross: I agree, I think its simpler if the application could set LogStream listeners for the SDK and the notification plugin (add setLogStream to foreground service), then the application can tag the log entries as they want and no changes to sdk-core are needed

erdemyerebasmaz: Logger is now non-static. edf235294e449b3cf241de05971d67a831c221ab

roeierez (approved):
LGTM

libs/sdk-core/src/models.rs
```diff
@@ -438,6 +438,7 @@ pub(crate) trait ReverseSwapServiceAPI: Send + Sync {
 /// Internal SDK log entry
 #[derive(Clone, Debug)]
 pub struct LogEntry {
+    pub tag: Option<String>,
```
dangeross, line 441: Are we still keeping the `tag`? It seems unused in the SDK itself

erdemyerebasmaz: It's only utilized on plugin packages for now and I believe we'll have a second iteration before next sdk release for the `ServiceLogger` on plugin packages where `LogStream` & `LogEntry` is not used at all.

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ServiceLogger.kt
```diff
@@ -0,0 +1,20 @@
+package breez_sdk_notification
+
+import android.util.Log
+import breez_sdk.LogEntry
+import breez_sdk.LogStream
+
+class ServiceLogger(private val logger: LogStream?) {
```
roeierez, line 7: Why do we need ServiceLogger and not pass directly LogStream? The tag can be constant for logs coming from the extension.

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/ServiceLogger.swift
```diff
@@ -0,0 +1,45 @@
...
+#if DEBUG && true
+fileprivate var os_log = Logger(
+    subsystem: Bundle.main.bundleIdentifier!,
+    category: "ServiceLogger"
+)
+#else
+fileprivate var os_log = Logger(OSLog.disabled)
+#endif
+
+class ServiceLogger {
```
roeierez, line 13: Same here, I don't see why we need the ServiceLogger class instead of directly use LogStream.

roeierez (changes requested):

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/ServiceLogger.swift
erdemyerebasmaz: Resolving as discussed on Slack. 

Decided to keep and make changes that makes ServiceLogger an internal class on iOS, the behavior is consistent with Android this way.

- Re-added `setLogger` on iOS that takes a `LogStream` to initialize `ServiceLogger`.
- Removed tag from `LogEntry` to avoid unwanted changes for SDK users
  - Default logs will have proper tags but the user has to set an umbrella tag if they use their custom LogStream.

roeierez (approved):

libs/sdk-core/src/models.rs
erdemyerebasmaz: Removed tag from LogEntry with latest changes.

PR #847, opened 2024-03-06
Add Windows python wheels

dangeross:
This PR adds to the CI building the x86_64 and x86 python wheels for Windows

Action Run: https://github.com/breez/breez-sdk/actions/runs/8169628513

I successfully tested installing the wheel on Windows 10 using python 3.11 and running `mnemonic_to_seed`. It's also been tested using python 3.10 by Cannabolt.


roeierez (approved):
Great addition.

.github/workflows/build-bindings-windows.yml
```diff
@@ -59,12 +59,31 @@ jobs:
...
     - name: Archive release
       uses: actions/upload-artifact@v3
       with:
         name: sdk-bindings-${{ matrix.target }}
-        path: libs/target/${{ matrix.target }}/release/breez_sdk_bindings.dll
+        path: |
+          libs/target/${{ matrix.target }}/release/breez_sdk_bindings.dll
+          libs/target/${{ matrix.target }}/release/msvcp140.dll
+          libs/target/${{ matrix.target }}/release/vcruntime140.dll
+          libs/target/${{ matrix.target }}/release/vcruntime140_1.dll
```
JssDWt, line 86: Note: these will also be included in the C# package now. Question is whether that's problematic or not.
For golang I guess they're not included, because they're not git added before tagging the go release.

JssDWt (commented):
Looks good!
Two small comments/questions.

.github/workflows/build-bindings-windows.yml
dangeross: I'll look into this

dangeross: Changed it so the redistributable DLLs are only included when building the windows wheels

JssDWt (approved):
Looks good. 
It appears there is a conflict. I don't understand why tbh.

PR #849, opened 2024-03-07
Fix `prepare_redeem_onchain_funds` fee calculation

dangeross:
Convert the redeemable amount to satoshi before calculating the transaction fee.  Should we also consider deducting the 546 min-emergency-funds CLN needs to reserver to close anchor channels?

Reference:
```
{"sweepTxFeeSat": 364500, "sweepTxWeight": 486}
```
https://mempool.space/tx/561cbc548fa6c96df15415a3bd6627847dfd07c2639203c2b6fe0251695c1f77


ok300 (approved):

ok300:
>  Should we also consider deducting the 546 min-emergency-funds CLN needs to reserver to close anchor channels?

Looks like CLN is working on reducing it to zero, I don't think we need any special handling on our side.

PR #850, opened 2024-03-07
Combine `max_reverse_swap_amount` and `onchain_payment_limits`

ok300:
Since both methods deal with reverse swap limits, it would make sense to merge them into one.

This PR includes the `max_reverse_swap_amount` output as a field in the `onchain_payment_limits` response, and deprecates `max_reverse_swap_amount`.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -438,6 +438,7 @@ dictionary ReverseSwapFeesRequest {
 dictionary OnchainPaymentLimitsResponse {
     u64 min_sat;
     u64 max_sat;
+    u64 max_sat_with_current_channels;
 };
 
 enum SwapAmountType {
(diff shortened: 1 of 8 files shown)


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -438,6 +438,7 @@ dictionary ReverseSwapFeesRequest {
 dictionary OnchainPaymentLimitsResponse {
     u64 min_sat;
     u64 max_sat;
+    u64 max_sat_with_current_channels;
```
roeierez, line 441: I think  it is better to have one field max_sat and set there the minimum of max amount based on the users balance and the value returned from the swapper.

roeierez (changes requested):
Good idea! Left one comment

libs/sdk-bindings/src/breez_sdk.udl
ok300: Good idea, done: ee8a9ef0856b7d0d98af4891944e3dc532929510

I also covered the case when `max_reverse_swap_amount` would return a value below the Boltz `min`. With the naive implementation it would lead to

```
min_sat: 50_000, // From Boltz
max_sat: 0       // min(Boltz max, max_reverse_swap_amount)
```

Both `min_sat` and `max_sat` are now set to 0 in that case, to indicate no reverse swap is possible.

erdemyerebasmaz (approved):
LGTM

dangeross (approved):
Looks good

roeierez (approved):

PR #854, opened 2024-03-13
Prevent RN UI thread lock on calling connect

dangeross:
Move setLogStream and connect into a thread to prevent UI lock


roeierez (approved):
LGTM

PR #857, opened 2024-03-13
docs: update react-native example

karliatto:
Based on typescript types updating documentation for react-native.

https://github.com/breez/breez-sdk/blob/main/libs/sdk-react-native/src/index.ts#L486


erdemyerebasmaz:
Thanks @karliatto 

CC: @dangeross @JssDWt Can we also catch compile errors for RN example during CI workflows and/or part of makefile commands on sdk-react-native?

dangeross (approved):
Thanks for the PR @karliatto!

PR #858, opened 2024-03-13
Update `channels_balance_msat` immediately after a payment arrives 

### LINKED ISSUE #845: Delay between `BreezEvent::InvoicePaid` and balance getting updated
We noticed there's a delay between the event `BreezEvent::InvoicePaid` and the balance returned by `node_info()` starting to reflect the just-received funds. This is because, at the moment, updating the balance requires using a full sync.

Given that the just-received payment starts immediately being returned in `list_payments()`, it would make sense to also update the local view of the balance before issuing the `InvoicePaid` event.

andrei-21:
Hi. Here is an idea how to handle issue with the balance not being up to date when a payment received.
I propose to expose an additional parameter `not_synced_balance_msat` such that different clients can act as they need (ignore or add the value to the balance). I do not feel that changing `node_info()` is the right way to do that, because the method has it is own guarantees. 

Closes #845.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1462,11 +1462,16 @@ impl BreezServices {
                                               debug!("invoice stream got new invoice");
                                               if let Some(gl_client::signer::model::greenlight::incoming_payment::Details::Offchain(p)) = i.details {
                                                   let payment: Option<crate::models::Payment> = p.clone().try_into().ok();
-                                                  if payment.is_some() {
+                                                  if let Some(ref payment) = payment {
                                                       let res = cloned
                                                           .persister
-                                                          .insert_or_update_payments(&vec![payment.clone().unwrap()], false);
-                                                      debug!("paid invoice was added to payments list {:?}", res);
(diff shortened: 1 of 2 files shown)


roeierez:
@andrei-21 the invoice paid event is always received before the balance changes on greenlight (and on clightning).
This is why I feel comfortable to update the node info and not introducing another API method.

roeierez:
That's right andrei, I see your point now. Thinking about it perhaps inbound liquidity lagging a few seconds behind is not really an issue.
Due to the nature of a remote node thereanyway  could be cases where we are not updated in realtime andperhaps it is a good trade-off for keeping it simple and intuitive. What do you think?

roeierez:
I rather look at this not as a user preference matter but be use case driven. The whole problem started because of inconsistency (payment received but balance not updated) so as far as I see it updating the balance makes it more accurate and more consistent. The fact we update the inbound liquidity a few seconds later won't break anything in the product and in fact also after sync we can't guarantee that (since sync is not atomic).
And since we can't (and don't) guarantee atomicity I feel comfortable with  updating the balance only and eliminate the need to explain  what "unsynced balance" stands for.

roeierez (approved):
LGTM

PR #860, opened 2024-03-14
Fix integer division issues & consider cases where proportional fee and fee limit is set to 0

erdemyerebasmaz:
Fixes https://github.com/breez/c-breez/issues/823

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
@@ -58,10 +58,15 @@ class LnurlPayInfoJob(
             // Calculate the maximum receivable amount within the fee limit (in millisatoshis)
             val feeLimitMsat: ULong = config.channelFeeLimitMsat
             val nodeInfo = breezSDK.nodeInfo()
-            val maxReceivableMsatFeeLimit = minOf(
-                nodeInfo.maxReceivableMsat,
-                feeLimitMsat / (ofp.proportional / 1000000u),
-            )
+            val proportionalPercent = ((ofp.proportional.toDouble() / 1000000.0))
+            val maxReceivableMsatFeeLimit = if (proportionalPercent != 0.0 && feeLimitMsat != 0UL) {
+                minOf(
+                    nodeInfo.maxReceivableMsat,
+                    (feeLimitMsat.toDouble() / proportionalPercent).toULong()
+                )
+            } else {
+                nodeInfo.maxReceivableMsat
+            }
             // Calculate the maximum sendable amount (in millisatoshis)
             val maxSendable = maxOf(
                 nodeInfo.inboundLiquidityMsats,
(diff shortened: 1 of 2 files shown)


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/LnurlPayInfo.kt
```diff
@@ -58,10 +58,15 @@ class LnurlPayInfoJob(
             // Calculate the maximum receivable amount within the fee limit (in millisatoshis)
             val feeLimitMsat: ULong = config.channelFeeLimitMsat
             val nodeInfo = breezSDK.nodeInfo()
-            val maxReceivableMsatFeeLimit = minOf(
-                nodeInfo.maxReceivableMsat,
-                feeLimitMsat / (ofp.proportional / 1000000u),
-            )
+            val proportionalPercent = ((ofp.proportional.toDouble() / 1000000.0))
+            val maxReceivableMsatFeeLimit = if (proportionalPercent != 0.0 && feeLimitMsat != 0UL) {
```
dangeross, line 62: I think it's enough to check only proportionalPercent,. If feeLimitMsat is 0 then the maxReceivableMsatFeeLimit is 0 and the maxSendable will be equal to inboundLiquidityMsat. I think this is correct
```suggestion
            val maxReceivableMsatFeeLimit = if (proportionalPercent != 0.0) {
```

erdemyerebasmaz: That's true but my reasoning was to treat this feature as disabled when feeLimitMsat is set to 0 where in this case `maxReceivableMsatFeeLimit` should resolve to `maxReceivableMsat`.

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/Task/LnurlPayInfo.swift
```diff
@@ -49,7 +49,12 @@ class LnurlPayInfoTask : LnurlPayTask {
             // Calculate the maximum receivable amount within the fee limit (in millisatoshis)
             let feeLimitMsat: UInt64 = config.channelFeeLimitMsat
             let nodeInfo = try breezSDK.nodeInfo()
-            let maxReceivableMsatFeeLimit = min(nodeInfo.maxReceivableMsat, feeLimitMsat / (UInt64(ofp.proportional) / 1000000))
+            let proportionalPercent = Double(ofp.proportional) / 1000000.0
+            let maxReceivableMsatFeeLimit = (proportionalPercent != 0.0 || feeLimitMsat != 0)
```
dangeross, line 53:
```suggestion
            let maxReceivableMsatFeeLimit = (proportionalPercent != 0.0 && feeLimitMsat != 0)
```

dangeross (changes requested):

libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/Task/LnurlPayInfo.swift
erdemyerebasmaz: Good catch!

dangeross (approved):

PR #861, opened 2024-03-14
use swapper client instead of fund manager client for swap in

JssDWt:
The fund manager client maps to `AddFundInitLegacy` on the server. The legacy swapper has lower maximum amounts for swaps, namely 900k instead of 4M sat. The swapper client maps to `AddFundInit` on the server, with a 4M sat limit.

Also, this PR caps the `max_allowed_deposit` for `receive_onchain` to `max_receivable_msat`

### DIFF
--- a/libs/sdk-core/src/swap_in/swap.rs
+++ b/libs/sdk-core/src/swap_in/swap.rs
@@ -221,6 +221,9 @@ impl BTCReceiveSwap {
             node_state.max_receivable_msat as i64,
             swap_reply.max_allowed_deposit,
         );
+        if max_allowed_deposit < swap_reply.min_allowed_deposit {
+            return Err(SwapError::Generic(anyhow!("No allowed deposit amounts")));
+        }
         let swap_info = SwapInfo {
             bitcoin_address: swap_reply.bitcoin_address,
             created_at: SystemTime::now()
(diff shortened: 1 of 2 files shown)


erdemyerebasmaz (approved):
LGTM!

ok300 (approved):

ok300:
Fixes https://github.com/breez/c-breez/issues/811

roeierez (approved):
LGTM

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -217,6 +217,10 @@ impl BTCReceiveSwap {
             return Err(SwapError::Generic(anyhow!("Wrong address: {address_str}")));
         }
 
+        let max_allowed_deposit = std::cmp::min(
```
roeierez, line 220: Now that we calculate the max_allowed_deposit I think we should check if it is bellow min_allowed_deposit and return error in such case.

roeierez (changes requested):

libs/sdk-core/src/swap_in/swap.rs
JssDWt: Done

roeierez (approved):
LGTM

Issue #859, opened 2024-03-14
INVOICE_AMOUNT_MISMATCH when trying to redeem swap


JssDWt:
Sync issues can cause a swap redeem to fail.
- trying to redeem a swap where an open channel is needed
- inside receive_payment, the invoice creation fails, however, the invoice is actually stored on the node
- this redeem errors, because the invoice creation failed
- the next time we try to redeem, the invoice is found on this line of code https://github.com/breez/breez-sdk/blob/main/libs/sdk-core/src/swap_in/swap.rs#L477
- however, this is the 'lower amount invoice', so the lsp fees are deducted. This won't work for redeeming the swap, because for the redeem the amount in the utxos should match the invoice.

Potential solution:
- delete the invoice on the node if the invoice cannot be found in the sync database and create a new one rather than fetching it from the node.
- at invoice creation, label the invoice as an open channel invoice. Then, when retrieving the invoice, recreate the larger invoice from the invoice on the node,
- otherwise, if that's not possible, it seems refund is the only option

PR #862, opened 2024-03-15
swap: fix invoice amount mismatch

JssDWt:
Fixes https://github.com/breez/breez-sdk/issues/859 by doing the following:
- when an invoice is created on the node, 'metadata' is added to the label if the channel is an open channel invoice, containing the payer amount.
- When we fall into the case where the preimage already exists when trying to create an invoice, we take that invoice from the node, and construct the larger invoice from it based on the stored metadata.

When wrapping the invoice in this place, it is unclear which steps had already occurred. Did the payment get registered to the LSP already? Was the larger invoice already stored? These functions, registering and storing, are idempotent, so we can simply re-execute them in this situation. Therefore the newly extracted function to wrap the invoice is able to do the registration at the lsp and persisting the payer invoice.

I _think_ this covers our scenario.
It does not fix existing failing swaps due to the amount mismatch though, these swaps have to be refunded after the timelock. This only works for new swaps that may eventually fall into the same issue.

State: 
- 'normal' swap tested
- swap with invoice on the node: untested

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -733,16 +733,14 @@ impl NodeAPI for Greenlight {
             .first()
             .cloned()
             .and_then(|invoice| {
-                invoice.bolt11.map(|bolt11| {
-                    serde_json::from_str::<InvoiceLabel>(&invoice.label)
-                        .map(|label| FetchBolt11Result {
-                            bolt11,
-                            payer_amount_msat: label.payer_amount_msat,
-                        })
+                invoice.bolt11.map(|bolt11| FetchBolt11Result {
+                    bolt11,
+                    payer_amount_msat: serde_json::from_str::<InvoiceLabel>(&invoice.label)
+                        .map(|label| label.payer_amount_msat)
                         .ok()
+                        .flatten(),
                 })
-            })
-            .flatten();
+            });
 
         Ok(result)
     }
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -62,6 +62,12 @@ pub(crate) struct Greenlight {
     persister: Arc<SqliteStorage>,
 }
 
+#[derive(Serialize, Deserialize)]
+struct InvoiceLabel {
+    pub unix_milli: u128,
+    pub payer_amount_msat: Option<u64>,
+}
+
 impl Greenlight {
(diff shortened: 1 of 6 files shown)


JssDWt:
Made the invoice label json, so it's more easily parsed.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -733,8 +732,19 @@ impl NodeAPI for Greenlight {
             .invoices
             .first()
             .cloned()
-            .and_then(|res| res.bolt11);
-        Ok(found_bolt11)
+            .and_then(|invoice| {
+                invoice.bolt11.map(|bolt11| {
+                    serde_json::from_str::<InvoiceLabel>(&invoice.label)
```
roeierez, line 737: If the label doesn't exist (or json parse fail) this function will return None? Do we want to return the invoice with payer amount equals to the invoice amount in that case?

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: Good catch!/We should return the invoice with payer amount equals None in that case.

JssDWt: Done in https://github.com/breez/breez-sdk/pull/862/commits/fd3fa20f91e396d47c3694ac7187725f1fe5004d

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -178,7 +178,7 @@ impl SqliteStorage {
         let con = self.get_connection()?;
         let mut prep_statement = con.prepare(
             "
-        INSERT INTO sync.open_channel_payment_info (
+        INSERT OR IGNORE INTO sync.open_channel_payment_info (
```
roeierez, line 181: No problem with that, just curious if you encountered this error in a real run.

roeierez (approved):
LGTM

libs/sdk-core/src/persist/transactions.rs
roeierez: I see... now that the call was moved to wrap_open_channel_invoice.

JssDWt: Yeah it can be called multiple times, so should be idempotent

dangeross (approved):
LGTM!

PR #863, opened 2024-03-15
Use Singleton pattern on BreezSDK using factory constructor

erdemyerebasmaz:
Apply Singleton pattern on `BreezSDK` using factory constructor.

- Move `SdkConfig` extension to it's own file.
- Export `bridge_generated.dart`, `breez_sdk.dart`, `exceptions.dart` & `sdk_config.dart` files through `sdk.dart`
  - Filter out generated binding classes from `bridge_generated.dart`
  
---

I assumed an SDK partner encountered this issue because we were lazily creating a new instance on our documentation for Dart snippets and if taken as 1:1 example, it's likely to encounter this issue. However, it may very well be a [symbol stripping issue on iOS](https://docs.flutter.dev/platform-integration/ios/c-interop#stripping-ios-symbols).

> Calling this function returns an error:
await BreezSDK().mnemonicToSeed(mnemonicCode);
```
Exception: Subclasses of
'FlutterRustBBreezSdkCorePlatform singletons - there should not be two instances
(runtimeType=BreezSdkCorePlatform)
```

---

Either way, I expect these changes to
- take away the resource/instance management responsibility away from users,
- reduce import statements,
- simplify API access.


erdemyerebasmaz:
App has no issues but tests are failing with 
`Invalid argument(s): Failed to lookup symbol 'store_dart_post_cobject': dlsym(RTLD_DEFAULT, store_dart_post_cobject): symbol not found
`
on C-Breez, which needs to be addressed before this PR gets merged.

erdemyerebasmaz:
> App has no issues but tests are failing with
> 
> > Invalid argument(s): Failed to lookup symbol 'store_dart_post_cobject': dlsym(RTLD_DEFAULT, store_dart_post_cobject): symbol not found
> 
> on C-Breez, which needs to be addressed before this PR gets merged. [stacktrace](https://github.com/breez/breez-sdk/files/14619556/stacktrace.txt)
> 
> Related: [Testing Library functions in a flutter package with Native Functions using Dart::ffi](https://stackoverflow.com/questions/66641396/testing-library-functions-in-a-flutter-package-with-native-functions-using-dart)

Actually, after investigating out test cases & how they are structured on C-Breez, I can say this is not a blocker.

roeierez (commented):
@erdemyerebasmaz I agree with fixing the imports and testing with the real library instead of mock data. I am a bit hesitating  regarding the singleton at this level. 
Given the original user error was due to symbol stripping, can you elaborate on what moving to singleton fixes here?

erdemyerebasmaz:
> @erdemyerebasmaz I agree with fixing the imports and testing with the real library instead of mock data. I am a bit hesitating regarding the singleton at this level. Given the original user error was due to symbol stripping, can you elaborate on what moving to singleton fixes here?

That's true but I believe the issue can be reproduced if there's more than once instance of BreezSDK running on app layer. I will be running tests to prove it next week. 

By moving to Singleton, 
- we take [resource management](https://github.com/breez/c-breez/pull/827/files#diff-167a33cdb99bd92a4e8373d76d696e9d3f4ce8a1eae391d2e6f2a585c0f2e3fcL55) & [log & event stream management](https://github.com/breez/c-breez/pull/827/files#diff-e61eb31d013d12616f5532636a88cfa63631dda8f7829e5424e68542214d1608L56-L59) responsibilities away from users.
- ensure there's only a single instance of `BreezSDK `running at a time that's [globally accessible](https://github.com/breez/c-breez/pull/827/files#diff-6a61d7a628afe314babfb813ebe1f26a1b5958b419955b14dbee4dc760fcf76dL14-R49)

PR #865, opened 2024-03-15
don't redeem on in_progress_swap

JssDWt:
The `in_progress_swap` function gets the current state of a swap, but doesn't try to execute pending swaps. This is potentially called many times, each request doing a call to the server. If the server is trying to pay out a swap already, the `in_progress_swap` function will error, while everything is fine. This PR lets the background service handle pending swaps, and makes the `in_progress_swap` function only return the current state. It does still check the mempool for potentially new transactions.


roeierez (approved):
LGTM

PR #868, opened 2024-03-17
Reuse grpc connection

roeierez:
It turns out we create a new channel with every breez server request. The network overhead for the API itself is 100ms - 200ms while the connection establishment is 1-3 seconds (depends on the load in the client).
There are two fixes here:
1. Use connect_lazy to establish a connection only on first call and cloning the channel when needed.
2. Take advantage of the first 5-6 seconds that we wait for the node to start and ping the breez server to establish the tcp connection.

Overall this saves ~3 seconds on startup in most cases.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -67,6 +67,7 @@ use crate::BuyBitcoinProvider::Moonpay;
 use crate::*;
 
 use self::error::ConnectError;
+use self::grpc::PingRequest;
 
 pub type BreezServicesResult<T, E = ConnectError> = Result<T, E>;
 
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -2031,10 +2032,20 @@ impl BreezServicesBuilder {
         );
 
         // breez_server provides both FiatAPI & LspAPI implementations
-        let breez_server = Arc::new(BreezServer::new(
-            self.config.breezserver.clone(),
-            self.config.api_key.clone(),
-        ));
+        let breez_server = Arc::new(
+            BreezServer::new(self.config.breezserver.clone(), self.config.api_key.clone())
+                .map_err(|e| ConnectError::Generic {
```
dangeross, line 2037: Would it make sense to change this to a ServiceConnectivity error? I know it's technically not "connecting" at this point, but would be a misconfiguration on the breez server url.
```suggestion
                .map_err(|e| ConnectError::ServiceConnectivity {
```

roeierez: Makes sense. Done

dangeross (approved):
LGTM!

ok300 (approved):

PR #873, opened 2024-03-18
Add in_progress_onchain_payments() wrapper

ok300:
Add a method to check in progress onchain payments, that matches the reverse swap v2 naming.

The new `in_progress_onchain_payments` is a wrapper around the existing (v1) `in_progress_reverse_swaps`, which is now marked as deprecated.


roeierez:
@ok300 I think we are also missing the flutter binding and the plugin wrapper code.

roeierez:
@ok300 I think we are missing also the flutter plugin wrapper code.

roeierez (approved):
LGTM

PR #874, opened 2024-03-19
Apply `env_logger` module filters on Dart & UniFFI binding loggers

erdemyerebasmaz:
This PR addresses the issue where [env_logger](https://docs.rs/env_logger/latest/env_logger/)'s module filters were not applying to logs listened through Dart & UniFFI binding loggers.

#### Changelist:
- Filter Dart & UniFFI binding logs through `env_logger` module filters
- Move loggers under `logger.rs`
- Make filter levels customizable for each logger
- Log node-logs according to their level
- Apply changes to bindings & generate them 
  - Create `LevelFilter` struct that's usable from bindings.
  - Add optional `filter_level` param to binding logger init & setters
  - Map `models:LevelFilter` to `log::LevelFilter`.
    - Set `LevelFilter::TRACE` as default filter level

#### TODO:
- [x] Fix RN Bindings:
  - https://github.com/breez/breez-sdk-rn-generator/pull/23
- Apply changes to:
  - [x] notification extension packages
  - [x] flutter plugin
  - [x] example projects
  - [x] test cases


erdemyerebasmaz:
Depends on:
- https://github.com/breez/breez-sdk-rn-generator/pull/23 (tests pass using locally built bindings)
-Go & C# bindings to be publish for tests to pass

libs/sdk-core/src/breez_services.rs
```diff
@@ -1532,7 +1528,12 @@ impl BreezServices {
                          log_message_res = log_stream.message() => {
                           match log_message_res {
                            Ok(Some(l)) => {
-                            info!("node-logs: {}", l.line);
+                            match l.line.to_ascii_lowercase() {
+                                s if s.starts_with("info") => info!("node-logs: {}", l.line.split_at(4).1.trim_start()),
+                                s if s.starts_with("debug") => debug!("node-logs: {}", l.line.split_at(5).1.trim_start()),
+                                s if s.starts_with("unusual") => trace!("node-logs: {}", l.line.split_at(7).1.trim_start()),
```
erdemyerebasmaz, line 1534: Level mapping of "unusual" level node logs needs to be reviewed, that's why I left this condition even tho it behaves the same with default case.

Here are a couple of sample logs.
`UNUSUAL plugin-gl-plugin-internal: JSON-RPC interface not yet available. Delaying 50ms. IoError(Os { code: 2, kind: NotFound, message: \"No such file or directory\" })`

`UNUSUAL plugin-gl-plugin-internal: Waiting to re-establish a channel: Deadline reached while waiting for at least one channel to be re-established\n\nStack backtrace:\n   0: ...`

`UNUSUAL 0264a62a4307d701c04a46994ce5f5323b1ca28c80c66b73c631dbcb0990d6e835-channeld-chan#4: Adding HTLC 36 too slow: killing connection`

`UNUSUAL 02c811e575be2df47d8b48dab3d3f1c9b0f6e16d0d40b5ed78253308fc2bd7170d-chan#3: Peer permanent failure in CHANNELD_NORMAL: channeld: received ERROR error channel 7a68f688d7b789d29012fcdcf0fa22e966d87d803c322a3ae71c43cdd78c9b09: Funding transaction spent`

`UNUSUAL 02c811e575be2df47d8b48dab3d3f1c9b0f6e16d0d40b5ed78253308fc2bd7170d-chan#1: Peer permanent failure in ONCHAIN: Funding transaction spent`

`UNUSUAL plugin-gl-plugin-internal: Timeout expired waiting`

Should they be upgraded to `Warn` or `Error`?

libs/sdk-core/src/logger.rs
```diff
@@ -0,0 +1,321 @@
...
+pub fn init_uniffi_logger(
+    log_stream: Box<dyn LogStream>,
+    filter_level: Option<BindingLevelFilter>,
+) {
+    INIT_UNIFFI_LOGGER.call_once(|| {
+        UniFFILogger::set_log_stream(log_stream, filter_level);
+    });
+}
+
+pub struct UniFFILogger {
```
roeierez, line 166: Why taking the specific loggers out of their specific crates?
Doesn't UniFFILogger belong to the sdk-bindings crate?

roeierez (changes requested):

libs/sdk-core/src/logger.rs
erdemyerebasmaz: Not necessarily, the loggers are not exposed through bindings.

I had brought them together to see the differences to find the root cause of the issue & I prefer having them together but we can bring them back to `uniffi_bindings.rs` & `bindings.rs` with little to no change.

libs/sdk-core/src/logger.rs
```diff
@@ -0,0 +1,321 @@
...
+rustls::anchors=warn,
+rustls::conn=warn,
+rustls::client::common=warn,
+rustls::client::hs=warn,
+rustls::client::tls13=warn,
+rustyline=warn,
+rusqlite_migration=warn,
+tower::buffer::worker=warn,
+vls_protocol_signer=warn,
+vls_protocol_signer::handler::HandlerBuilder::do_build=warn
```
JssDWt, line 56: NOTE: I've changed the log filters in #1098

libs/sdk-core/src/logger.rs
```diff
@@ -0,0 +1,321 @@
...
+
+fn init_env_logger(target: Option<Target>, filter_level: Option<LevelFilter>) -> Logger {
+    let mut binding = Builder::new();
+    let builder = binding
+        .parse_filters(ENV_LOGGER_FILTER)
+        .format(|buf, record| {
+            writeln!(
+                buf,
+                "[{} {} {}:{}] {}",
+                Local::now().format("%Y-%m-%d %H:%M:%S%.3f"),
```
JssDWt, line 67: Let's use UTC if possible with ISO8601 format

libs/sdk-core/src/logger.rs
JssDWt: I agree putting these in the bindings crate seems logical

JssDWt (commented):
Concept ACK! A few nits. (And ofcourse this needs a rebase, which will be painful)

libs/sdk-core/src/logger.rs
erdemyerebasmaz: Good point! Done.

[Switch timestamp format from local to UTC in ISO 8601 format](https://github.com/breez/breez-sdk-greenlight/pull/874/commits/48c1ea809b4c3ffaa54a89cb7f61cc1f50b19e4a)

libs/sdk-core/src/logger.rs
erdemyerebasmaz: Adding the missing `sdk_common=debug,` should be enough imo.

We can make the necessary individual changes after making comparisons if there are any.

libs/sdk-core/src/logger.rs
erdemyerebasmaz: Done.
- [Move Dart & UniFFI loggers back to their respective places](https://github.com/breez/breez-sdk-greenlight/pull/874/commits/8f661bb4cd812ad01068eefd95540b77dd2d9398)

libs/sdk-core/src/binding.rs
```diff
@@ -51,6 +51,85 @@ use crate::{
     SignMessageResponse, StaticBackupRequest, StaticBackupResponse,
 };
 
+use lazy_static::lazy_static;
+use log::{
+    max_level, set_boxed_logger, set_max_level, warn, Log, Metadata, Record, STATIC_MAX_LEVEL,
+};
+use parking_lot::RwLock;
```
JssDWt, line 58: What's the reason to use `parking_lot` here? Can we use `std::sync::RwLock` or `tokio::sync::RwLock`?

libs/sdk-core/src/breez_services.rs
```diff
@@ -1756,7 +1752,14 @@ impl BreezServices {
                          log_message_res = log_stream.message() => {
                           match log_message_res {
                            Ok(Some(l)) => {
-                            info!("node-logs: {}", l.line);
+                            let prefix_len = l.line.find(':').map_or(0, |len| len + 2);
+                            let log_message = l.line.split_at(prefix_len).1.trim_start();
+                            match l.line.to_ascii_lowercase().as_str() {
+                                s if s.starts_with("unusual") => error!("node-logs: {}", log_message),
```
JssDWt, line 1758: I think `unusual` can be `warn` 

And `broken` needs to be added as `error`

libs/sdk-core/src/logger.rs
```diff
@@ -0,0 +1,185 @@
...
+breez_sdk_core::input_parser=warn,
+breez_sdk_core::persist::reverseswap=info,
+breez_sdk_core::reverseswap=info,
+sdk_common=debug,
+gl_client::node=info,
+gl_client::node::service=info,
+gl_client::tls=info,
+gl_client::scheduler=info,
+gl_client::signer=info,
+gl_client=debug,
```
JssDWt, line 30: If I understand correctly, `debug` level logs would be filtered out regardless of the client's loglevel. I would put all these at debug, or even trace, because they would eventually be filtered out by the client log level.

libs/sdk-core/src/logger.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use std::fs::OpenOptions;
+use std::io::Write;
+
+use chrono::Utc;
+
+/* env_logger */
+
+const ENV_LOGGER_FILTER: &str = r#"
+debug,
```
JssDWt, line 18: This should either depend on client log level or be `info` in my opinion

JssDWt: Perhaps we should create one of these filter statements per log level

JssDWt (commented):
Added a few comments

dangeross (commented):
These RN codegen changes in https://github.com/breez/breez-sdk-rn-generator/pull/23 should be applied in `libs/sdk-bindings/bindings-react-native`

PR #879, opened 2024-03-20
Start the signer from the SDK

dangeross:
Changes the signer startup to handle upgrading the node and run the signer loop from within the SDK, purposely removing the wait for the scheduler has been scheduled notification.


dangeross:
Failure logs: 
[sdk.log](https://github.com/breez/breez-sdk/files/14666799/sdk.log)

roeierez (approved):
LGTM

PR #880, opened 2024-03-20
Reverse swap validation: fix service fee rounding

ok300:
The service fee formula deals with fractions of a sat.

This value has to be rounded up, not down, to match the actual Boltz service fee.


roeierez (commented):
Worth adding some unit tests for the calculation with different edge case numbers and test that the reverse calculation resolves to the expected service fee

ok300:
Done: 28f85d297359ace8ceb0e4fe6214f371988fa09f

erdemyerebasmaz (approved):
LGTM

PR #882, opened 2024-03-21
Fix sync edge case affecting outbound payments

ok300:
If `since_timestamp` is set due to a newly received payment, but an outbound payment was created shortly before that, then the filter would normally miss this outbound payment.

We therefore extend the filter to also check if `completed_at` happened after `since_timestamp`.


roeierez (approved):
LGTM

erdemyerebasmaz (approved):

dangeross (approved):

erdemyerebasmaz (approved):
🎉

roeierez (approved):
LGTM

PR #890, opened 2024-03-23
Consolidate reverse swap service fee calculation

ok300:
There were still a few places that were doing the fee calculation explicitly, instead of (re-)using the dedicated method.

This PR
- re-uses `get_service_fee_sat` (the "forward calculation" of the service fee)
- encapsulates and adds tests for `get_invoice_amount_sat` (the "reverse calculation" of the invoice amount)

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -825,7 +825,7 @@ impl BreezServices {
                 res.fees_claim = BTCSendSwap::calculate_claim_tx_fee(claim_tx_feerate)?;
             }
 
-            let service_fee_sat = swap_out::calculate_service_fee_sat(amt, res.fees_percentage);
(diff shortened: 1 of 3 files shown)


dangeross (commented):
Looks good, but I feel like the `test_invoice_amount_calculation_in_reverse` test would benefit from asserting against a defined calculated invoice amount, even if its not the same as the invoice_amount_sat param because of rounding

ok300:
Good point @dangeross , a `<=` comparison is too broad.

I refined the test to check for one of the two valid results, `invoice_amount_sat` or `invoice_amount_sat - 1`.

Does this address your point?

dangeross (approved):
LGTM. Are there any actions from the meeting discussion that need addressing?

ok300:
Not from my side. Will keep this open so @roeierez can confirm as well.

roeierez (approved):
LGTM

Issue #894, opened 2024-03-25
Unable to parse upper case Bech32 bitcoin addresses


erdemyerebasmaz:
Encountered this issue when investigating:
- https://github.com/breez/c-breez/issues/844

QR code we create on "Receive via BTC Address" page on C-Breez is uppercase as per:

> Bech32 addresses consisting of uppercase and digits results in more compact QR codes.
> NB. If we ever generate BIP21 URI's that have query parameters, this will have to be fixed so that
> we will not change the case of the parameters because BIP21 parameters are case sensitive.
> Ref. https://bitcoinops.org/en/bech32-sending-support/#creating-more-efficient-qr-codes-with-bech32-addresses
>
> CC: @roeierez

and they failed to get parsed using `parseInput` API.

roeierez:
@erdemyerebasmaz  I just use sdk.parse and passed the "received via btc address" output and it worked just fine.

ok300:
After some talks with @erdemyerebasmaz , it turns out the problem is when the `BITCOIN:` prefix is in uppercase.

I added a fix in #896.

ok300:
Fixed by #896

PR #896, opened 2024-03-26
Fix BIP21 parsing where the prefix is in uppercase

ok300:
An edge-case around parsing the `bitcoin:` prefix caused issues when this prefix was in uppercase.

This PR handles this, without affecting the casign of the optional BIP21 label or message.

### DIFF
--- a/libs/sdk-core/src/input_parser.rs
+++ b/libs/sdk-core/src/input_parser.rs
@@ -256,7 +256,7 @@ where
 fn prepend_if_missing(prefix: &str, input: &str) -> String {
     match input.to_lowercase().starts_with(prefix) {
         true => input.into(),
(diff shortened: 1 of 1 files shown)


ok300:
Fixes #894

erdemyerebasmaz (approved):
LGTM!

ok300:
Looks like this is the correct way, as described in the BIP:

> The scheme component ("bitcoin:") is case-insensitive, and implementations must accept any combination of uppercase and lowercase letters. The rest of the URI is case-sensitive, including the query parameter keys. 
>
> https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#abnf-grammar

erdemyerebasmaz (approved):

PR #897, opened 2024-03-26
Fix swap confirmed block to be the earliest.

roeierez:
This ensures the swap confirmation only considers these with block height greater than zero and also use safe subtraction with unsigned integers.
potentially fixes https://github.com/breez/breez-sdk-relai/issues/6

### DIFF
--- a/libs/sdk-core/src/swap_in/swap.rs
+++ b/libs/sdk-core/src/swap_in/swap.rs
@@ -352,24 +352,18 @@ impl BTCReceiveSwap {
             .chain_service
             .address_transactions(bitcoin_address.clone())
             .await?;
-        let confirmed_txs: Vec<OnchainTx> = txs
+        let utxos = get_utxos(bitcoin_address.clone(), txs.clone())?;
+        let confirmed_block = txs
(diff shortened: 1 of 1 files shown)


dangeross (approved):

ok300:
There was a very similar fix from @JssDWt a few days ago: https://github.com/breez/breez-sdk/pull/869 , looks like it was reverted or overwritten in the meantime?

ok300 (approved):

ok300:
Ah, this PR is merging into `v0.3.9` branch, @JssDWt 's fix was applied to `main`.

Maybe it's simpler to cherry-pick his commit?

roeierez:
> Ah, this PR is merging into `v0.3.9` branch, @JssDWt 's fix was applied to `main`.
> 
> Maybe it's simpler to cherry-pick his commit?

That's right but the code base has changed in a way that cherry picking doesn't match the current code base, this is why I created a new commit.

roeierez:
@ok300 nevertheless I matched the implementations as much as possible. I do think the Jesses's one was simpler.

PR #898, opened 2024-03-27
ChainService: Fallback to next mempool.space endpoint on error

ok300:
This PR adds support for fallback chainservice URLs.

This includes:
- fetching a list of mempool.space endpoints from the breez server
- changing `Config::mempoolspace_url` to an optional value. If the callers sets it when initializing the SDK, this value will be used. If no value is set (default), then we use the list.
- changing the chainservice methods to support calls that fallback on to the next URL, for both GET and POST requests

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -27,6 +27,7 @@ use tonic::{Request, Status};
 use crate::backup::{BackupRequest, BackupTransport, BackupWatcher};
 use crate::chain::{
     ChainService, Outspend, RecommendedFees, RedundantChainService, RedundantChainServiceTrait,
+    DEFAULT_MEMPOOL_SPACE_URL,
 };
 use crate::error::{
     LnUrlAuthError, LnUrlPayError, LnUrlWithdrawError, ReceiveOnchainError, ReceiveOnchainResult,
@@ -2078,7 +2079,10 @@ impl BreezServicesBuilder {
                     0 => {
                         // If we have no cached values, or we cached an empty list, fetch new ones
 
-                        let fresh_urls = breez_server.fetch_mempoolspace_urls().await?;
+                        let fresh_urls = breez_server
+                            .fetch_mempoolspace_urls()
+                            .await
+                            .unwrap_or(vec![DEFAULT_MEMPOOL_SPACE_URL.into()]);
                         persister.set_mempoolspace_base_urls(fresh_urls.clone())?;
                         fresh_urls
                     }
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -25,7 +25,9 @@ use tonic::transport::{Channel, Endpoint};
 use tonic::{Request, Status};
 
 use crate::backup::{BackupRequest, BackupTransport, BackupWatcher};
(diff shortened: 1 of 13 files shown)


ok300:
Fixes https://github.com/breez/breez-sdk-relai/issues/8

roeierez:
Thanks @ok300 here are some of my thoughts:
Another approach to implement redundancy without changing the MempoolSpace implementation is to create another implementation of the ChainService trait RedundantChainService that has a list of ChainServices inside and for every call loops between them, saving internally the last succeeded one as the first one for the next call.
It seem to me that it decouple the redundancy logic from the real chain service logic and also doesn't use unavailable services again and again.
One other advantage is that we can introduce another ChainService implementation (non mempool space) and we won't need to change any code related to the redundancy.
What do you think?

libs/sdk-core/src/breez_services.rs
```diff
@@ -2073,6 +2068,45 @@ impl BreezServicesBuilder {
...
+                        // If we have no cached values, or we cached an empty list, fetch new ones
+
+                        let fresh_urls = breez_server.fetch_mempoolspace_urls().await?;
+                        persister.set_mempoolspace_base_urls(fresh_urls.clone())?;
+                        fresh_urls
+                    }
+                    _ => {
+                        // If we already have cached values, return those
+
+                        // Start thread that refreshes cache in the background
```
dangeross, line 2086: The only disadvantage I see to this is that these are not then available to the current instance, only when disconnecting/connecting. But I guess that list will not change too much

ok300: Yes, that's the downside. The benefit is somewhat faster startup times, as we're not waiting for this call.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2073,6 +2071,48 @@ impl BreezServicesBuilder {
...
+                        let cloned_breez_server = breez_server.clone();
+                        let cloned_persister = persister.clone();
+                        tokio::spawn(async move {
```
roeierez, line 2095: Perhaps it is better to do that on start method? It is the point where currently we start all async tasks.

roeierez (changes requested):
Looks good, dropped a small comment.

libs/sdk-core/src/breez_services.rs
ok300: I tried to, but unfortunately the `start` method only has access to `BreezServices`.

The new `fetch_mempoolspace_urls` method is on `BreezServer` itself. It's not exposed by any API (`NodeAPI`, `LspAPI`, etc) because it didn't seem to belong in any of them.

Should I expose it in an API, so I can call it in `start`? If so, which API makes more sense?

roeierez: @ok300  Perhaps we don't need to expose it in any of these traits and just create a BreezServer instance on demand for that? The same way we do with service_health_check?
We can add a new trait for settings but I feel it is overkill for now.

ok300: Done: becd69ab685ab2eeef765b69ddf9cf2f82abcadb

roeierez (approved):
LGTM

dangeross (approved):

PR #900, opened 2024-03-27
Diagnostic data for supporting users.

roeierez:


libs/sdk-core/src/breez_services.rs
```diff
@@ -1850,6 +1858,32 @@ impl BreezServices {
...
+        let payments = serde_json::to_string(
+            &self
+                .persister
+                .list_payments(ListPaymentsRequest::default())?,
+        )?;
+        let channels = serde_json::to_string(&self.persister.list_channels()?)?;
+        let settings = serde_json::to_string(&self.persister.list_settings()?)?;
+        let reverse_swaps = serde_json::to_string(&self.persister.list_reverse_swaps()?)?;
+        let swaps = serde_json::to_string(&self.persister.list_swaps()?)?;
+        let list_id = serde_json::to_string(&self.persister.get_lsp_id()?)?;
```
dangeross, line 1873:
```suggestion
        let lsp_id = serde_json::to_string(&self.persister.get_lsp_id()?)?;
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -1042,6 +1042,14 @@ impl BreezServices {
         Ok(self.node_api.execute_command(command).await?)
     }
 
+    // Collects various user data from the node and the sdk storage.
+    // This is used for debugging and support purposes only.
+    pub async fn collect_user_data(&self) -> SdkResult<String> {
```
dangeross, line 1047: I would avoid calling it `collect_user_data`, especially after reading the Apple privacy manifest guide. Maybe something like `get_diagnostic_data`

erdemyerebasmaz:
> Does `sdk-flutter/lib/breez_sdk.dart` need updating also?

Yes, it does.

roeierez:
> Does `sdk-flutter/lib/breez_sdk.dart` need updating also?

Added

dangeross (approved):
LGTM

PR #906, opened 2024-03-28
Use swap event to complete swap in the background

roeierez:
Since both the job and the sdk are racing for redeeming the swap, one of them can fail and if the job fails we show failure to redeem although the redeem succeed.
It is better to rely on the swap event as the source of truth here.

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/RedeemSwap.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/RedeemSwap.kt
@@ -50,7 +50,7 @@ class RedeemSwapJob(
                     logger.log(TAG, "Received swap updated event: ${swapInfo.bitcoinAddress} current address: ${address} status: ${swapInfo.status}", "TRACE")
                     if (swapInfo.bitcoinAddress == address) {
                         if (swapInfo.paidMsat.toLong() > 0) {
-                            notifySuccessAndShutdown()
+                            notifySuccessAndShutdown(address)
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


libs/sdk-bindings/bindings-swift/Sources/BreezSDKNotification/Task/RedeemSwap.swift
```diff
@@ -21,26 +22,44 @@ class RedeemSwapTask : TaskProtocol {
...
         } catch let e {
             self.logger.log(tag: TAG, line: "failed to decode payload: \(e)", level: "ERROR")
             self.onShutdown()
             throw e
         }
-
+        
+        guard let address = swapAddress else {
+            self.logger.log(tag: TAG, line: "Failed to process swap notification: swap address not in payload", level: "ERROR")
+            self.onShutdown()
```
dangeross, line 55: I think a return is needed here
```suggestion
            self.onShutdown()
            return
```

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/RedeemSwap.kt
```diff
@@ -33,32 +34,43 @@ class RedeemSwapJob(
...
+                }
+
+                else -> {}
+            }
+        }
     }
 
-    override fun onEvent(e: BreezEvent) {}
+    private  fun notifySuccessAndShutdown() {
+        logger.log(TAG, "Swap address redeemed succesfully", "INFO")
```
erdemyerebasmaz, line 64: We do log address information on SwapUpdated event but I suggest we log which address was successful here as well.
```suggestion
    private  fun notifySuccessAndShutdown(address: String) {
        logger.log(TAG, "Swap address ${address} redeemed successfully", "INFO")
```

dangeross (approved):
LGTM

PR #907, opened 2024-03-28
Improve Android notification job management

dangeross:
This PR adds to the Android Notification Plugin
- Management of running jobs in a list
- Delay shutdown only when jobs list is empty
- Forward SDK events to all jobs from a service listener

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
@@ -6,7 +6,9 @@ import android.os.Handler
 import android.os.IBinder
 import android.os.Looper
 import breez_sdk.BlockingBreezServices
+import breez_sdk.BreezEvent
 import breez_sdk.ConnectRequest
+import breez_sdk.EventListener
 import breez_sdk.LogStream
 import breez_sdk_notification.BreezSdkConnector.Companion.connectSDK
 import breez_sdk_notification.Constants.MESSAGE_TYPE_ADDRESS_TXS_CONFIRMED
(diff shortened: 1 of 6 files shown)


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
```diff
@@ -46,19 +53,32 @@ abstract class ForegroundService : SdkForegroundService, Service() {
...
-    override fun pushbackShutdown() {
+    private fun resetShutdown() {
         shutdownHandler.removeCallbacksAndMessages(null)
-        shutdownHandler.postDelayed(shutdownRunnable, SHUTDOWN_DELAY_MS)
     }
 
-    override fun shutdown() {
+    private fun delayedShutdown() {
+        if (jobs.isEmpty()) {
+            shutdownHandler.postDelayed(shutdownRunnable, SHUTDOWN_DELAY_MS)
```
roeierez, line 77: What is the reason not to shutdown immediately here?

roeierez (changes requested):
Looks good! one question

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
dangeross: I thought maybe it's good to leave the SDK running a bit longer rather than immediately shutting down after the last job. What do you think?

roeierez: I think if the jobs signals then it means it has been completed so I don't see any particular reason right now.
I am OK with adding several seconds, can't hurt, but right now it adds 1 minute.

dangeross: Ok, I will remove the delay then

roeierez (approved):

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
erdemyerebasmaz: > What is the reason not to shutdown immediately here?
@roeierez The reason is to keep the service running for other tasks that may come while we have an SDK connection.

dangeross:
@erdemyerebasmaz can you review when you have chance?

erdemyerebasmaz (approved):
LGTM

PR #910, opened 2024-03-29
Add freezed annotation on SDK structs used through Dart bindings

erdemyerebasmaz:
- All SDK structs exposed through Dart bindings that reference another SDK struct has to be annotated. Including the referenced struct.

- Not needed on structs that only has supported types so enums are excluded. https://cjycode.com/flutter_rust_bridge/v1/feature/lang_simple.html

### DIFF
--- a/libs/sdk-core/src/binding.rs
+++ b/libs/sdk-core/src/binding.rs
@@ -9,7 +9,7 @@
 //!
 //! The first option loses the `SdkError` type. The second option keeps the type, which we can retrieve
 //! with `anyhow::Error::downcast_ref` (or equivalent Dart method). We therefore use the second approach.
-
+#![allow(unused_imports)]
 use std::future::Future;
 use std::sync::Arc;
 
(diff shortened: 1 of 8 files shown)


dangeross:
@erdemyerebasmaz with v2 of the flutter_rust_bridge do we still need to do this?

erdemyerebasmaz:
> flutter_rust_bridge

@dangeross I'm not exactly sure if v2 changes would also generate `toString` methods without these annotations.
> - Use arbitrary Rust and Dart types without manual intervention, even if they are not serializable or non-clone.
>  - Generate hashCode/equals for non-freezed structs.

 I'll setup an example project using v2 to find out.

roeierez:
@erdemyerebasmaz it seems that v2 also supports structure errors so we should consider seriously to upgrade.

erdemyerebasmaz:
> @erdemyerebasmaz it seems that v2 also supports structure errors so we should consider seriously to upgrade.

I didn't want to upgrade yet as it's in beta and we were getting ready for releasing c-breez. I'll look into it.

Should we merge this as there's an open PR on C-Breez that depends on this?

erdemyerebasmaz:
Superseded by #940

PR #914, opened 2024-04-05
Improve payment asynchronicity 

### LINKED ISSUE #908: Payment asynchronicity
Hi all,

i've been using the Breez SDK to implement an app that segregates the underlying node into accounts. Similar to LNBits, it is possible to create multiple accounts that share the same underlying node, each with its own balance and permissions.

When implementing the "outgoing payment" functionality, I noticed that I can run into concurrency issues between the moment the payment is sent and the moment the event is received because the sending function returns the payment_hash only after the payment has been processed.

In order to implement the account functionality, I need to correlate the payment with a user account, so I need to save the payment in DB and then be able to find it back once it is received in an event. The reason is that if a payment is sent and then we crash, once the event is read upon restart, we won't be able to correlate it with a user. 

What I would like:
- A way to add labels to a payment so I can add a unique payment identifier that is different than the `payment_hash`, and retrieve it in the event that would have these labels `attached`.
- An asynchronous function that sends the payment but does not wait for its success and returns the `payment_hash`.
- any a similar solution enabling payment asynchronicity.

I noticed that the `Payment` returned when the function is successful in `LnUrlPayResult` does not contain all the `Payment` details in a similar way than the event `PaymentSucceed`, so I need the event even in the event where we implement it synchronously.

dangeross:
This PR changes the following:

- Allow payment metadata to be set during successful `pay_lnurl`, `send_payment` or `send_spontaneous_payment` requests. This is specifically added for `pay_lnurl` where the payment hash is not known beforehand and therefore cannot associate the request to the `PaymentSucceed` event which is emitted before the request finishes. Also added to `send_payment` and `send_spontaneous_payment` for consistency. 
- Add the payment details to the `LnUrlPaySuccessData` response when `pay_lnurl` is successful to be consistent with `send_payment` and `send_spontaneous_payment`.

Fixes #908

### DIFF
--- a/libs/sdk-core/src/lnurl/pay.rs
+++ b/libs/sdk-core/src/lnurl/pay.rs
@@ -121,7 +121,7 @@ fn validate_invoice(user_amount_msat: u64, bolt11: &str, network: Network) -> Ln
 pub(crate) mod model {
     use crate::lnurl::error::{LnUrlError, LnUrlResult};
     use crate::lnurl::pay::{Aes256CbcDec, Aes256CbcEnc};
-    use crate::{ensure_sdk, input_parser::*};
+    use crate::{ensure_sdk, input_parser::*, Payment};
 
     use aes::cipher::{block_padding::Pkcs7, BlockDecryptMut, BlockEncryptMut, KeyIvInit};
     use anyhow::{anyhow, Result};
@@ -144,6 +144,7 @@ pub(crate) mod model {
     /// * `PayError` indicates that an error occurred while trying to pay the invoice from the LNURL endpoint.
     /// This includes the payment hash of the failed invoice and the failure reason.
     #[derive(Debug, Serialize, Deserialize)]
+    #[allow(clippy::large_enum_variant)]
     pub enum LnUrlPayResult {
         EndpointSuccess { data: LnUrlPaySuccessData },
         EndpointError { data: LnUrlErrorData },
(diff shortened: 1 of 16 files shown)


libs/sdk-core/src/lnurl/pay.rs
```diff
@@ -159,6 +160,7 @@ pub(crate) mod model {
     #[derive(Serialize, Deserialize, Debug)]
     pub struct LnUrlPaySuccessData {
         pub payment_hash: String,
+        pub payment: Payment,
```
hydra-yse, line 163: Why do we keep `payment_hash` given the new field ?

hydra-yse (commented):
Looking good! Just a small comment, everything else is fine

libs/sdk-core/src/lnurl/pay.rs
dangeross: > Why do we keep `payment_hash` given the new field ?

Good point, the `payment.id` can be used

roeierez:
@dangeross Looks good. I am thinking though on an edge case where the payment is pending for a long time and the caller gets timeout on send_payment.
The payment may be completed at a later time where in such case we will loose the metadata.
Perhaps a better idea is to save the metadata before we actually invoke the node send_payment instead of on_payment_completed?
Perhaps we can use the label field that can be sent as the pay payload to the node API for this purpose? The advantage is that we need less to sync accross apps.

dangeross:
> Perhaps a better idea is to save the metadata before we actually invoke the node send_payment instead of on_payment_completed? 

👍 

> Perhaps we can use the label field that can be sent as the pay payload to the node API for this purpose? The advantage is that we need less to sync accross apps.

@roeierez So keeping the metadata option but adding a label option also?

roeierez:
> @roeierez So keeping the metadata option but adding a label option also?

How about only using label for now? It is a lot more lightweight and solve the user case.

roeierez (approved):
LGTM

PR #915, opened 2024-04-07
Fix db deadlock

roeierez:
This PR fixes the deadlock that happens sometimes on the sqlite database.
The reason for the deadlock origins in the way we load payments.
Before this PR we select from the payments list and then to load the associated swaps and reverse swaps we execute a query for each payment.
That would be OK if we have reused the same connection to load the associations but since we opened a new connection it was blocked at times there was a pending writer waiting, result in a deadlock.
It is better to use join for such query and I tried to do that without duplicating the queries.
Currently the `sql_row_to_payment` function that creates a Payment from an SQL Row uses internally `sql_row_to_swap` and `sql_row_to_reverse_swap` to create the associated information, where both of these methods are used also to create the stand alone versions of swaps as well. The fields in the Row are keyed differently (using a given prefix) on the stand alone query than on the JOIN query.

### DIFF
--- a/libs/sdk-core/src/persist/transactions.rs
+++ b/libs/sdk-core/src/persist/transactions.rs
@@ -293,11 +293,11 @@ impl SqliteStorage {
            ON
             p.id = o.payment_hash
           LEFT JOIN ({swap_query}) as swaps
-            ON
-              json_extract(p.details, '$.payment_hash') = hex(swaps_payment_hash) COLLATE NOCASE
+           ON
+            p.id = hex(swaps_payment_hash) COLLATE NOCASE
           LEFT JOIN ({rev_swap_query}) as revswaps
-            ON
-              json_extract(p.details, '$.payment_preimage') = hex(revswaps_preimage) COLLATE NOCASE
+           ON
+            json_extract(p.details, '$.payment_preimage') = hex(revswaps_preimage) COLLATE NOCASE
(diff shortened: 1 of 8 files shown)


libs/sdk-core/src/persist/transactions.rs
```diff
@@ -283,54 +246,78 @@ impl SqliteStorage {
...
+           p.id = e.payment_id
+          LEFT JOIN sync.payments_metadata m
+          ON
+            p.id = m.payment_id
+          LEFT JOIN sync.open_channel_payment_info o
+           ON
+            p.id = o.payment_hash
+          LEFT JOIN ({swap_query}) as swaps
+            ON
+              json_extract(p.details, '$.payment_hash') = hex(swaps_payment_hash) COLLATE NOCASE
```
dangeross, line 297: Is it possible to `p.id` here instead?
```suggestion
              p.id = hex(swaps_payment_hash) COLLATE NOCASE
```

ok300 (approved):

ok300:
Fixes #916

libs/sdk-core/src/persist/transactions.rs
roeierez: Yes,  indeed better. I fixed that.

dangeross (approved):
LGTM

Issue #840, opened 2024-03-01
New "LnUrlPayWithZaps" type in the `InputType` enum


simongltr:
Hey Breez team, we are in the process of implementing Zaps in our app.

The function `parse_input` of the sdk returns a `LnUrlPay` `InputType` which is normal. The issue is that the `nostrPubkey` and `allowsNostr` fields (from [NIP57](https://github.com/nostr-protocol/nips/blob/master/57.md)) are removed because not part of the [lud06](https://github.com/lnurl/luds/blob/luds/06.md) spec. 

To avoid workarounds like fetching 2 times the same thing (1 to retrieve nostr zaps data, 1 by the `parse_input` function), I feel like it would make sense to add a new "LnUrlPayWithZaps" type in the `InputType` enum. It would contain the data from `LnUrlPay` and the `nostrPubkey` and `allowsNostr` fields.

Any other equivalent improvement is welcome.

Thank you,
Simon

JssDWt:
Other options could be
- adding `nostr_pubkey` and `allows_nostr` fields to the `LnUrlPayRequestData` struct
- adding a dictionary named `other` to the `LnUrlPayRequestData` struct, containing unknown fields

simongltr:
I let you make the final choice, having access to `nostr_pubkey` and `allows_nostr` is enough for us right now.

With option 2 future protocols data will be accessible without any change to the SDK though

roeierez:
@simongltr LNURLPay has a metadata field which arbitrary data can be encoded inside, did you consider making use of that? @JssDWt what do you think?

JssDWt:
@roeierez It's a standard in NIP-57 https://github.com/nostr-protocol/nips/blob/master/57.md, it contains fields inside the lnurlp response. So these fields are not in the metadata. Standard lnurlp response looks like this:

```
{
    "callback": string, 
    "maxSendable": number, 
    "minSendable": number, `maxSendable`
    "metadata": string, 
    "tag": "payRequest"
}
```

Response with zaps support looks like this:

```
{
    "callback": string, 
    "maxSendable": number, 
    "minSendable": number, `maxSendable`
    "metadata": string, 
    "tag": "payRequest",
    "nostrPubkey": string,
    "allowsNostr": boolean
}
```

So it's not in the metadata field.
All major lnurl providers support this. If you get a random lnurlp response from Alby or Walletofsatoshi for example, the fields are there.

PR #918, opened 2024-04-08
input parser: add nip-57 fields

JssDWt:
Adds fields to `LnurlPayRequestData` relevant for Nostr Zaps ([NIP-57](https://github.com/nostr-protocol/nips/blob/master/57.md)).

Tested manually the input parser with lnurlp with zap support and without zap support.

Fixes https://github.com/breez/breez-sdk/issues/840

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -626,6 +626,8 @@ dictionary LnUrlPayRequestData {
     string metadata_str;
     u16 comment_allowed;
     string domain;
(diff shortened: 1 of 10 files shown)


dangeross:
Aside from this PR to correctly parse the input, we could add an issue to:
1. Include nostrPubkey as a setting in c-breez (and include it in the lnurl info response)
2. Extend the lnurl server to pass the zap request to the notification plugin and include it in the invoice
3. Create the zap receipt on received payment

dangeross (approved):

PR #919, opened 2024-04-08
Cache LSP information

dangeross:
This PR caches the LspInformation and only refreshes it when a record of the opening fee params expires (hourly) or on reconnect. The aim is to help reduce the requests sent to the Breez server when we already have valid fee param promises.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -693,7 +693,7 @@ impl BreezServices {
         &self,
         req: OpenChannelFeeRequest,
     ) -> SdkResult<OpenChannelFeeResponse> {
-        let lsp_info = self.lsp_info().await?;
+        let lsp_info = self.lsp_info(false).await?;
         let fee_params = lsp_info
             .cheapest_open_channel_fee(req.expiry.unwrap_or(INVOICE_PAYMENT_FEE_EXPIRY_SECONDS))?
             .clone();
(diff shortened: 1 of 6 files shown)


dangeross:
> It seems that now we can get into a state where the expensive fee param will be chosen for the default one hour invoice?

No, if any of the fee param values are expired, it is deemed invalid and it requests an updated version from the server. It uses the same validate function as when validating the initial response from the server:
https://github.com/breez/breez-sdk/blob/64a928f0355d757b74c823a805b46d9f2faa07aa/libs/sdk-core/src/models.rs#L1203-L1211
and is re-validated when lsp_info() is used:
https://github.com/breez/breez-sdk/blob/14c8b3104e11c1cd7e74c9da5e9b0106fda174a7/libs/sdk-core/src/breez_services.rs#L1275-L1281

dangeross:
> @dangeross One of the requests to this is from invoice creation. We need that to match the invoice expiry. It seems that now we can get into a state where the expensive fee param will be chosen for the default one hour invoice?

Ah, I see I think, for the expiry of the invoice is needs to be valid

dangeross:
I've added a `skip_cache` param, but it this case previously it was getting the latest lsp information anyway.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2606,12 +2613,34 @@ impl PaymentReceiver {
...
+                    None
+                }
+            }
+        });
+    match persisted_lsp_info {
+        Some(lsp_info) => {
+            debug!("Using persisted LSP information");
+            Ok(lsp_info)
+        }
+        None => {
```
ok300, line 2633:
```suggestion
    let persisted_lsp_info = persister.get_lsp_information()?;
    let cached_lsp_info = match skip_cache {
        true => None,
        false => persisted_lsp_info.filter(|lsp_info| lsp_info.opening_fee_params_list.is_valid()),
    };

    match cached_lsp_info {
        Some(lsp_info) => {
            debug!("Using cached LSP info");
            Ok(lsp_info)
        }
        None => {
            debug!("Ignoring cached LSP info: skip_cache={skip_cache} lsp_info={lsp_info:?}");
```

ok300: Added a suggestion on how the first part of the method could be rewritten for clarity.

ok300:
IMO this looks good.

PR #921, opened 2024-04-09
Handle `max_allowed_deposit`

### LINKED ISSUE #871: Update `max_allowed_deposit` on previously created swaps if possible
If any swap address we've created in the past hasn't been used, we return that same address when `create_swap_address` is called(through `receive_payment`). 

Due to this behavior, changes introduced in
- #861

only apply to newly created swap addresses.

If it's possible to retrieve `max_allowed_deposit` using SwapperAPI without creating a new swap, we should update `max_allowed_deposit` of the unused swap with the latest limits.

In theory, you should only have a single unused address at a time and calling a new API to fetch latest maximum values for swaps created before this transitory period is perhaps going a bit overboard as these limits rarely change afaik.

ok300:
Fixes #871 

This PR adds the following behaviors:
- refactor: when new swaps are created, `max_allowed_deposit` is determined based on absolute max and the node `max_receivable_msat`
- new: when an existing swap is fetched, the `max_allowed_deposit` is recalculated, validated and persisted
- new: store the `max_allowed_deposit_abs` per swap, indicating the swapper max value at the moment this swap was created

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -555,6 +555,7 @@ dictionary SwapInfo {
     sequence<string> confirmed_tx_ids;
     i64 min_allowed_deposit;
     i64 max_allowed_deposit;
+    i64 max_allowed_deposit_abs;
     string? last_redeem_error;
     OpeningFeeParams? channel_opening_fees;
     u32? confirmed_at;
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -555,7 +555,7 @@ dictionary SwapInfo {
     sequence<string> confirmed_tx_ids;
     i64 min_allowed_deposit;
     i64 max_allowed_deposit;
-    i64 max_allowed_deposit_abs;
+    i64 max_swapper_payable;
     string? last_redeem_error;
     OpeningFeeParams? channel_opening_fees;
     u32? confirmed_at;
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -1374,10 +1374,12 @@ pub struct SwapInfo {
     pub unconfirmed_tx_ids: Vec<String>,
     /// Transaction IDs that have been confirmed on-chain.
     pub confirmed_tx_ids: Vec<String>,
-    /// The minimum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].   
+    /// The minimum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
     pub min_allowed_deposit: i64,
-    /// The maximum amount of sats one can send in order for the swap to succeed. Received from [SwapperAPI::create_swap].
+    /// The maximum amount of sats one can send in order for the swap to succeed. This is determined based on `max_allowed_deposit_abs` and the node's local balance.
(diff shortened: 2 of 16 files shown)


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -555,6 +555,7 @@ dictionary SwapInfo {
     sequence<string> confirmed_tx_ids;
     i64 min_allowed_deposit;
     i64 max_allowed_deposit;
+    i64 max_allowed_deposit_abs;
```
roeierez, line 558: NIT: rename to `max_swapper_payable` ? because it is actually the maximum the swapper is willing to pay.

libs/sdk-core/src/models.rs
```diff
@@ -1475,6 +1479,14 @@ impl SwapInfo {
             _ => SwapStatus::Initial,
         }
     }
+
+    pub(crate) fn validate_max_allowed_deposit(&self) -> SwapResult<()> {
```
roeierez, line 1483: NIT: validate_swap_limits ?

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -181,13 +181,38 @@ impl BTCReceiveSwap {
...
+            // Check max_allowed_deposit and, if it changed, persist and validate changes
+            let current_max = fn_max_allowed_deposit(unused_swap.max_allowed_deposit_abs);
+            let res_swap = match current_max == unused_swap.max_allowed_deposit {
+                true => unused_swap,
+                false => {
+                    info!("max_allowed_deposit for this swap has changed, updating it");
+                    let mut new_swap = unused_swap.clone();
+
+                    new_swap.max_allowed_deposit = current_max;
+                    new_swap.validate_max_allowed_deposit()?;
```
roeierez, line 206: If we are adding validation here perhaps it is a good idea to add a specific error? `UnsupportedSwapLimits` ?

roeierez (changes requested):
Looks good! Some small NITs

libs/sdk-bindings/src/breez_sdk.udl
ok300: Done: a29023c2e97b9c13a86c04944e04a7b021ee6eea

libs/sdk-core/src/swap_in/swap.rs
ok300: Done: a29023c2e97b9c13a86c04944e04a7b021ee6eea

roeierez (approved):
LGTM

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -427,6 +427,11 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
        ALTER TABLE swaps_info ADD COLUMN total_incoming_txs INTEGER;
        UPDATE swaps_info SET status = 0;
        ",
+       // Add max absolute value payable by the swapper. For existing swaps, initialize it to max_allowed_deposit.
+       "
+       ALTER TABLE sync.swaps ADD COLUMN max_swapper_payable INTEGER NOT NULL;
```
dangeross, line 432: Will this `NOT NULL` cause a constraint error if some swaps rows already exist?

roeierez: Good point. @ok300 perhaps we should add default zero here?

PR #926, opened 2024-04-12
Verify notification manager exist before starting service

dangeross:
This PR is a fix on v0.4.0 to pre-check the android notification manager exists before starting the foreground service. 

It could be that the device receives a push message but the application doesn't have permissions to post notifications. In this case the plugin is trying to create the foreground service notification to call `startForeground`, but results in the application crashing with:
```
android.app.RemoteServiceException$CannotPostForegroundServiceNotificationException: 
    Bad notification for startForeground
```
So we first check in the MessageService if we should start the foreground service by checking the existence on the notification manager.


erdemyerebasmaz (approved):
Great catch!

roeierez (approved):
LGTM

PR #931, opened 2024-04-14
Add Android short service timeout

dangeross:
This PR is a fix on v0.4.0 to add a Notification Plugin service timeout of 3 minutes to the running short service in Android. If the service runs for an extended period of time or does not respect the timeout of the short service, the application will be declared ANR (Application Not Responsive). We add the manual timeout instead of using `Service.onTimeout` because the plugin targets compile API level 33 and the function `onTimeout` was introduced in API level 34.

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
@@ -16,6 +16,7 @@ import breez_sdk_notification.Constants.MESSAGE_TYPE_LNURL_PAY_INFO
 import breez_sdk_notification.Constants.MESSAGE_TYPE_LNURL_PAY_INVOICE
 import breez_sdk_notification.Constants.MESSAGE_TYPE_PAYMENT_RECEIVED
 import breez_sdk_notification.Constants.NOTIFICATION_ID_FOREGROUND_SERVICE
+import breez_sdk_notification.Constants.SERVICE_TIMEOUT_MS
 import breez_sdk_notification.Constants.SHUTDOWN_DELAY_MS
 import breez_sdk_notification.NotificationHelper.Companion.notifyForegroundService
 import breez_sdk_notification.job.Job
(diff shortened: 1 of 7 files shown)


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/ForegroundService.kt
```diff
@@ -62,14 +64,27 @@ abstract class ForegroundService : SdkForegroundService, EventListener, Service(
     }
 
     /** Stop the service */
+    private val serviceTimeoutHandler = Handler(Looper.getMainLooper())
+    private val serviceTimeoutRunnable: Runnable = Runnable {
```
roeierez, line 68: @dangeross can you please elaborate when do we extend the serviceTimeout vs where we extend the shutdownHandler ?

dangeross: The serviceTimeout is extended (reset to 3 mins) on each call to `onStartCommand`. The shutdownHandler is stopped on a call to `onStartCommand` and started when the last job is complete. So the shutdownHandler keeps the service "warm" for 1 min before shutting down.

roeierez: Thanks for the explanation.
I see here that when a running service calls startForegroundService it gets an extended running period of another 3 minutes: https://developer.android.com/about/versions/14/changes/fgs-types-required#short-service
Do we use this mechanism?

dangeross: Yes, it's called in `onStartCommand`

dangeross: I think I have to change the implementation to use
```
ServiceCompat.startForeground(0, notification, FOREGROUND_SERVICE_TYPE_SHORT_SERVICE)
```

dangeross: After testing this, just `startForeground(0, notification)` is sufficient. Also ServiceCompat.startForeground is not available in API level 33

roeierez (approved):
LGTM

erdemyerebasmaz (approved):
LGTM

PR #934, opened 2024-04-16
Signer loop: exit when shutdown signal received during upgrade loop

ok300:
This PR makes the signer upgrade loop exit in case the shutdown signal is received.

The signer loop wrapper method (`Greenlight::run_forever`) has 2 parts, the upgrade loop (a) and the actual signer loop (b).

Inside it, the upgrade loop (a) has the following logic:
- try to upgrade the signer with `scheduler.maybe_upgrade`
  - if `maybe_upgrade` fails due to a connection issue (error code `Code::Unavailable` when connecting to the scheduler), wait 3 seconds and retry
  - if `maybe_upgrade` fails due to any other reason, exit with an error
  - if it succeeds (e.g. returns `Ok`), then this loop exits and the signer loop (b) starts

The signer loop (b) is already listening and reacting to the shutdown signal.

However, the upgrade loop (a) which is just above it, detailed above, doesn't. This means that, in a situation where the scheduler is not reachable, the upgrade loop (a) would run forever, even if the shutdown signal is received.

This PR wraps  `scheduler.maybe_upgrade` in a `tokio::select` that also listens to the shutdown signal. If received, the entire signer loop is exited.

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -187,33 +187,38 @@ impl Greenlight {
         info!("Entering the upgrade loop");
         loop {
             #[allow(deprecated)]
-            let maybe_upgrade_res = scheduler
-                .maybe_upgrade(UpgradeRequest {
-                    initmsg: self.signer.get_init(),
-                    signer_version: self.signer.version().to_owned(),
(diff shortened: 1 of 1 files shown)


ok300:
Superseded by #936

PR #936, opened 2024-04-16
Consider shutdown signal during entire signer loop

ok300:
Before this PR, the signer loop consisted of
- signer upgrade loop
  - which disregards the shutdown signal
- signer loop
  - which reacts to shutdown signal, but only when `signer.run_once` is not running, which may take 10 minutes to return. In other words, in the worst case the shutdown signal is handled (e.g. signer loop exits) 10 minutes after it's sent.

This PR wraps both loops above in a single `run_forever_inner` thread. In parallel to it, it listens to the shutdown signal and exits the thread as soon as the signal is received.

Possibly overrides #934.

---

**Note to reviewers / Open question**

I'm not sure if the signer upgrade loop (`scheduler.maybe_upgrade`) should be interrupted in the case of an upgrade. WIth this PR, it would be interrupted if we get a shutdown signal.

On the other hand, as described in #934, if we disregard the shutdown signal while we're in the signer upgrade loop, we may end up in an endless loop in the case when the GL scheduler is not reachable.

What do you think? How to best handle the upgrade loop?

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -172,10 +172,8 @@ impl Greenlight {
         })
     }
 
-    /// The actual signer loop. Connects to, upgrades and keeps alive the connection to the signer.
-    ///
-    /// Used as inner loop for `run_forever`.
-    async fn run_forever_inner(&self) -> Result<(), anyhow::Error> {
+    /// Create and, if necessary, upgrade the scheduler
+    async fn init_scheduler(&self) -> Result<SchedulerClient<tonic::transport::channel::Channel>> {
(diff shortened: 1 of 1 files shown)


roeierez (approved):
LGTM and I tested it successfully!
We should follow up with a PR to Greenlight.

dangeross (approved):
LGTM. Solves the shutdown issue after interacting with the signer

ok300:
As discussed, I separated the scheduler loop from the signer loop, such that calls to `scheduler.maybe_upgrade` remain unaffected by the shutdown signal.

PR is ready for final review.

roeierez (approved):
LGTM

PR #941, opened 2024-04-19
Persist LNURL-pay comment sent to the recipient

andrei-21:
We want to persist and expose the comment the user sends to the recipient when using LNURL-pay. Do you like this approach?


dangeross:
I think this is a good way to do it.

I would note the PR still needs:
- the bindings UDL file updating so the field is exposed to the generated bindings
- a table migration adding to the database

dangeross (commented):
Looks good, but can you run the flutter and react native code generators on the PR also?

dangeross (approved):
LGTM

ok300 (approved):

PR #943, opened 2024-04-23
Yield call to send_payment after a timeout and return a pending payment

dangeross:
This adds the following to the SDK:
- [x] A **_PaymentStarted_** event to notify to beginning of a payment.
- [x] Run the `send_payment()` request in a thread with a timeout to prevent the method call hanging too long. If it times out, the pending **_Payment_** is returned and the client can then use the **_PaymentSucceed_** / **_PaymentFailed_** events to handle the completion of the payment.
- [x] Run the `lnurl_pay()` request (which uses `send_payment()`) without a timeout.

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -401,6 +401,7 @@ interface BreezEvent {
     NewBlock(u32 block);
     InvoicePaid(InvoicePaidDetails details);
     Synced();
+    PaymentStarted(Payment details);
     PaymentSucceed(Payment details);
     PaymentFailed(PaymentFailedData details);
     BackupStarted();
@@ -749,6 +750,7 @@ dictionary SendPaymentRequest {
     string bolt11;
     u64? amount_msat = null;
     string? label = null;
+    u64? pending_timeout_sec = null;
 };
 
 dictionary SendSpontaneousPaymentRequest {
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -328,32 +328,31 @@ impl BreezServices {
                 self.notify_event_listeners(BreezEvent::payment_started(pending_payment.clone()))
                     .await?;
 
-                let pending_timeout_sec = req.pending_timeout_sec.unwrap_or(u64::MAX);
-                let (tx, rx) = std::sync::mpsc::channel();
+                let (tx, rx) = tokio::sync::oneshot::channel();
                 let cloned = self.clone();
                 tokio::spawn(async move {
+                    let label = req.label;
+                    let payee_pk = parsed_invoice.payee_pubkey.clone();
+                    let bolt11 = parsed_invoice.bolt11.clone();
+
                     let payment_res = cloned
                         .node_api
-                        .send_payment(
-                            parsed_invoice.bolt11.clone(),
-                            req.amount_msat,
-                            req.label.clone(),
-                        )
+                        .send_payment(bolt11, req.amount_msat, label.clone())
                         .map_err(Into::into)
                         .await;
                     let result = cloned
-                        .on_payment_completed(
-                            parsed_invoice.payee_pubkey.clone(),
-                            Some(parsed_invoice),
-                            req.label,
-                            payment_res,
-                        )
+                        .on_payment_completed(payee_pk, Some(parsed_invoice), label, payment_res)
                         .await;
                     let _ = tx.send(result);
                 });
 
-                match rx.recv_timeout(Duration::from_secs(pending_timeout_sec)) {
-                    Ok(result) => result.map(SendPaymentResponse::from_payment),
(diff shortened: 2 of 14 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -313,20 +317,41 @@ impl BreezServices {
...
-                    .await?;
-                Ok(SendPaymentResponse { payment })
+                let pending_payment =
+                    self.persist_pending_payment(&parsed_invoice, amount_msat, req.label.clone())?;
+                self.notify_event_listeners(BreezEvent::PaymentStarted {
+                    details: pending_payment.clone(),
+                })
+                .await?;
+
+                let timeout = Duration::from_secs(self.config.payment_timeout_sec as u64);
```
dangeross, line 327: TODO: Change the configured timeout param

libs/sdk-core/src/models.rs
```diff
@@ -871,14 +872,17 @@ pub struct ReceivePaymentResponse {
...
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct SendPaymentRequest {
     /// The bolt11 invoice
     pub bolt11: String,
     /// The amount to pay in millisatoshis. Should only be set when `bolt11` is a zero-amount invoice.
     pub amount_msat: Option<u64>,
     /// The external label or identifier of the [Payment]
     pub label: Option<String>,
+    /// If set, a timeout in seconds, that [crate::BreezServices::send_payment] will return a
+    /// pending payment if not already finished.
```
ok300, line 884:
```suggestion
    /// If set, a timeout in seconds, after which [crate::BreezServices::send_payment] will return
    /// the pending payment if not already finished.
    /// 
    /// If not set, [crate::BreezServices::send_payment] will return when payment succeeds.
```

libs/sdk-core/src/breez_services.rs
```diff
@@ -313,20 +317,63 @@ impl BreezServices {
...
+                    details: pending_payment.clone(),
+                })
+                .await?;
+
+                match req.pending_timeout_sec {
+                    Some(secs) => {
+                        let (tx, rx) = std::sync::mpsc::channel();
+                        let cloned = self.clone();
+                        tokio::spawn(async move {
+                            let payment_res = cloned
```
ok300, line 332: Would it work to call `node_api.send_payment` just before this `match`, and then only use the `match` to control waiting for `on_payment_completed`?

That way we could get rid of some code duplication.

dangeross: I tried various ways to achieve this, in the end doing so adds complications sharing the future across threads. This seems the simplest solution to me

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -749,6 +750,7 @@ dictionary SendPaymentRequest {
     string bolt11;
     u64? amount_msat = null;
     string? label = null;
+    u64? pending_timeout_sec = null;
```
roeierez, line 753: Do we need this? Why not using the payment_timeout from the config?

libs/sdk-core/src/breez_services.rs
```diff
@@ -314,25 +323,39 @@ impl BreezServices {
...
+                let pending_payment =
+                    self.persist_pending_payment(&parsed_invoice, amount_msat, req.label.clone())?;
+                self.notify_event_listeners(BreezEvent::payment_started(pending_payment.clone()))
                     .await?;
-                Ok(SendPaymentResponse { payment })
+
+                let pending_timeout_sec = req.pending_timeout_sec.unwrap_or(u64::MAX);
+                let (tx, rx) = std::sync::mpsc::channel();
+                let cloned = self.clone();
+                tokio::spawn(async move {
```
roeierez, line 334: Instead of using an async task and a channel perhaps we can simplify by using tokio::time::timeout ?
for example: 

`let result = tokio::time::timeout(Duration::from_seconds(pending_timeout), node_api.send_payment... )`

roeierez (changes requested):
A small simplification suggested.

libs/sdk-core/src/breez_services.rs
dangeross: I've set this PR to draft while we work on other things, @ok300 has a proposal to improve it

But regarding the tokio timeout, isn't the future cancelled if the timeout occurs? In our case the future has to complete with or without timeout

roeierez: I see, good point as we don't want to show error while still pending.

dangeross: Also we want the sync to complete in the background if pending is already returned and the Payment(Succeed/Failed) to be emitted

ok300: I added #958 to fix some issues around that and improve handling.

It's using a combination of both your ideas:
- the channel (`tx`, `rx`)
- `tokio::time::timeout(Duration::from_secs(pending_timeout_sec), rx).await`

ok300:
The failing tests would be fixed by #958

PR #944, opened 2024-04-23
Jitpack build: Remove unnecessary --insecure curl flag

ok300:
The certificate issues are fixed on our mvn server and `curl` works without the `--insecure` flag:

```sh
curl https://mvn.breez.technology/releases/breez_sdk/bindings-android/0.4.0-rc5/bindings-android-0.4.0-rc5.aar --output bindings-android-0.4.0-rc5.aar
```


dangeross (approved):
LGTM. This script is run on the jitpack server when building, so the only way fully test is to merge and build.

ok300:
I ran the build after merging, looks good:

https://github.com/breez/breez-sdk/actions/workflows/build-bindings-android.yml

PR #946, opened 2024-04-24
Merge RN generation to the SDK

dangeross:
This PR moves generator code from https://github.com/breez/breez-sdk-rn-generator into the SDK as a React Native binding. This helps keep generated code in line with SDK versions if there are generator changes or hot fixes that need to keep old generated interface.


libs/sdk-bindings/bindings-react-native/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "bindings-react-native"
+version = "0.0.10"
```
ok300, line 3: Isn't the correct version `0.0.13`?

https://github.com/breez/breez-sdk-rn-generator/blob/59b4ba2a92bca5c2cb83f42d44f8a28583bd6566/Cargo.toml#L3C12-L3C18

dangeross: True, fixed

ok300 (approved):

PR #949, opened 2024-04-26
Add listchannels to execute_dev_command

JssDWt:
The existing implementation of execute_dev_command did not allow for parameterized commands. The following things were changed:
- the execute_dev_command sdk function can now take parameterized commands like `"listchannels --source xxx"`
- The sdk-cli now handles parameterized commands properly as if it's a commandline. 
- the `listchannels` function was added, with optional parameters `source`, `destination`, and `short_channel_id`, so we are able to query the graph on greenlight

### DIFF
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -113,6 +113,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d96bd03f33fe50a863e394ee9718a706f988b9079b20c3784fb726e7678b62fb"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8901269c6307e8d93993578286ac0edf7f195079ffff5ebdeea6a59ffb7e36bc"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c75ac65da39e5fe5ab759307499ddad880d724eed2f6ce5b5e8a26f4f387928c"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


ok300:
Isn't it overkill to include `clap`, a CLI parsing dependency, in the SDK?

If the goal is to extend dev commands with flexible calls that support optional parameters, then I suppose there's no way around it.

But if we instead want to add selected command variants that are especially useful, maybe we can instead go for e.g.
- `listChannelsForSource <source>` with one expected arg
- `listChannelsForSourceAndDest <source> <dest>` with two expected args

What are your thoughts on this?

ok300:
cc @erdemyerebasmaz : this is also relevant for Breez Cloud.

JssDWt:
> Isn't it overkill to include `clap`, a CLI parsing dependency, in the SDK?

My reasoning is: either we write parsing logic for parameterized commands ourselves, or we add a dependency to do so. The advantage is it maps to the enum fields easily. If we would do `listChannelsForSource <source>`, we would do something like:
- split string
- get first element, if there is no first element, error
- parse first element to enum variant
- pass the second element somewhere outside of this flow, so it can be used in the listchannels variant

I thought a cli parsing library was easier and more robust, also considering potential quoted strings with spaces (even though they are not part of the interface currently)

dangeross:
> Isn't it overkill to include `clap`, a CLI parsing dependency, in the SDK?

The only other option I would see is to implement a nested subcommand into the CLI to handle the params specific to `listchannels`, then instead of passing a command string to `execute_dev_command`, pass an enum with a variant to include the optional params

JssDWt:
> The only other option I would see is to implement a nested subcommand into the CLI to handle the params specific to `listchannels`, then instead of passing a command string to `execute_dev_command`, pass an enum with a variant to include the optional params

That would work too. It would be a breaking interface change, but maybe favorable over magic commandline strings?

JssDWt:
> The only other option I would see is to implement a nested subcommand into the CLI to handle the params specific to `listchannels`, then instead of passing a command string to `execute_dev_command`, pass an enum with a variant to include the optional params

@roeierez What do you think?

JssDWt:
> What do you think about exposing the `node::ClnClient` from the NodeAPI and the BreezServices?

It does tie us to the core lightning implementation. Thinking about the 'LDK support' in the roadmap. Also I wonder how that plays with language bindings. I don't personally mind the maintenance burden of supporting this enum.

roeierez:
> > What do you think about exposing the `node::ClnClient` from the NodeAPI and the BreezServices?
> 
> It does tie us to the core lightning implementation. Thinking about the 'LDK support' in the roadmap. Also I wonder how that plays with language bindings. I don't personally mind the maintenance burden of supporting this enum.

We can do that without exposing the implementation by returning `Any` type as the underline type and let the user downcast it but the binding is indeed won't work well with that.
I am good with the enum.

PR #953, opened 2024-04-26
Add Reverse Swap notification

### LINKED ISSUE #924: Add notification for reverse swap chain confirmation
The same we do for regular swaps can do also for reverse swap.
As part of pay_onchain we can register for notification about the address of the lockup transaction output and once the notification arrives add a job to process the reverse swap until completion.

dangeross:
This PR adds background processing of the reverse swap flow using webhooks:

- [x] Register a reverse swap lockup address for notifications
- [x] Add `process_reverse_swap` method to handle processing an individual reverse swap by address
- [x] Adapt notification plugin handling for the push notification
- [x] Add ReverseSwapUpdated event

Fixes #924

### DIFF
--- a/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/NotificationHelper.kt
+++ b/libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/NotificationHelper.kt
@@ -17,6 +17,15 @@ import androidx.annotation.RequiresApi
 import androidx.core.app.ActivityCompat
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
+import breez_sdk_notification.Constants.ADDRESS_TXS_CONFIRMED_NOTIFICATION_CHANNEL_DESCRIPTION
+import breez_sdk_notification.Constants.ADDRESS_TXS_CONFIRMED_NOTIFICATION_CHANNEL_NAME
+import breez_sdk_notification.Constants.ADDRESS_TXS_CONFIRMED_WORKGROUP_DESCRIPTION
+import breez_sdk_notification.Constants.ADDRESS_TXS_CONFIRMED_WORKGROUP_ID
+import breez_sdk_notification.Constants.ADDRESS_TXS_CONFIRMED_WORKGROUP_NAME
+import breez_sdk_notification.Constants.DEFAULT_ADDRESS_TXS_CONFIRMED_NOTIFICATION_CHANNEL_DESCRIPTION
+import breez_sdk_notification.Constants.DEFAULT_ADDRESS_TXS_CONFIRMED_NOTIFICATION_CHANNEL_NAME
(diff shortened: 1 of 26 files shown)


libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/ConfirmTransaction.kt
```diff
@@ -22,38 +23,63 @@ data class AddressTxsConfirmedRequest(
...
             this.bitcoinAddress = request.address
-            breezSDK.redeemSwap(request.address)
-            logger.log(TAG, "Found swap for ${request.address}", "INFO")
         } catch (e: Exception) {
-            logger.log(TAG, "Failed to manually redeem swap notification: ${e.message}", "WARN")
-            notifyFailure()
+            logger.log(TAG, "Failed to decode payload: ${e.message}", "WARN")
         }
+
+        this.bitcoinAddress?.let {address ->
```
roeierez, line 45: I think it would be nice to be able to recognize whether it is a swap or reverse swap without trying both.
If we pass a query param app_data in the webhook url it returns back in the notification payload.
See here where it is grapped from the query param: https://github.com/breez/notify/blob/main/http/router.go#L40
And here how it is populated into the payload: https://github.com/breez/notify/blob/main/breezsdk/init.go#L43
Perhaps we can use it here to identify if it is a swap or reverse swap?

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -940,6 +941,9 @@ interface BlockingBreezServices {
    [Throws=SdkError]
    sequence<ReverseSwapInfo> in_progress_reverse_swaps();
    
+   [Throws=SdkError]
+   void process_reverse_swap(string lockup_address);
```
roeierez, line 945: NIT: Perhaps `claim_reverse_swap` ?

roeierez (changes requested):
Looks good! dropped two comments.

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/job/ConfirmTransaction.kt
roeierez: That's a good argument. I am ok then to leave this inexpensive check.

libs/sdk-bindings/bindings-android/lib/src/main/kotlin/breez_sdk_notification/NotificationHelper.kt
```diff
@@ -194,20 +194,20 @@ class NotificationHelper {
                 group = LNURL_PAY_WORKGROUP_ID
             }
             val swapTxConfirmedNotificationChannel = NotificationChannel(
```
erdemyerebasmaz, line 196: Should we rename the notification channel variable as well?

erdemyerebasmaz (approved):
LGTM

Have one naming suggestion.

roeierez (approved):
LGTM

PR #954, opened 2024-04-29
Simplify payment yield flow

ok300:
This is an attempt to reduce code duplication and improve readability of the payment yield flow.


ok300:
I just realised (see previous commit) that the yield timeout is essentially `req.pending_timeout_sec` or infinity.

The upper bound is practically the node API timeout, but collapsing the timeout handling branches into one IMO achieves the same purpose, but in a simpler way.

JssDWt (approved):
Looks good!

dangeross (approved):

PR #957, opened 2024-04-30
Only consider pending_timeout_sec if it is set

ok300:
This PR uses `tokio::select` with a conditional `tokio::sleep`, which allows us to conditionally sleep.

If `pending_timeout_sec` is set, the first of `send_payment`, `sleep` will return. So if the timeout occurs first, the method will return.

If `pending_timeout_sec` is not set, it will only execute `send_payment`, because the second branch of `tokio::select` is ignored.


dangeross:
I think this can be closed because `tokio::select` doesn't fulfil the intension of completing the payment branch timeout or not.

ok300:
Correct. I forgot to close this as I was iterating through fixes.

Superseded by #958 .

Issue #966, opened 2024-05-08
Add keep-alive pings during payment


roeierez:
Currently users get transport errors when payments takes more than a certain time (probably around 30 seconds).
It seems that it relates to the fact that the underline gRPC connection keep alive doesn't work as expected and that is under work in greenlight.
Meantime we can mitigate that by sending Ping requests over greenlight gRPC interface every let's say 15 seconds to keep the node connection alive.

PR #968, opened 2024-05-08
Keep node connection alive during payments

roeierez:
Transport errors still happen ocasionaly and the reason is because the node connection is lost during a relatively long period that the grpc API is not used.
In theory the gRPC keep alive mechanism should take care of keeping the connection alive but from unknown reason atm it is not as expected.
This causes payments to fail since during payments we are not using the gRPC API (except for the pending send_payment request).
Following last discussion with greenlight team regarding the transport errors we have decided to implement a keep alive mechanism in the sdk until this is resolved in greenlight.

fixes https://github.com/breez/breez-sdk/issues/966

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -1266,30 +1266,37 @@ impl NodeAPI for Greenlight {
 
     async fn start_keep_alive(&self, mut shutdown: watch::Receiver<()>) {
         info!("keep alive started");
-        let mut client = self.get_node_client().await.unwrap();
         let mut interval = tokio::time::interval(Duration::from_secs(15));
         interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
         loop {
             tokio::select! {
-              _ = shutdown.changed() => {
-                info!("keep alive exited");
-                break;
-              }
-              _ = interval.tick() => {
-                let inprogress_payments = self.inprogress_payments.load(Ordering::Relaxed);
-                if inprogress_payments == 0 {
-                  continue
-                }
-                let res = client.getinfo(cln::GetinfoRequest {}).await;
-                match res {
-                  Ok(_) => {
-                    info!("keep alive ping sent, in progress payments: {inprogress_payments}");
+                  _ = shutdown.changed() => {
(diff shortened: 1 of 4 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1245,6 +1264,36 @@ impl NodeAPI for Greenlight {
...
+        loop {
+            tokio::select! {
+              _ = shutdown.changed() => {
+                info!("keep alive exited");
+                break;
+              }
+              _ = interval.tick() => {
+                let inprogress_payments = self.inprogress_payments.load(Ordering::Relaxed);
+                if inprogress_payments == 0 {
+                  continue
```
ok300, line 1281: No keep_alive if there are no pending payments?

What if the SDK has no activity for a long time, then the connection is lost, then a payment is attempted?

roeierez: When the payment is attempted (or Any other API) the connection is established, so no issues here.
The only problem is when there is a pending request and then not API call is executed for some time which causes the connection to be killed in the middle of the request

ok300 (approved):

libs/sdk-core/src/greenlight/node_api.rs
ok300: Then makes sense.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1245,6 +1264,36 @@ impl NodeAPI for Greenlight {
         }
     }
 
+    async fn start_keep_alive(&self, mut shutdown: watch::Receiver<()>) {
+        info!("keep alive started");
+        let mut client = self.get_node_client().await.unwrap();
```
dangeross, line 1269: Is there a safer way to handle this unwrap to prevent a panic?

roeierez: Of course, good catch. Was just a left over from my test. will fix now.

roeierez: Done

dangeross (approved):
LGTM

Issue #969, opened 2024-05-08
vls-protocol-signer-0.11.0 compilation error


darioAnongba:
While updating the dependencies of our project with `cargo update`, I ran into the following compiling error linked to VLS. This is on MacOS M2 using the Breez SDK v0.4.0 (tried with 0.4.0-rc1 as well).
I'm not sure what the issue could be here:

```
error[E0507]: cannot move out of dereference of `Octets`
   --> /.../.cargo/registry/src/index.crates.io-6f17d22bba15001f/vls-protocol-signer-0.11.0/src/handler.rs:690:36
    |
690 |                   let data: Vec<_> = m
    |  ____________________________________^
691 | |                     .u5bytes
692 | |                     .clone()
    | |____________________________^ move occurs because value has type `std::vec::Vec<u8>`, which does not implement the `Copy` trait
693 |                       .into_iter()
    |                        ----------- value moved due to this method call
    |
note: `into_iter` takes ownership of the receiver `self`, which moves value
   --> ....rustup/toolchains/1.76-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/traits/collect.rs:268:18
    |
268 |     fn into_iter(self) -> Self::IntoIter;
    |                  ^^^^
help: you can `clone` the value and consume it, but this might not be your desired behavior
    |
690 ~                 let data: Vec<_> = <std::vec::Vec<u8> as Clone>::clone(&m
691 |                     .u5bytes
692 ~                     .clone())    |
```

Using `rustc 1.76.0`. Pretty surprised no-one is running into this issue so I'm wondering if I messed up something. Sadly for me, I didn't commit the Cargo.lock to revert so now I'm stuck :/.

I tried `cargo clean` without success.

JssDWt:
I ran into this same error when trying to `cargo install` gl-plugin in a debian bookworm docker container on an M3 mac. `cargo build --release` does work though. It's an error compiling the vls-protocol-signer. I haven't been able to figure out exactly why the build fails. I think it has something to do with the `bitcoinconsensus` dev dependency being compiled.

darioAnongba:
any idea how to bypass this? Even a temporary fix ?

darioAnongba:
I opened an issue directly at Greenlight here: https://github.com/Blockstream/greenlight/issues/434 but this issue makes the Rust Breez SDK unusable at the moment as it does not compile.

[...]

Issue #584, opened 2023-11-06
Could not compile "breez-sdk-core"


darioAnongba:
Hi, when trying to use the SDK in Rust, it fails to compile whether I try to use the `main` branch or the `0.2.7` rev as such. Probably I'm doing something wrong.

```rust
[dependencies]
...
breez-sdk-core = { git = "https://github.com/breez/breez-sdk", branch = "main" } // or
breez-sdk-core = { git = "https://github.com/breez/breez-sdk", rev = "0.2.7" }
```

## Error when using `branch ="main"`

```
error[E0432]: unresolved imports `ecies::utils::aes_decrypt`, `ecies::utils::aes_encrypt`
 --> /Users/darioanongbavarela/.cargo/git/checkouts/breez-sdk-d2a5b0f6a3ca00f6/f5b24c7/libs/sdk-core/src/backup.rs:9:20
  |
9 | use ecies::utils::{aes_decrypt, aes_encrypt};
  |                    ^^^^^^^^^^^  ^^^^^^^^^^^ no `aes_encrypt` in `utils`
  |                    |
  |                    no `aes_decrypt` in `utils`

error[E0432]: unresolved imports `ecies::utils::aes_decrypt`, `ecies::utils::aes_encrypt`
  --> /Users/darioanongbavarela/.cargo/git/checkouts/breez-sdk-d2a5b0f6a3ca00f6/f5b24c7/libs/sdk-core/src/greenlight/node_api.rs:16:20
   |
16 | use ecies::utils::{aes_decrypt, aes_encrypt};
   |                    ^^^^^^^^^^^  ^^^^^^^^^^^ no `aes_encrypt` in `utils`
   |                    |
   |                    no `aes_decrypt` in `utils`

error[E0277]: the trait bound `SecpError: std::error::Error` is not satisfied
 --> /Users/darioanongbavarela/.cargo/git/checkouts/breez-sdk-d2a5b0f6a3ca00f6/f5b24c7/libs/sdk-core/src/crypt.rs:6:29
  |
6 |         Err(err) => Err(err.into()),
  |                             ^^^^ the trait `std::error::Error` is not implemented for `SecpError`
  |
  = note: required for `anyhow::Error` to implement `From<SecpError>`
  = note: required for `SecpError` to implement `Into<anyhow::Error>`

error[E0277]: the trait bound `SecpError: std::error::Error` is not satisfied
[...]

PR #971, opened 2024-05-10
CI: Add step to test if SDK compiles with latest dependencies

ok300:
This PR adds a new CI step that creates a new plain Rust project, adds the SDK as single dependency and tries to compile it. This tests whether the SDK compiles with the latest version of the dependencies that can be updated.

TLDR: When this CI step fails, it's an early warning signal that allows us to become aware of and solve issues such as

- #584
- #969
- #1010 

before customers report them.

---

Our checked-in Cargo.lock contains the specific combination of all direct and transitive dependency versions. This dependency tree snapshot was tested against during development and is what we release.

However, when integrating the SDK in a new Rust project, Cargo not use our Cargo.lock. Instead, it will try to generate a new Cargo.lock based on our (and our dependencies') Cargo.toml files. This means that, where a dependency version range is used in a Cargo.toml, Cargo will try to upgrade it to the latest matching version. If this happens, this new dependency version may even result in the whole project failing to compile. In that case, the only solution is to manually pin the problematic dependencies to the last known good versions, in the application's Cargo.toml.

Since this is the situation new projects are faced with when adding the SDK as a Rust dependency, we simulate it here to get an early warning signal, should any newer dependency cause it to fail to compile.

See discussion at https://github.com/breez/breez-sdk/issues/969#issuecomment-2104700522

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -273,3 +273,56 @@ jobs:
 
       - name: dart-format
         run: dart format -o none --set-exit-if-changed -l 110 .
+
+  # Create a new plain Rust project, add the SDK as single dependency and try to compile it.
+  # This tests whether the SDK compiles with the latest version of the dependencies that can be updated.
+  #
+  # Our checked-in Cargo.lock contains the specific combination of all direct and transitive dependency versions.
+  # This dependency tree snapshot was tested against during development and is what we release.
+  #
+  # However, when integrating the SDK in a new Rust project, Cargo not use our Cargo.lock. Instead, it will try to generate
+  # a new Cargo.lock based on our (and our dependencies') Cargo.toml files. This means that, where a dependency version range
+  # is used in a Cargo.toml, Cargo will try to upgrade it to the latest matching version. If this happens, this new dependency
+  # version may even result in the whole project failing to compile. In that case, the only solution is to manually pin
+  # the problematic dependencies to the last known good versions, in the application's Cargo.toml.
+  #
+  # Since this is the situation new projects are faced with when adding the SDK as a Rust dependency, we simulate it here
+  # to get an early warning signal, should any newer dependency cause it to fail to compile.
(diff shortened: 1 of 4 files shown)


dangeross (approved):
Looks good. Seem to be catching the issue from yesterday

ok300:
Rebased on #1014 to check if it sees that fix as valid.

.github/workflows/main.yml
```diff
@@ -273,3 +273,56 @@ jobs:
...
+  # a new Cargo.lock based on our (and our dependencies') Cargo.toml files. This means that, where a dependency version range
+  # is used in a Cargo.toml, Cargo will try to upgrade it to the latest matching version. If this happens, this new dependency
+  # version may even result in the whole project failing to compile. In that case, the only solution is to manually pin
+  # the problematic dependencies to the last known good versions, in the application's Cargo.toml.
+  #
+  # Since this is the situation new projects are faced with when adding the SDK as a Rust dependency, we simulate it here
+  # to get an early warning signal, should any newer dependency cause it to fail to compile.
+  #
+  # See discussion at https://github.com/breez/breez-sdk/issues/969#issuecomment-2104700522
+  check-sdk-as-dependency:
```
roeierez, line 293: I think we can achieve that by adding `cargo install --path .` to the sdk-cli or I am missing something?

roeierez (commented):
LGTM although I think we could

.github/workflows/main.yml
ok300: Unfortunately not, because that would use the checked-in `Cargo.lock` of `sdk-cli`, so it wouldn't simulate the situation of "fresh new app with Breez SDK dependency".

AFAIK for applications, Rust relies on the available `Cargo.lock`. For libraries (like Breez SDK), it ignores it and reconstructs a fresh one from the library's `Cargo.toml`.

PR #972, opened 2024-05-13
Update Android build clang to 17

dangeross:
This PR updates the clang version used to build android libraries to 17

### DIFF
--- a/libs/sdk-bindings/build.rs
+++ b/libs/sdk-bindings/build.rs
@@ -2,7 +2,7 @@ use anyhow::*;
 use std::env;
 use std::path::Path;
 
-const DEFAULT_CLANG_VERSION: &str = "14.0.7";
+const DEFAULT_CLANG_VERSION: &str = "17";
 
 /// Adds a temporary workaround for an issue with the Rust compiler and Android
 /// in x86_64 devices: https://github.com/rust-lang/rust/issues/109717.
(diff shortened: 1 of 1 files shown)


roeierez (approved):
LGTM

JssDWt:
I don't have clang 17 on my machine, what do you have to do to install that clang version? Is this all simply tied to the NDK version? Should we force a fixed NDK version? Or do some kind of match statement based on the NDK version?

dangeross:
@JssDWt the current default NDK version on macos-latest is 26.3.11579264, with the included version of clang being 17. I think it recently changed from 25.2.9519653 being the default.

https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#android

I'm not sure what the best approach is, but it needs to support building locally and in the CI

JssDWt:
> I'm not sure what the best approach is, but it needs to support building locally and in the CI

I think it should be fine this way. If someone has a different clang version they can set the `NDK_CLANG_VERSION` environment variable during the build. Maybe let's change the panic error message on line 33 to `Path {linkpath} not exists. Try setting the NDK_CLANG_VERSION environment variable.`

dangeross:
> Maybe let's change the panic error message on line 33 to `Path {linkpath} not exists. Try setting the NDK_CLANG_VERSION environment variable.`

@JssDWt done

JssDWt (approved):

PR #975, opened 2024-05-15
Pattern match the linux x86_64 clang directory

dangeross:
This PR uses glob in the build to pattern match the linux x86_64 clang directory

### DIFF
--- a/libs/sdk-bindings/build.rs
+++ b/libs/sdk-bindings/build.rs
@@ -6,7 +6,7 @@ use std::result::Result::Ok;
 /// Adds a temporary workaround for an issue with the Rust compiler and Android
 /// in x86_64 devices: https://github.com/rust-lang/rust/issues/109717.
 /// The workaround comes from: https://github.com/smartvaults/smartvaults/blob/827805a989561b78c0ea5b41f2c1c9e9e59545e0/bindings/smartvaults-sdk-ffi/build.rs
-fn setup_x86_64_android_workaround() -> Result<()> {
+fn setup_x86_64_android_workaround() {
     let target_os = env::var("CARGO_CFG_TARGET_OS").expect("CARGO_CFG_TARGET_OS not set");
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").expect("CARGO_CFG_TARGET_ARCH not set");
     if target_arch == "x86_64" && target_os == "android" {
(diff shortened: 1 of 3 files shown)


libs/sdk-bindings/build.rs
```diff
@@ -20,23 +19,22 @@ fn setup_x86_64_android_workaround() {
...
+        let linux_x86_64_lib_pattern = format!(
+            "{android_ndk_home}/toolchains/llvm/prebuilt/{build_os}-x86_64/lib*/clang/**/lib/linux/"
         );
-        let linkpath = format!("{android_ndk_home}/{linux_x86_64_lib_dir}");
-        if Path::new(&linkpath).exists() {
-            println!("cargo:rustc-link-search={android_ndk_home}/{linux_x86_64_lib_dir}");
-            println!("cargo:rustc-link-lib=static=clang_rt.builtins-x86_64-android");
-        } else {
-            panic!("Path {linkpath} not exists. Try setting the NDK_CLANG_VERSION environment variable.");
+        match glob(&linux_x86_64_lib_pattern)?.last() {
```
roeierez, line 25: NIT: Since we already panic at this function perhaps using unwrap here instead of changing the method prototype?

roeierez (commented):
LGTM - seems like a robust way to fix different paths in different NDKs.
Small NIT.

libs/sdk-bindings/build.rs
dangeross: Done

roeierez (approved):
LGTM

libs/sdk-bindings/build.rs
```diff
@@ -20,17 +19,15 @@ fn setup_x86_64_android_workaround() {
...
+        let linux_x86_64_lib_pattern = format!(
+            "{android_ndk_home}/toolchains/llvm/prebuilt/{build_os}-x86_64/lib*/clang/**/lib/linux/"
         );
-        let linkpath = format!("{android_ndk_home}/{linux_x86_64_lib_dir}");
-        if Path::new(&linkpath).exists() {
-            println!("cargo:rustc-link-search={android_ndk_home}/{linux_x86_64_lib_dir}");
-            println!("cargo:rustc-link-lib=static=clang_rt.builtins-x86_64-android");
-        } else {
-            panic!("Path {linkpath} not exists. Try setting the NDK_CLANG_VERSION environment variable.");
+        match glob(&linux_x86_64_lib_pattern).unwrap().last() {
```
JssDWt, line 25: Looks good. I have a nit that may not be useful. This might not always take the latest clang version if multiple are available. (Is that even ever the case?). (Do we even need the latest version?). Anyway, v9.0.0 will be the 'last' if v10.0.0 is also available, because the results from the glob are sorted alphabetically.

ok300: The clang version is in the double-digits, so that won't be a problem for a while.

JssDWt: > The clang version is in the double-digits, so that won't be a problem for a while.

Yeah true only minor and patch versions maybe.

JssDWt (approved):

PR #978, opened 2024-05-18
[frb] Allow SDK streams to be re-initialised between sessions

erdemyerebasmaz:
This PR adds the ability to re-initialise SDK streams between sessions without the need for Stream & StreamSubscription management on Flutter side.

- frb v2 is required to be able to wait for Rust function to finish execution before the stream can be obtained in Dart. 
  - [Control whether to wait](https://cjycode.com/flutter_rust_bridge/guides/types/translatable/stream#control-whether-to-wait) 
 - This feature prevents main thread to lock whenever a stream event needs to be emitted from Rust side and allows stream functions to be asynchronous, which enables us to set or update streams during initialisation.

There's no demand for it but this opens the ability to switch between event handler logic from UI during an ongoing session.

[In draft]


erdemyerebasmaz:
Close in favor of #980

PR #980, opened 2024-05-19
[Flutter] Improve SDK stream management

erdemyerebasmaz:
This PR addresses 
- #979

SDK streams references are created on `initialize` which can be listened, subscribed to & unsubscribed from during BreezSDK's lifetime.

#### Changelist:
- Subscribe and unsubscribe to SDK streams on each connect/disconnect 90bb47e

#### Other changelist:
- Move documentation of streams to their correct places dd1716a
- Change serviceHealthCheck to a block function for consistency af31210

### DIFF
--- a/libs/sdk-flutter/lib/breez_sdk.dart
+++ b/libs/sdk-flutter/lib/breez_sdk.dart
@@ -14,14 +14,14 @@ class BreezSDK {
 
   /* Streams */
 
-  /// Listen to paid Invoice events
   final StreamController<InvoicePaidDetails> _invoicePaidStream = StreamController.broadcast();
 
+  /// Listen to paid Invoice events
   Stream<InvoicePaidDetails> get invoicePaidStream => _invoicePaidStream.stream;
 
-  /// Listen to payment results
   final StreamController<Payment> _paymentResultStream = StreamController.broadcast();
 
+  /// Listen to payment results
   Stream<Payment> get paymentResultStream => _paymentResultStream.stream;
 
   /* SDK Streams */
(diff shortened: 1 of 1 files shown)


libs/sdk-flutter/lib/breez_sdk.dart
```diff
@@ -113,9 +87,8 @@ class BreezSDK {
   Future connect({
     required ConnectRequest req,
   }) async {
-    await _lnToolkit.connect(
-      req: req,
-    );
+    await _lnToolkit.connect(req: req);
+    _subscribeToSdkStreams();
```
roeierez, line 91: I don't think this will work atm. The rust side only accept one registration in a lifetime. The second time connect will be called it will fail.

roeierez (changes requested):

libs/sdk-flutter/lib/breez_sdk.dart
erdemyerebasmaz: We have a single registration on initialize, which passes a copy of the value that references the SDK's stream to instance's Stream:
```
 // Flutter Instance Stream
Stream<BreezEvent>? _breezEventsStream;

void _initializeEventsStream() {
    // Stream from Rust
    _breezEventsStream = _lnToolkit.breezEventsStream(); 
}
```

which we can subscribe & unsubscribe from via `StreamSubscription<BreezEvent>? _breezEventsSubscription;`

I will test it today.

erdemyerebasmaz: @roeierez Works as expected.

Made a required change to be able to listen the streams multiple times: https://github.com/breez/breez-sdk/pull/980/commits/6f52b18e3474e5cd5787feaf5b03b3ac9b5a2af1

libs/sdk-flutter/lib/breez_sdk.dart
```diff
@@ -14,82 +14,56 @@ class BreezSDK {
...
-            onError: (e) => _logStreamController.addError(e),
-          );
+          .map((log) => LogEntry(line: log["line"], level: log["level"]));
     } else {
-      _lnToolkit.breezLogStream().listen((logEntry) {
-        _logStreamController.add(logEntry);
-      }, onError: (e) {
-        _logStreamController.addError(e);
-      });
+      _breezLogStream = _lnToolkit.breezLogStream().asBroadcastStream();
```
roeierez, line 53: @erdemyerebasmaz  I think I understand now what you are trying to do here. As far as I understand now this line actually calls the rust side logic in bindings.rs. Is that right?

erdemyerebasmaz: Yes, that's true and we ensure it's called once, create a broadcast stream reference of it that can be subscribed/unsubscribed from via their StreamSubscription's.

dangeross (approved):

PR #981, opened 2024-05-21
Document deprecated methods with `deprecated` macro

andrei-21:
Rust compiler will issue a warning in case of usage of such method what helps spot the change.


roeierez (approved):
LGTM

ok300 (approved):

PR #985, opened 2024-05-24
Move greenlight to new credential mechanism

JssDWt:
Greenlight has moved to a new credential mechanism. This PR migrates to that new mechanism.

- Old credentials retrieved from the scheduler are now deprecated. They are removed from the database in a migration. So the initial run with this new version will recover the node and replace the credentials with the new credentials.
- The `GreenlightCredentials` are only used for `recover` and `register`. In fact they are not device credentials. This PR renames the fields `device_key` and `device_cert` in these credentials to `developer_key` and `developer_cert`.
- Internally the scheduler and signer are initialized with `Device` credentials received from the scheduler. They now contain a rune and a ca certificate. The credentials are 'upgraded' on initial fetch. This may help with future releases of the sdk where the 'upgrade' mechanism may change.
- The `node_credentials` function now returns a serialized representation of the greenlight device credentials, rather than a device key and device cert.

Tested with
- [x] connect with non-upgraded node with existing database (drops credentials and calls recover)
- [x] connect with non-upgraded node without existing database (calls recover)
- [x] connect with upgraded node with existing database (calls connect)
- [x] connect with upgraded node without existing database (calls recover)
- [x] connect with new node with invite code
- [ ] connect with new node with partner credentials


JssDWt:
@roeierez @ok300 this PR is ready for review

ok300 (approved):

roeierez:
@JssDWt this can be rebased and merged. The upgrade issue is not specific to this version.

roeierez:
@JssDWt this can be rebased and merged.

PR #986, opened 2024-05-24
Parse web URLs with `lightning` query param

andrei-21:


ok300:
What is the use-case for such URLs?

When specified as a URL, we risk not getting part of it, because AFAIK different browsers and clients have a max length for an URL with GET params.

Isn't a more standard-conform way doing it with the `lightning:` URI prefix, which the SDK already handles?

roeierez:
@ok300  it is part of the standard to pass lnurl as part of other uri schemes: https://github.com/lnurl/luds/blob/luds/01.md

We do support it in breezmobile and Idont see why not to support it.

roeierez:
We don't control the way lnurl is encoded as some wallets use the fallback scheme like this.

ok300:
@andrei-21 can you please rebase and fix the conflicts?

We can then review and merge.

ok300 (approved):

PR #988, opened 2024-05-26
Dev commands on GL: return output as pretty-printed JSON

### LINKED ISSUE #939: JSON output for dev commands
Hey,

We would like to use the data from dev commands in the settings of our app to debug our users. 

The issue is that the current format is a stringified rust structure, which is unusable for us. We can try to convert it to JSON but if we can get it directly from the SDK it would be simpler.

Thanks,
Simon

ok300:
Depends on #985

Closes #939

cc @erdemyerebasmaz : I think it's relevant for C-Breez too


erdemyerebasmaz:
@ok300 no further changes are required on C-Breez.

ok300:
Superseded by #1031

PR #989, opened 2024-05-26
Chain service error

### LINKED ISSUE #983: Generic: All chain service instances failed
LOG  {"level": "TRACE", "line": "put; add idle connection for (\"https\", mempool.emzy.de)"}
 LOG  {"level": "DEBUG", "line": "pooling idle connection for (\"https\", mempool.emzy.de)"}
 LOG  {"level": "ERROR", "line": "Call to chain service https://mempool.emzy.de/api failed: error decoding response body: expected value at line 1 column 1"}
 LOG  {"level": "TRACE", "line": "client tx closed"}
 LOG  {"level": "TRACE", "line": "pool closed, canceling idle interval"}
 LOG  {"level": "TRACE", "line": "State::close_read()"}
 LOG  {"level": "TRACE", "line": "State::close_write()"}
 LOG  {"level": "TRACE", "line": "flushed({role=client}): State { reading: Closed, writing: Closed, keep_alive: Disabled }"}
 LOG  {"level": "TRACE", "line": "shut down IO complete"}
 LOG  Generic: All chain service instances failed

roeierez:
fixed #983 

This PR address two issues:
1. Ensures we only parse the json in case of success status code (2xx)
2. log any chain service errors with the corresponding url and associated error returned.
3. When attempting to lookup for closing tx onchain don't fail the sync process if the tx was not found (which technically could happen in case of zero conf).

### DIFF
--- a/libs/sdk-core/src/input_parser.rs
+++ b/libs/sdk-core/src/input_parser.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 
 use anyhow::{anyhow, Result};
 use bip21::Uri;
+use reqwest::StatusCode;
 use serde::Deserialize;
 use serde::Serialize;
 
@@ -214,7 +215,7 @@ pub async fn parse(input: &str) -> Result<InputType> {
         }
 
         lnurl_endpoint = maybe_replace_host_with_mockito_test_host(lnurl_endpoint)?;
-        let lnurl_data: LnUrlRequestData = get_parse_and_log_response(&lnurl_endpoint)
+        let lnurl_data: LnUrlRequestData = get_parse_and_log_response(&lnurl_endpoint, false)
             .await
             .map_err(|_| anyhow!("Failed to parse response"))?;
         let temp = lnurl_data.into();
@@ -258,7 +259,8 @@ pub(crate) async fn post_and_log_response(url: &str, body: Option<String>) -> Sd
 /// Makes a GET request to the specified `url` and logs on DEBUG:
 /// - the URL
 /// - the raw response body
-pub(crate) async fn get_and_log_response(url: &str) -> SdkResult<String> {
+/// - the response HTTP status code
+pub(crate) async fn get_and_log_response(url: &str) -> SdkResult<(String, StatusCode)> {
     debug!("Making GET request to: {url}");
 
     let response = get_reqwest_client()?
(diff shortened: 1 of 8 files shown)


libs/sdk-core/src/input_parser.rs
```diff
@@ -261,17 +261,26 @@ pub(crate) async fn post_and_log_response(url: &str, body: Option<String>) -> Sd
...
+            .text()
+            .await
+            .map_err(|e| SdkError::ServiceConnectivity { err: e.to_string() }),
+        false => {
+            let err = format!(
+                "GET request {} failed with status: {}",
+                url,
+                response.status()
+            );
+            error!("{err}");
```
dangeross, line 280: Is it worth logging the response body in unsuccessful requests or is the http code enough to diagnose potential issues?

roeierez: I think it worth since it shouldn't happen too much and when it happens like in the associated issue it helps identify the problem.

libs/sdk-core/src/input_parser.rs
```diff
@@ -261,17 +261,26 @@ pub(crate) async fn post_and_log_response(url: &str, body: Option<String>) -> Sd
...
         .map_err(|e| SdkError::ServiceConnectivity { err: e.to_string() })?;
-    debug!("Received raw response body: {raw_body}");
-
-    Ok(raw_body)
+    match response.status().is_success() {
+        true => response
+            .text()
+            .await
+            .map_err(|e| SdkError::ServiceConnectivity { err: e.to_string() }),
+        false => {
```
ok300, line 274: Unfortunately this would break error handling for LNURL.

Some specs (like LNURL but maybe others too) define errors in terms of how the response payload looks like, regardless of HTTP status.

If we now throw `Err` here in the `GET` response parsing in case of HTTP error code, this means we will not return anymore the LNURL error JSON payload

```json
{"status": "ERROR", "reason": "error details..."}
```

which means we lose the info on the error cause indicated by the server (not enough funds, etc).

IMO we need to return both the HTTP status and the raw payload, then let the caller decide. For mempool API calls, HTTP error code means error. For LNURL, the HTTP error code is only one mode of failure, the other is the error payload (can be returned with HTTP 200 or HTTP error).

ok300: I tested it and indeed:

with the PR, trying to LNURL-withdraw from a wallet with no funds returns

> Error: Service connectivity: Service connectivity: GET request https://stacker.news/api/lnwith?k1=ce529c69344c7ae... failed with status: 400 Bad Request

whereas on `main` the same LNURL call returns

```json
{
  "ErrorStatus": {
    "data": {
      "reason": "insufficient funds"
    }
  }
}
```

These LNURL error objects are useful to apps integrating with us, as they can surface the error reason to the end-user.

ok300: How about #990? IMO it covers both cases.

ok300:
Resolved conflicts with `main`.

ok300 (approved):

libs/sdk-core/src/input_parser.rs
dangeross: I think it would be good to add test cases for the two ways a response is handled

JssDWt: HTTP status has (should have) no meaning at all in LNURL: https://github.com/lnurl/luds/blob/luds/01.md#http-status-codes-and-content-type

roeierez: @dangeross added test. Now both cases are tested.

libs/sdk-core/src/input_parser.rs
```diff
@@ -1075,15 +1075,24 @@ pub(crate) mod tests {
...
             Some(err_reason) => {
                 ["{\"status\": \"ERROR\", \"reason\": \"", &err_reason, "\"}"].join("")
             }
         };
 
+        let status = match &return_lnurl_error {
+            None => 200,
+            Some(_) => 400,
+        };
+
```
erdemyerebasmaz, line 1089: I think we can match once and return the result as tuple here:
```
let (response_body, status) = match &return_lnurl_error {
    None => (expected_lnurl_withdraw_data, 200),
    Some(err_reason) => (
        ["{\"status\": \"ERROR\", \"reason\": \"", &err_reason, "\"}"].join(""),
        400,
    ),
};
```

erdemyerebasmaz (approved):
Changes look good
Have a recommendation regarding redundant match on `&return_lnurl_error`

libs/sdk-core/src/input_parser.rs
erdemyerebasmaz: @roeierez `cargo clippy` has this warning:
```
error: this expression creates a reference which is immediately dereferenced by the compiler
    --> sdk-core/src/input_parser.rs:1081:60
     |
1081 |                 ["{\"status\": \"ERROR\", \"reason\": \"", &err_reason, "\"}"].join(""),
     |                                                            ^^^^^^^^^^^ help: change this to: `err_reason`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
     = note: `-D clippy::needless-borrow` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::needless_borrow)]`
```

PR #990, opened 2024-05-27
Allow HTTP status code check in GET response parsing

ok300:
We want to support two kinds of GET response parsing:
- first, where an error HTTP status code results automatically in `Err` (mempool.space, boltz API)
- second, where the HTTP status code is not checked, but we attempt to parse the payload into possibly an error payload (LNURL)

This PR adds support for the first type, in addition to the second which was already supported.

### DIFF
--- a/libs/sdk-core/src/input_parser.rs
+++ b/libs/sdk-core/src/input_parser.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 
 use anyhow::{anyhow, Result};
 use bip21::Uri;
+use reqwest::StatusCode;
 use serde::Deserialize;
 use serde::Serialize;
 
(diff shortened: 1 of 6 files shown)


libs/sdk-core/src/input_parser.rs
```diff
@@ -258,39 +259,51 @@ pub(crate) async fn post_and_log_response(url: &str, body: Option<String>) -> Sd
...
+pub(crate) async fn get_parse_and_log_response<T>(
+    url: &str,
+    enforce_status_check: bool,
+) -> SdkResult<T>
 where
     for<'a> T: serde::de::Deserialize<'a>,
 {
-    let raw_body = get_and_log_response(url).await?;
+    let (raw_body, status) = get_and_log_response(url).await?;
+    let parse_res = serde_json::from_str::<T>(&raw_body);
```
roeierez, line 298: Better to move this line after the enforce_status_check block

roeierez (changes requested):
Looks good. A small comment.

libs/sdk-core/src/input_parser.rs
ok300: Done: f9ecc4e8936e3846dcbfba4a3e38672acdc11b36

PR #1000, opened 2024-06-10
Use remote fee policy when constructing route hints

### LINKED ISSUE #999: Route hints may contain local fees instead of remote
If greenlight doesn't include the lsp route hint in a normal invoice (no channel open), we currently add the _local_ routing fees in the route hint rather than the _remote_ routing fees.

Relevant code here:
https://github.com/breez/breez-sdk/blob/2ff9646d03eac9f737834fdaeedc871d640d9649/libs/sdk-core/src/greenlight/node_api.rs#L1693-L1696

The code there has to be updated to use the remote policy (another call to the cln api `listchannels` api probably) to include the remote policies.
Another, additional, option can be to replace routing fees with the fees coming from `lsp_information` if the hint was not in the invoice originally, but it is included in the `get_routing_hints` call.

dangeross:
Fixes #999

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -47,9 +47,9 @@ use crate::bitcoin::{
 use crate::invoice::{parse_invoice, validate_network, InvoiceError, RouteHintHop};
 use crate::lightning::util::message_signing::verify;
 use crate::lightning_invoice::{RawBolt11Invoice, SignedRawBolt11Invoice};
-use crate::models::*;
 use crate::node_api::{CreateInvoiceRequest, FetchBolt11Result, NodeAPI, NodeError, NodeResult};
 use crate::persist::db::SqliteStorage;
+use crate::{models::*, LspInformation};
 use crate::{
     NodeConfig, PrepareRedeemOnchainFundsRequest, PrepareRedeemOnchainFundsResponse, RouteHint,
 };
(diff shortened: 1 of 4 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1671,32 +1672,71 @@ impl NodeAPI for Greenlight {
...
+                            peer_channel.peer_id.ok_or(anyhow!("no peer id"))?,
+                        ),
                         short_channel_id: scid,
-                        fees_base_msat: c.fee_base_msat.clone().unwrap_or_default().msat as u32,
-                        fees_proportional_millionths: c
-                            .fee_proportional_millionths
-                            .unwrap_or_default(),
+                        fees_base_msat: fees_base_msat.unwrap_or(peer_fee_base_msat),
+                        fees_proportional_millionths: fees_proportional_millionths
+                            .unwrap_or(peer_fees_proportional_millionths),
```
JssDWt, line 1732: If the remote fee policy cannot be found, we should not include the route hint in the response. It should never make sense to include the local channel policy in a route hint.

JssDWt (commented):
One comment. I like that `listchannels` call where the local node is the destination. Smart.

roeierez (approved):
LGTM

libs/sdk-core/src/greenlight/node_api.rs
dangeross: Ok, I changed it so that it only adds route hints where the source channel is found

JssDWt (commented):
I think this should be OK. 

Thinking about it, I think there's still the possibility I suppose that there is an open channel with the LSP, but no route hint will be added, because we don't have the remote channel policy of the lsp. 

So in the case:
- no route hint in the invoice
- no remote channel policy
- with active channel

There will not be an LSP hint added.

roeierez (approved):
LGTM

Issue #998, opened 2024-06-08
Making validate_lnurl_pay public


darioAnongba:
Hi all,

It seems that the `validate_lnurl_pay` function of the SDK (Rust) is a standalone function that can be made public as it is generic. It looks a lot like the `parse` function and could be beneficial to many apps that can use it even without using the `BreezServices`. 

What do you think ? 

Cheers,

roeierez:
Indeed. We are about to extract out some generic logic (including these functions) to a different crate which the sdk will depend on so the functionality will be exposed as public.

darioAnongba:
If I may add my 2 cents from the experience gained from implementing the SDK:
- Decoding the SuccessAction with the `payment_preimage` deserves to be part of the public package as well (reason below)

- Having a `lnurl_pay` function that does multiple steps in the background seems to not be the ideal behaviour because the validation of the input happens inside the function, which forces it to return an object with multiple possible error types (`LnUrlPayResult`). The steps for LNURL pay are:
  1. parse
  2. validate_lnurl_pay
  3. send_payment (normal bolt11 one) from a `CallbackResponse`
  4. decode_success_action
  
As you can see, only step iii. is an adapter action on the actual node. Currently the function can either fail with  `LnUrlPayError` or succeed with an object that itself can be 2 different types of errors. Also, the payment can succeed and the function still fail because of line 420: `sha256::Hash::from_str(&details.payment_preimage)?;` or line 445: `self.persister.insert_payment_external_info`.

All of that to say that `lnurl_pay` should probably be a function that accepts a `CallbackResponse` and does nothing else. Up to the user to generate that `CallbackResponse` using the public `parse` and `validate_lnurl_pay`. This way, the payment that happens externally and the validation are separated.

ok300:
Related to #668.

PR #1002, opened 2024-06-11
Extract `sdk-common` crate

ok300:
- [x] LNURL-auth
  - extracted `sdk-utils` for REST utils used in both `sdk-lnurl` and in the SDK
  - changed `LnUrlError::InvalidInvoice` from accepting an inner `InvoiceError` to a simple `String`, because `InvoiceError` is not available in `sdk-lnurl`
  - renamed `InputType::LnUrlError` to `LnUrlEndpointError` to avoid naming conflicts with the `LnUrlError` struct
  - [x] Fix FRB
- [x] LNURL-pay
  - [x] How to handle `Payment` in `LnUrlPaySuccessData`?
- [x] LNURL-withdraw
- [x] `invoice` module
- [x] `input_parser` module
- [x] Fix tests


erdemyerebasmaz:
@ok300 Will Fiat Currency module extracted separately, as I believe it relies on server url from Config?

libs/sdk-core/src/lib.rs
```diff
@@ -203,13 +202,178 @@ pub use breez_services::{
...
-};
-pub use invoice::{parse_invoice, LNInvoice, RouteHint, RouteHintHop};
-
-pub use lnurl::pay::model::*;
 pub use lsp::LspInformation;
 pub use models::*;
+pub use sdk_common::prelude::*;
 pub use swap_out::reverseswap::{ESTIMATED_CLAIM_TX_VSIZE, ESTIMATED_LOCKUP_TX_VSIZE};
+
+// === FRB mirroring
```
roeierez, line 210: Can we move all the mirroring definitions to the binding.rs? They are only used in flutter.

roeierez (changes requested):
Looks good.
I think it is better that at first step the lnurl result structures will be defined in the sdk-core/sdk-liquid since:
1. They depends on payment which is different and implies that this is a result that belongs to the specific implementation
2. It has PayError which the lnurl logic is not aware of (paying is not part of lnurl exposed logic, it is the specific crate action).

libs/sdk-core/src/lib.rs
ok300: I tried different ways to do this, but unfortunately it doesn't work.

FRB mirroring expects the mirrors to be found in the root namespace of the crate, so `crate::*`. If I place them in `binding`, it fails compilation. This fails even if I re-export them in `lib.rs` as `pub use binding::*`, which should make them visible in the root namespace.

libs/sdk-common/src/lnurl/specs/pay.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// <https://github.com/lnurl/luds/blob/luds/09.md>
+    ///
+    /// * `EndpointError` indicates a generic issue the LNURL endpoint encountered, including a freetext
+    /// field with the reason.
+    ///
+    /// * `PayError` indicates that an error occurred while trying to pay the invoice from the LNURL endpoint.
+    /// This includes the payment hash of the failed invoice and the failure reason.
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    #[allow(clippy::large_enum_variant)]
+    pub enum LnUrlPayResult {
```
roeierez, line 156: Is this enum in use? I don't see it in use. Also it has a variant for PayError which indicates it should be located at the breez-sdk level.

libs/sdk-common/src/invoice.rs
```diff
@@ -75,6 +75,12 @@ impl From<SystemTimeError> for InvoiceError {
     }
 }
 
+impl From<InvoiceError> for LnUrlError {
```
roeierez, line 78: Perhaps move this to to the lnurl module?

roeierez (changes requested):

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -1,57 +1,212 @@
...
+// END sdk-common mirror imports
+////////////////////////////////
+
+
+//////////////////////////////////
+// BEGIN sdk-common wrappers
+// These are connecting structures that glue relevant sdk-common structs to the SDK
+
+[Enum]
+interface WrappedLnUrlPayResult {
```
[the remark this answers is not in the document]
ok300: > Any reason why we call it `WrappedLnUrlPayResult`?
Because it partially overlaps, partially adds extra data to the `sdk-common::LnUrlPayResult`.

The discussed approach to stick to the `sdk-common` returned structs, and duplicate them in the SDK where necessary, unfortunately didn't work so I needed a different approach.

The `sdk-common` LNURL-pay methods directly return a

```rust
pub enum LnUrlPayResult {
    EndpointSuccess { data: LnUrlPaySuccessData },
    EndpointError { data: LnUrlErrorData },
    PayError { data: LnUrlPayErrorData },
}
```

but in case of success, the SDK has to include the `Payment`, so the SDK method returns

```rust
 // Wrapped.. because inner data below
pub enum WrappedLnUrlPayResult {
    // Inner data = sdk-common struct + payment
    EndpointSuccess { data: WrappedLnUrlPaySuccessData },

    // Rest is identical
    EndpointError { data: LnUrlErrorData },
    PayError { data: LnUrlPayErrorData },
}
```

>  I think it should be as before, no need for changing the naming.

One way to achieve that would be
- rename `Wrapped..` into the `sdk-common` names (so to the SDK user they're unchanged)
- rename the `sdk-common` names into `Base..` or `Raw..`

What do you think?

---

> is there any other changes to the udl? If not can we revert to the original udl file because as it is currently it is difficult to detect the changes.

Done, I moved segments around to simplify the diff for easier review: 0df7665685bea6787980111d48af8179098e29fd

I would still move them back after review, to make the UDL more structured:
- `sdk-common` definitions
- "glue" or wrapper definitions
- SDK definitions

libs/sdk-core/src/lnurl/pay.rs
```diff
@@ -1,375 +1,40 @@
...
-    }
+use sdk_common::prelude::*;
+use serde::Serialize;
+
+use crate::Payment;
+
+/// Wrapper around [LnUrlPayResult], which includes the [Payment] on success
+#[derive(Serialize)]
+#[allow(clippy::large_enum_variant)]
+pub enum WrappedLnUrlPayResult {
```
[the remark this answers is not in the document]
ok300: > Can we move all the tests to the sdk-common

Unfortunately not, as these are the E2E tests, which test how the service methods behave (`lnurl_pay`, etc) in different success / failure scenarios.

IIRC, we discussed these belong in the SDK.

libs/sdk-common/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+pub mod input_parser;
+pub mod invoice;
+mod lnurl;
+mod model;
+mod utils;
+
+#[rustfmt::skip]
```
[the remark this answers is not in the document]
ok300: Yes, these `rustfmt::skip` annotations were there before, AFAIK @erdemyerebasmaz needed them so imports are in a certain order for FRB.

You can view this file's (`libs/sdk-core/src/lib.rs`) full diff here: https://github.com/breez/breez-sdk/pull/1002/files#diff-748ba44c3a4b363292f955b3fc58dd7f0b5ce9e654d7e62621064241ad0472b5

libs/sdk-common/src/lnurl/specs/pay.rs
ok300: Done, removed: 64b4dc16a2f3c32b93f7079ebdcc0a9d1499070a

In the SDK level, its represented by the `Wrapped..` version.

libs/sdk-common/src/invoice.rs
ok300: Unfortunately Rust won't allow it:
> Only traits defined in the current crate can be implemented for arbitrary types

`InvoiceError` is defined here in `invoice.rs`.

libs/sdk-common/src/lib.rs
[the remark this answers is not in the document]
roeierez: OK, in that case let's not change it.

libs/sdk-common/src/invoice.rs
roeierez: Not following. lnurl module still exists at the same crate as invoice.rs (sdk-common).

libs/sdk-bindings/src/breez_sdk.udl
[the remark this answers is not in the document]
roeierez: I think we just need to remove the Wrapped prefix and stay with the old names. I see no problem that both sdk-common and sdk-core use the same names as they exist in different modules and creates.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -720,7 +720,7 @@ interface InputType {
     LnUrlPay(LnUrlPayRequestData data);
     LnUrlWithdraw(LnUrlWithdrawRequestData data);
     LnUrlAuth(LnUrlAuthRequestData data);
-    LnUrlError(LnUrlErrorData data);
+    LnUrlEndpointError(LnUrlErrorData data);
```
roeierez, line 723: Do we have to change the naming here? It is better to leave the current API unmodified in this refactoring.

libs/sdk-core/src/lnurl/pay.rs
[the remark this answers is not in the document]
roeierez: There are lots of tests that do not involve E2E. For example all the tests in pay.rs seems purely belong to sdk-common. Also all the tests in withdraw.rs seems unit tests that belong to the sdk-common.

roeierez (changes requested):

ok300:
Closes #998

libs/sdk-common/src/invoice.rs
ok300: Moved it in `lnurl/error.rs`, next to where `LnUrlError` is defined: f7a977ccd37912afbfdb9e710738daa4a5309f07

libs/sdk-bindings/src/breez_sdk.udl
ok300: Done: fcebd47fd8bfb369d9d829cd0d42c6548b214390

It was needed due to a FRB conflict with the identically named struct, but that was fixed.

libs/sdk-bindings/src/breez_sdk.udl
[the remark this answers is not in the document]
ok300: Done: b1b249129eda9a80231ae3af8101700db5ade97d

libs/sdk-core/src/lnurl/pay.rs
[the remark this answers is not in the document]
ok300: Done: dfe28ce37eed5798af2187106b30a6ba854669d0 , c3c76c4479e7b120c9387c5047f3ff56f4ea393c

I moved the non-E2E tests to `sdk-common`. The remaining ones use things like `MockBreezServer` and / or mock an HTTP server for E2E LNURL flows.

roeierez (approved):
LGTM

PR #1004, opened 2024-06-13
Propagate timeout in LNURL withdraw flow

andrei-21:
We have an issue with withdrawing from lnbits, sometimes we get `operation timed out`, but after some time the payment usually arrives. The reason for that that the SDK has 30 seconds timeout for http requests, when the wallet submits an invoice, lnbits tries to pay it synchronously. If the payment takes more than 30 seconds the SDK fails the request and returns the error, but lnbits is still paying the invoice and likely will pay.
In that case the issue is on lnbits side (it should not wait for successful payment before responding to the http request), I reported them https://github.com/lnbits/withdraw/issues/37. But I afraid the are not going to fix it soon and not every instance is going to upgrade quick.

Here is a proposal to propagate this timeout back to the client and let him decide if the request needs to be retried, tolerated, or failed.

### DIFF
--- a/libs/sdk-common/src/lnurl/specs/withdraw.rs
+++ b/libs/sdk-common/src/lnurl/specs/withdraw.rs
@@ -1,5 +1,6 @@
 use std::str::FromStr;
 
+use crate::error::{ServiceConnectivityError, ServiceConnectivityErrorKind};
 use crate::prelude::*;
 
 /// Validates invoice and performs the second and last step of LNURL-withdraw, as per
@@ -38,7 +39,10 @@ pub async fn validate_lnurl_withdraw(
             data: LnUrlWithdrawSuccessData { invoice },
         },
         Ok(LnUrlCallbackStatus::ErrorStatus { data }) => LnUrlWithdrawResult::ErrorStatus { data },
-        Err(e) if e.to_string().contains("operation timed out") => LnUrlWithdrawResult::Timeout {
+        Err(ServiceConnectivityError {
+            kind: ServiceConnectivityErrorKind::Timeout,
+            err: _,
+        }) => LnUrlWithdrawResult::Timeout {
             data: LnUrlWithdrawSuccessData { invoice },
         },
         Err(e) => return Err(LnUrlError::ServiceConnectivity(e.to_string())),
--- a/libs/sdk-common/src/lnurl/specs/withdraw.rs
+++ b/libs/sdk-common/src/lnurl/specs/withdraw.rs
@@ -33,14 +33,15 @@ pub async fn validate_lnurl_withdraw(
 
     // Send invoice to the LNURL-w endpoint via the callback
     let callback_url = build_withdraw_callback_url(&req_data, &invoice)?;
-    let callback_res: LnUrlCallbackStatus = get_parse_and_log_response(&callback_url, false)
-        .await
-        .map_err(|e| LnUrlError::ServiceConnectivity(e.to_string()))?;
-    let withdraw_status = match callback_res {
-        LnUrlCallbackStatus::Ok => LnUrlWithdrawResult::Ok {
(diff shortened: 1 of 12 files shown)


dangeross:
Sorry for the slow review @andrei-21, can you resolve the conflicts?

roeierez:
I am fine with the solution. We may also increase the timeout to 1 minute in this case which will also make the case work in most of the times.
Needs rebase.

roeierez (approved):
LGTM

dangeross (approved):

libs/sdk-common/src/lnurl/specs/withdraw.rs
```diff
@@ -33,14 +33,15 @@ pub async fn validate_lnurl_withdraw(
...
-        .map_err(|e| LnUrlError::ServiceConnectivity(e.to_string()))?;
-    let withdraw_status = match callback_res {
-        LnUrlCallbackStatus::Ok => LnUrlWithdrawResult::Ok {
+    let withdraw_status = match get_parse_and_log_response(&callback_url, false).await {
+        Ok(LnUrlCallbackStatus::Ok) => LnUrlWithdrawResult::Ok {
             data: LnUrlWithdrawSuccessData { invoice },
         },
-        LnUrlCallbackStatus::ErrorStatus { data } => LnUrlWithdrawResult::ErrorStatus { data },
+        Ok(LnUrlCallbackStatus::ErrorStatus { data }) => LnUrlWithdrawResult::ErrorStatus { data },
+        Err(e) if e.to_string().contains("operation timed out") => LnUrlWithdrawResult::Timeout {
```
ok300, line 41: I don't know if this is the best solution. It makes this conditional on the exact error string from lnbits, which can change anytime even with a simple HTTP server upgrade.

Perhaps a more robust approach is to make the REST client timeout configurable in the SDK config? Then clients that expect to interact with slower endpoints can bump it to 60s?

Or maybe we can just bump the default from 30 to 60s?

What do you guys think?

ok300: Not against merging this, just wondering if there's a better way.

roeierez: @andrei-21 since it is lnbit specific (which is also against the spec) I think @ok300 has a point.
What do you think? bumping to 60 seconds will solve this issue?

ok300: The timeout idea was a practical suggestion for what appears to be a temporary problem, namely that lnbits is not (yet) spec-conform in handling LNURL-withdraws.

The `is_timeout` idea makes sense.

A few thoughts on the details: The right place to handle it is probably `get_and_log_response`, which is where the reqwest error originates. This means the error it typically throws, `ServiceConnectivityError`, would need two variants, a `Generic { err }` and a `Timeout`. This would propagate all the way to `get_parse_and_log_response` and `validate_lnurl_withdraw`, where it can result in the `LnUrlWithdrawResult::Timeout` you added.

ok300: @andrei-21 I have a general question about this.

AFAIK the timeout would not distinguish between
- a connection issue (network, etc)
- a server issue (down for maintenance, etc)
- the synchronous payment issue (lnbits trying to pay the invoice, which takes >30s, so timeout)

In the PR its written
> ... a proposal to propagate this timeout back to the client and let him decide if the request needs to be retried, tolerated, or failed.

Isn't it a bad idea for the client to retry a timed-out Withdraw, not knowing the actual root cause of the timeout? Couldn't the 2nd, 3rd Withdraw call lead to a 2nd and 3rd actual withdrawal, and not just a "retry"?

Or am I missing something and there's a way to tell the difference?

PR #1007, opened 2024-06-17
greenlight: use remote cltv delta in route hints

JssDWt:
Next to the recently modified [base fee and fee rate](https://github.com/breez/breez-sdk/pull/1000), we now also use the cltv delta from the remote routing policy in the route hints.

I've dropped in some small optimizations on iterations and cloning, so this function is less resource intensive when there are many channels.


roeierez (approved):
LGTM

dangeross (approved):
Looks good!

PR #1009, opened 2024-06-17
log reason for invoice wrapping failure

JssDWt:
There was a case where the preimage was already on the node, but wrapping the invoice fails. The logs don't contain the reason for failure wrapping the invoice. They contain `Preimage already known, invoice found, failed to ensure route hint`. This PR adds the failure reason to the logs.


roeierez (approved):
LGTM

PR #1012, opened 2024-06-18
Extract LNURL-auth structs to `sdk-common`

ok300:
This PR makes `sdk-common` usable externally for LNURL-auth:
- moves the needed structs in `sdk-common`
- re-exports the `bitcoin` crate, so callers can use the error conversion between `bip32::Error` and `LnUrlAuthError` defined in `sdk-common`, which is only usable if the caller also uses the exact `bitcoin` crate version we're using. To spare callers the effort of keeping aligned with our version, especially across updates, we re-export `bitcoin` here, so callers can use it as `sdk_common::bitcoin`.


libs/sdk-common/src/lib.rs
```diff
@@ -4,6 +4,12 @@ mod lnurl;
 mod model;
 mod utils;
 
+// Re-export, to make it easy for callers to refer to the specific bitcoin crate version we're using
+// This is important because certain error conversions defined in this crate map to or from structs
+// from this specific bitcoin crate version. If the caller would use a different version, the Into
+// traits defined here would not be usable by them (e.g. impl From<bip32::Error> for LnUrlError)
+pub use bitcoin;
```
dangeross, line 11: Should we re-export `lightning` and `lightning_invoice` also?

roeierez: @dangeross  Do we use it directly in the sdk or is it use internally in sdk-common only?

ok300: In the SDK we use the `lightning` indirectly. For example we use `boltz_client::Bolt11Invoice` which is a boltz re-export of `lightning_invoice::Bolt11Invoice`.

We also use `bitcoin` indirectly, e.g. `lwk_wollet::bitcoin::Witness`.

IMO re-exporting them here does no harm for `sdk-common` and can help Liquid SDK depend less on external crates.

> Should we re-export `lightning` and `lightning_invoice` also?

Done: 6915cf048771067568ffe86c0db0acc64701b0af

roeierez (approved):
LGTM

PR #1019, opened 2024-06-20
`sdk-common`: remove `rusqlite` dependency

ok300:
This PR removes `rusqlite` as a dependency from `sdk-common`, which is a pre-requisite for https://github.com/breez/breez-liquid-sdk/pull/308 .


roeierez (approved):
LGTM

PR #1021, opened 2024-06-21
Breez Server: add ability to fetch Boltz Swapper URLs

ok300:
This PR is a pre-requisite for https://github.com/breez/breez-liquid-sdk/issues/135

### DIFF
--- a/libs/sdk-common/Cargo.toml
+++ b/libs/sdk-common/Cargo.toml
@@ -22,7 +22,12 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 strum_macros = { workspace = true }
 thiserror = { workspace = true }
-tonic = { workspace = true }
+tonic = { workspace = true, features = [
+    "tls",
+    "transport",
+    "tls-roots",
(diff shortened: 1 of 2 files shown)


libs/sdk-common/Cargo.toml
```diff
@@ -22,7 +22,12 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 strum_macros = { workspace = true }
 thiserror = { workspace = true }
-tonic = { workspace = true }
+tonic = { workspace = true, features = [
+    "tls",
+    "transport",
+    "tls-roots",
+    "tls-webpki-roots",
+] }
```
roeierez, line 31: What was the requirement to add these features?

ok300: Without these, gRPC calls to the Breez Server failed.

Looks like `tonic` doesn't include basic tls features in its default setup, so they have to be added explicitly. We have these extra features in Breez SDK too, I assume for the same reason.

PR #1022, opened 2024-06-24
use `to_us_msat` for `pending_onchain_balance`

JssDWt:
Previously `pending_onchain_balance` was defined by `min_to_us_msat`. However, it confused users, because it appears that the funds were gone while eventually they would resolve to the user. Theoretically the peer could steal the difference between `to_us_msat` and `min_to_us_msat`, but since this scenario is highly unlikely given we have a watchtower and/or the node will come online every week or so. So use `to_us_msat` instead, which should reflect the real balance the user is receiving in 99.99% of the scenarios.


roeierez (approved):
LGTM

dangeross (approved):

PR #1024, opened 2024-06-26
Wait for channel to be swept before marking as closed

roeierez:
When a channel is force closed, during the time period between the closing tx confirmation and the fully swept funds the channels is till returned in `listpeerchannels` API and its status is set to `onchain`.
Perviously we wrongly marked this channel as Closed while it should be still PendingClose as it is not fully swept.
Once the channel will be removed from this list it will be marked as closed.
Note that it might take a delay of 100 blocks after the sweep is finished but this is better than having it closed and misleading the user potentially for a much longer time.

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -2188,7 +2188,6 @@ impl From<cln::ListpeersPeersChannels> for Channel {
                 ChannelState::PendingOpen
             }
             ChanneldNormal => ChannelState::Opened,
-            Onchain => ChannelState::Closed,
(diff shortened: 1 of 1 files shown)


dangeross (approved):
Looks good. So as soon as its fully swept it appears in `node_api.list_closed_channels()` during the sync/pull_changes?

roeierez:
> Looks good. So as soon as its fully swept it appears in `node_api.list_closed_channels()` during the sync/pull_changes?

Yes. We also have a defensive mechanism that mark channels as closed if they do not appear in any of the lists (channels, closed_channels).

JssDWt (approved):
LGTM

PR #1027, opened 2024-07-01
chore: update ci deps

JssDWt:
- update used action versions 
- update the used protoc version


dangeross (approved):

ok300 (approved):
Looks good, long overdue!

Esp. the protobuf update, as we're already using `proto3` in `breez.proto`.

PR #1031, opened 2024-07-04
Use json as command output

### LINKED ISSUE #939: JSON output for dev commands

roeierez:
fixed #939


erdemyerebasmaz (approved):
LGTM

Let's not forget to close 
- https://github.com/breez/breez-sdk/pull/988

ok300 (approved):

PR #1034, opened 2024-07-08
Trampoline payments

JssDWt:
This PR adds trampoline payment functionality to the SDK. 

If a client sends a payment, normally the greenlight node will find routes through the gossip graph, and attempt to pay over those routes. If a route fails, it tries again. For every attempt there are multiple rounds where the greenlight node has to communicate with the signer. Trampoline payments have the advantage there will be only one outgoing payment attempt, and the LSP will do the retrying, therefore less communication rounds are necessary, hopefully improving payment performance. The tradeoff is the user may pay more fees for the route, and the user loses some privacy in the current setup, because the LSP will learn the payment destination. The initial default trampoline policy will be 0,5%.

This PR leans on the corresponding greenlight PR for trampoline payments, which is still a work in process: https://github.com/Blockstream/greenlight/pull/475

The LSP will need to run this plugin: https://github.com/breez/trampoline

Some notes about this PR:
- Trampoline payments should only be attempted when the LSP is not in the route hint. This is because the LSP side trampoline plugin doesn't cooperate nicely with lspd in it's current state.
- The trampoline payments are automatically fetched with `listpays`. That's nice, because it doesn't require any changes to sync. The payments from `listpays` have the wrong `amount_msat` due to the way trampoline payments work. We solve that by putting the actual amount in a payment label for trampoline payments.

TODO:
- needs update of the greenlight commit id before getting merged, since it's pointing at a PR

### DIFF
--- a/libs/sdk-common/src/lnurl/specs/pay.rs
+++ b/libs/sdk-common/src/lnurl/specs/pay.rs
@@ -132,6 +132,10 @@ pub mod model {
         pub data: LnUrlPayRequestData,
         /// The amount in millisatoshis for this payment
         pub amount_msat: u64,
+        /// Trampoline payments outsource pathfinding to the LSP. Trampoline payments can improve
+        /// payment performance, but are generally more expensive in terms of fees and they
+        /// compromise on privacy.
+        pub use_trampoline: bool,
         /// An optional comment for this payment
         pub comment: Option<String>,
         /// The external label or identifier of the [Payment]
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -365,6 +365,7 @@ impl BreezServices {
                 let pay_req = SendPaymentRequest {
                     bolt11: cb.pr.clone(),
                     amount_msat: None,
+                    use_trampoline: req.use_trampoline,
                     label: req.payment_label,
                 };
                 let invoice = parse_invoice(cb.pr.as_str())?;
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1412,15 +1412,15 @@ impl BreezServices {
             .await;
 
         // Sync node state
-        let sync_breez_services = self.clone();
-        match sync_breez_services.persister.get_node_state()? {
+        match self.persister.get_node_state()? {
             Some(node) => {
-                info!("Starting existing node {}", node.id)
+                info!("Starting existing node {}", node.id);
+                self.connect_lsp_peer(node.id).await?;
             }
             None => {
                 // In case it is a first run we sync in foreground to get the node state.
                 info!("First run, syncing in foreground");
-                sync_breez_services.sync().await?;
+                self.sync().await?;
                 info!("First run, finished running syncing in foreground");
             }
         }
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -673,19 +673,22 @@ impl BreezServices {
 
     /// Select the LSP to be used and provide inbound liquidity
     pub async fn connect_lsp(&self, lsp_id: String) -> SdkResult<()> {
-        match self.list_lsps().await?.iter().any(|lsp| lsp.id == lsp_id) {
-            true => {
-                self.persister.set_lsp_id(lsp_id)?;
-                self.sync().await?;
-                if let Some(webhook_url) = self.persister.get_webhook_url()? {
-                    self.register_payment_notifications(webhook_url).await?
-                }
-                Ok(())
+        let lsp_pubkey = match self.list_lsps().await?.iter().find(|lsp| lsp.id == lsp_id) {
+            Some(lsp) => lsp.pubkey.clone(),
+            None => {
+                return Err(SdkError::Generic {
+                    err: format!("Unknown LSP: {lsp_id}"),
+                })
             }
-            false => Err(SdkError::Generic {
-                err: format!("Unknown LSP: {lsp_id}"),
-            }),
+        };
+
+        self.persister.set_lsp_id(lsp_id)?;
+        self.persister.set_lsp_pubkey(lsp_pubkey)?;
+        self.sync().await?;
+        if let Some(webhook_url) = self.persister.get_webhook_url()? {
+            self.register_payment_notifications(webhook_url).await?
         }
+        Ok(())
     }
 
     /// Get the current LSP's ID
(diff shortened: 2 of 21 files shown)


dangeross (commented):
Looking good. I would suggest also:

- Checking that the LSP can do trampoline payments (via an LSPInformation flag?)
- Having a config option to not use trampoline payments (for privacy preserving)
- Is it worth prechecking the outbound liquidity to the LSP before the trampoline payment or can the LSP more efficiently do that? (This is think about the edge case of switching LSP and maybe most/all of the liquidity is still with the old LSP)

roeierez:
Good points @dangeross .
Regarding checking lsp liquidity since this check comes with some latency price for each payment and it is an edge case I would consider trying trampoline and if failing for liquidity fallback to regular payment. I think it should fail fast since the node plugin should check that before sending any htlcs

JssDWt:
@dangeross @roeierez please check my last 5 commits.
Instead of calling lsp_info (which is a network call) I added two changes:
- the LSP pubkey is now cached for quick lookup.
- the node_state now contains the peer's 'features' as well (currently only trampoline is a supported feature)

With these two pieces of information we can check:
- does the route hint contain a hint to our LSP?
- Does the LSP support trampoline?

Next to that I added a flag in the SendPaymentRequest and LnurlPayRequest to signal the user doesn't want to use trampoline.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1140,32 +1232,46 @@ impl BreezServices {
...
+                .connect_peer(node_id.clone(), address.clone())
+                .await
+                .map_err(|e| SdkError::ServiceConnectivity {
+                    err: format!("(LSP: {node_id}) Failed to connect: {e}"),
+                })?;
+            node_state.connected_peers.push(ConnectedPeer {
+                id: node_id.clone(),
+                features,
+            });
+            self.persister.set_node_state(&node_state)?;
```
dangeross, line 1271: I'm wondering if updating the node_state's connected peers could done in `node_api.connect_peer()`, but AFAIK this is the only time its called

libs/sdk-core/src/models.rs
```diff
@@ -839,6 +871,11 @@ pub struct ReceivePaymentResponse {
 pub struct SendPaymentRequest {
     /// The bolt11 invoice
     pub bolt11: String,
+    /// By default, if the LSP supports it, trampoline payments will be tried. Trampoline payments
+    /// outsource pathfinding to the LSP. Trampoline payments improve payment performance, but are
+    /// generally more expensive in terms of fees and they compromise on privacy. Setting this flag
+    /// to `true` will ensure no trampoline payment is attempted.
+    pub no_trampoline: bool,
```
dangeross, line 878: I think I would find `skip_trampoline` slightly more descriptive, but `no_trampoline` is also good.

dangeross (commented):
Looking good @JssDWt

libs/sdk-core/src/models.rs
JssDWt: It's always hard to find a good name for a negative boolean. I considered skip_trampoline, but I found no_trampoline more descriptive and a little less difficult to understand.

JssDWt:
I had a case where the trampoline payment was not attempted. I don't think this mechanism is robust enough.

JssDWt:
The status of this PR today is:
- Trampoline payments work. 
- There is a transport error from greenlight's trampoline plugin that makes it seem the payment failed.
- The peer feature check doesn't seem robust enough. Trampoline payments were skipped at least once when testing. It could be we have to remove the following lines so the greenlight trampoline plugin is in charge of the feature check. If these lines are removed, the corresponding peerfeatures migration and the udl change can also be removed.
https://github.com/breez/breez-sdk/pull/1034/files#diff-6f499d3ded5e493c1bd95b827a0a93aeb3d2bfa3d04a6f26be62d87aff026e43R370-R384 
- I wonder whether `trampoline_pay` has the same assurences as `pay` in the sense that it should always work and wait for the node to be available.

JssDWt:
- Trampoline payments now are opt-in through the `use_trampoline` flag.
- Removed the feature bit check on the client side, meaning the client will always call `trampoline_pay` when opting in to trampoline payments. Even if the LSP doesn't support them. The call should fail fast though.

libs/sdk-common/src/lnurl/specs/pay.rs
```diff
@@ -132,6 +132,11 @@ pub mod model {
         pub data: LnUrlPayRequestData,
         /// The amount in millisatoshis for this payment
         pub amount_msat: u64,
+        /// Trampoline payments outsource pathfinding to the LSP. Trampoline payments can improve
+        /// payment performance, but are generally more expensive in terms of fees and they
+        /// compromise on privacy. Setting this flag to `true` will ensure no trampoline payment is
+        /// attempted.
```
dangeross, line 138: Guess because of the rename
```suggestion
        /// compromise on privacy. Setting this flag to `false` will ensure no trampoline payment is
        /// attempted.
```

libs/sdk-core/src/models.rs
```diff
@@ -847,6 +847,11 @@ pub struct ReceivePaymentResponse {
 pub struct SendPaymentRequest {
     /// The bolt11 invoice
     pub bolt11: String,
+    /// Trampoline payments outsource pathfinding to the LSP. Trampoline payments can improve
+    /// payment performance, but are generally more expensive in terms of fees and they
+    /// compromise on privacy. Setting this flag to `true` will ensure no trampoline payment is
```
dangeross, line 852:
```suggestion
    /// compromise on privacy. Setting this flag to `false` will ensure no trampoline payment is
```

dangeross (commented):
Looks good I think, just 2 issues with the comments

libs/sdk-common/src/lnurl/specs/pay.rs
JssDWt: Thanks, I removed that sentence, it's overly verbose now this is a 'positive' boolean.

libs/sdk-core/src/models.rs
JssDWt: Thanks, I removed that sentence, it's overly verbose now this is a 'positive' boolean.

dangeross (approved):
LGTM.

I imagine at some point people will ask about LSP trampoline fees. Maybe we should follow up at some point to add the fee rate to the LSP info?

JssDWt:
> I imagine at some point people will ask about LSP trampoline fees. Maybe we should follow up at some point to add the fee rate to the LSP info?

Yes.
The fee rate is currently not advertised in our custom protocol. That could be an additional feature. We assume a fixed fee of 0.5%.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1171,32 +1247,40 @@ impl BreezServices {
...
-                        })?;
-                }
-                debug!("connected to lsp {node_id}@{address}");
-            }
+            Some(lsp) => lsp.clone(),
+            None => return Ok(()),
+        };
+
+        self.persister.set_lsp_id(lsp.id)?;
+        self.persister.set_lsp_pubkey(lsp.pubkey.clone())?;
```
roeierez, line 1261: As it seems these functions are always being called together now, perhaps better to have one function that takes these two arguments?

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1147,6 +1155,47 @@ impl NodeAPI for Greenlight {
...
+            .into_inner();
+
+        let client = self.get_node_client().await?;
+
+        // Before returning from send_payment we need to make sure it is
+        // persisted in the backend node. We do so by polling for the payment.
+        // TODO: Ensure this works with trampoline payments
+        // NOTE: If this doesn't work with trampoline payments, the sync also
+        // needs updating.
+        let payment = Self::fetch_outgoing_payment_with_retry(client, result.payment_hash).await?;
```
roeierez, line 1195: About the comment, do we know if that works with trampoline?

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -2114,16 +2163,29 @@ impl TryFrom<cln::ListpaysPays> for Payment {
...
-            .unwrap_or_default();
         let payment_amount_sent = payment
             .amount_sent_msat
             .clone()
             .map(|a| a.msat)
             .unwrap_or_default();
+
+        // For trampoline payments the amount_msat doesn't match the actual
+        // amount. If it's a trampoline payment, take the amount from the label.
+        let (payment_amount, client_label) = serde_json::from_str::<PaymentLabel>(payment.label())
```
roeierez, line 2174: Do we have the bolt11 we paid in case of trampoline? If so perhaps better to take the amount from there? If we don't perhaps it is better to persist it as part of the label? I think from UX perspective the user needs the paid invoice.

roeierez (changes requested):
Looks good!

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: Not yet, for that we need the transport error to be resolved.

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: The invoice is there, but possibly it's a zero amount invoice (which we support). Having the amount in the label handles both cases.

libs/sdk-core/src/breez_services.rs
JssDWt: This is currently in 2 places, and setting lsp id without pubkey is in one place [here](https://github.com/breez/breez-sdk/blob/22023cd8f221337740d76936ff681576aa7b933a/libs/sdk-core/src/breez_services.rs#L2406). If you still think we should make that a single function, I'll fix that. We could make the function check whether the new lsp id is the same, if it's not, erase the lsp pubkey if not supplied.

roeierez: I see, I missed that. Then it is less priority but I think it is slightly better to have one function.

libs/sdk-core/src/greenlight/node_api.rs
roeierez: Right. I think we have a mechanism that adds the amount (in case of zero invoice) to the extrenal_info (persist_pending_payment).
But I am not against of having that in the label which I think is better.

libs/sdk-core/src/breez_services.rs
JssDWt: Now setting these in a single function in 4a0a150457c1d191ae850badb7846571266a2b48

JssDWt:
@dangeross re-requesting review due to this added commit: 4a0a150457c1d191ae850badb7846571266a2b48

JssDWt:
Bumped greenlight because https://github.com/Blockstream/greenlight/pull/475 was merged

tools/sdk-cli/src/commands.rs
```diff
@@ -50,6 +50,10 @@ pub(crate) enum Commands {
         /// The external label or identifier of the payment
         #[clap(name = "label", short = 'l', long = "label")]
         label: Option<String>,
+
+        /// If use_trampoline is set, trampoline payments will be attempted.
+        #[clap(long, action)]
```
dangeross, line 55: Just to check, do you want to flip the action default now it's opt-in? `action=ArgAction::SetFalse`

dangeross: Ignore that

dangeross (approved):

PR #1035, opened 2024-07-09
Register and un-register for notifications on all used LSPs

### LINKED ISSUE #970: Bug: switching LSP causes webhook registration with original LSP to be lost
Given that `register_webhook()` only registers the webhook URL with the currently selected LSP and that webhook registrations expire after a while, users can stop receiving notifications if they switch LSP. We've experienced that with one of our wallets.

Here's the sequence of events:
1. Client selects LSP `lsp_a` and gets a channel from that LSP
2. Client changes the selected LSP to `lsp_b` so that any new channels are opened with `lsp_b`
3. Client registers the webhook for background receive using `register_webhook()`
4. Any payments sent to the user won't trigger a notification because registration has only been made with `lsp_b`, but the user has a single channel with `lsp_a`

Probably the SDK needs to keep track of all LSPs with which it has channels to so that `register_webhook()` registers with all of them.

ok300:
Consider all LSPs used by this node, active and historical, when registering or un-registering for notifications.

Fixes #970

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1891,36 +1891,44 @@ impl BreezServices {
 
     /// Registers for lightning payment notifications. When a payment is intercepted by the LSP
     /// to this node, a callback will be triggered to the `webhook_url`.
+    ///
+    /// Note: these notifications are registered for all LSPs used by this node, active and historical.
     async fn register_payment_notifications(&self, webhook_url: String) -> SdkResult<()> {
         let message = webhook_url.clone();
         let sign_request = SignMessageRequest { message };
         let sign_response = self.sign_message(sign_request).await?;
-        let lsp_info = self.lsp_info().await?;
-        self.lsp_api
-            .register_payment_notifications(
-                lsp_info.id,
-                lsp_info.lsp_pubkey,
-                webhook_url.clone(),
-                sign_response.signature,
-            )
-            .await?;
+        for lsp_info in get_used_lsps(self.persister.clone(), self.lsp_api.clone()).await? {
+            self.lsp_api
+                .register_payment_notifications(
+                    lsp_info.id,
+                    lsp_info.lsp_pubkey,
+                    webhook_url.clone(),
+                    sign_response.signature.clone(),
+                )
+                .await?;
+        }
+
         Ok(())
     }
 
     /// Unregisters lightning payment notifications with the current LSP for the `webhook_url`.
+    ///
+    /// Note: these notifications are unregistered for all LSPs used by this node, active and historical.
     async fn unregister_payment_notifications(&self, webhook_url: String) -> SdkResult<()> {
         let message = webhook_url.clone();
         let sign_request = SignMessageRequest { message };
         let sign_response = self.sign_message(sign_request).await?;
-        let lsp_info = self.lsp_info().await?;
-        self.lsp_api
-            .unregister_payment_notifications(
-                lsp_info.id,
-                lsp_info.lsp_pubkey,
-                webhook_url.clone(),
-                sign_response.signature,
-            )
-            .await?;
+        for lsp_info in get_used_lsps(self.persister.clone(), self.lsp_api.clone()).await? {
+            self.lsp_api
+                .unregister_payment_notifications(
+                    lsp_info.id,
+                    lsp_info.lsp_pubkey,
+                    webhook_url.clone(),
+                    sign_response.signature.clone(),
+                )
+                .await?;
+        }
+
         Ok(())
     }
 
@@ -2610,6 +2618,18 @@ async fn get_lsp_by_id(
         .iter()
         .find(|&lsp| lsp.id.as_str() == lsp_id)
         .cloned())
+}
+
+/// Convenience method to return all LSPs used by this node, active and historical.
+async fn get_used_lsps(
+    persister: Arc<SqliteStorage>,
+    lsp_api: Arc<dyn LspAPI>,
+) -> SdkResult<Vec<LspInformation>> {
+    let node_pubkey = persister
+        .get_node_state()?
+        .ok_or(SdkError::generic("Node info not found"))?
+        .id;
+    lsp_api.list_used_lsps(node_pubkey).await
 }
 
 #[cfg(test)]
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -2481,12 +2481,13 @@ impl PaymentReceiver {
         invoice: &str,
         lsp_info: &LspInformation,
     ) -> Result<String, ReceivePaymentError> {
-        ensure_sdk!(
-            self.node_api.has_active_channel_to(lsp_info).await?,
-            ReceivePaymentError::InvoiceNoRoutingHints {
+        info!("Getting routing hints from node");
+        let (mut hints, has_public_channel) = self.node_api.get_routing_hints(lsp_info).await?;
+        if !has_public_channel && hints.is_empty() {
+            return Err(ReceivePaymentError::InvoiceNoRoutingHints {
                 err: "Must have at least one active channel".into(),
-            }
-        );
+            });
+        }
 
         let parsed_invoice = parse_invoice(invoice)?;
 
@@ -2494,8 +2495,6 @@ impl PaymentReceiver {
         info!("Existing routing hints {:?}", parsed_invoice.routing_hints);
 
         // limit the hints to max 3 and extract the lsp one.
-        info!("Getting routing hints from node");
-        let (mut hints, _) = self.node_api.get_routing_hints(lsp_info).await?;
         if let Some(lsp_hint) = Self::limit_and_extract_lsp_hint(&mut hints, lsp_info) {
             if parsed_invoice.contains_hint_for_node(lsp_info.pubkey.as_str()) {
                 return Ok(String::from(invoice));
@@ -2659,7 +2658,9 @@ async fn get_notification_lsps(
             }
             false => {
                 // Consider only historical LSPs with whom we have an active channel
-                if node_api.has_active_channel_to(&lsp).await? {
+                // TODO
+                let has_active_channel_to_lsp = false;
+                if has_active_channel_to_lsp {
                     notification_lsps.push(lsp);
                 }
             }
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -2467,13 +2467,12 @@ impl PaymentReceiver {
         invoice: &str,
         lsp_info: &LspInformation,
     ) -> Result<String, ReceivePaymentError> {
-        info!("Getting routing hints from node");
-        let (mut hints, has_public_channel) = self.node_api.get_routing_hints(lsp_info).await?;
-        if !has_public_channel && hints.is_empty() {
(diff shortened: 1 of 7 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -2612,6 +2620,18 @@ async fn get_lsp_by_id(
...
+/// Convenience method to return all LSPs used by this node, active and historical.
+async fn get_used_lsps(
+    persister: Arc<SqliteStorage>,
+    lsp_api: Arc<dyn LspAPI>,
+) -> SdkResult<Vec<LspInformation>> {
+    let node_pubkey = persister
+        .get_node_state()?
+        .ok_or(SdkError::generic("Node info not found"))?
+        .id;
+    lsp_api.list_used_lsps(node_pubkey).await
```
roeierez, line 2632: I think perhaps out of the historical lsps only we can filter out those that we don't have channels with since we won't get any notifications from them.

roeierez (changes requested):
Looks good!

libs/sdk-core/src/breez_services.rs
ok300: Done: 4407106d8835a198c28b7a30d3b5a3cbec8317e7

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1736,6 +1736,11 @@ impl NodeAPI for Greenlight {
         }
         Ok((hints, has_public_channel))
     }
+
+    async fn has_active_channel_to(&self, lsp_info: &LspInformation) -> NodeResult<bool> {
```
roeierez, line 1740: This will create several requests for each lsp. It is better IMO to call node api once and get the list of channels which then can be compared to the lsp list.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2612,6 +2635,29 @@ async fn get_lsp_by_id(
...
+    let node_pubkey = persister
+        .get_node_state()?
+        .ok_or(SdkError::generic("Node info not found"))?
+        .id;
+
+    let used_lsps = lsp_api.list_used_lsps(node_pubkey).await?;
+
+    // Only consider LSPs with whom we have a channel
+    let mut lsps_with_active_channel = vec![];
+    for lsp in used_lsps {
```
roeierez, line 2653: We only need to search for active channel in the historical lsps, not in the current lsp. We should register for the current lsp even if not channel exists since it will be used also to open a new channel during payment (we will need the notification in that case).

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: Done: c9dfa4c06d4289ddb92a3c113d4c427ff0978da8

libs/sdk-core/src/lsp.rs
```diff
@@ -108,6 +108,23 @@ impl LspAPI for BreezServer {
         Ok(lsp_list)
     }
 
+    async fn list_used_lsps(&self, pubkey: String) -> SdkResult<Vec<LspInformation>> {
```
JssDWt, line 111: This function name implies the LSP is 'used'. However, the `lsp_full_list` returns any lsp really, also ones that may be unused. Maybe this just needs a rename.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1745,6 +1760,15 @@ impl NodeAPI for Greenlight {
         }
         Ok((hints, has_public_channel))
     }
+
+    async fn get_open_peer_channels(&self) -> NodeResult<HashMap<Vec<u8>, Channel>> {
```
JssDWt, line 1764: Suggestion to create a function called `get_open_peers` instead, returning a `Vec<Vec<u8>>` or `HashSet<Vec<u8>>`. The channel argument is unused. If the channel argument were used, it would have to be a vector of channels.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1924,36 +1924,58 @@ impl BreezServices {
 
     /// Registers for lightning payment notifications. When a payment is intercepted by the LSP
     /// to this node, a callback will be triggered to the `webhook_url`.
+    ///
+    /// Note: these notifications are registered for all LSPs (active and historical) with whom
+    /// we have a channel.
```
JssDWt, line 1929: this function is called by the `connect_lsp` function. I don't think it's necessary to register payment notifications for old LSPs when called from `connect_lsp`

libs/sdk-core/src/breez_services.rs
```diff
@@ -1924,36 +1924,58 @@ impl BreezServices {
...
+        .await?
+        {
+            self.lsp_api
+                .register_payment_notifications(
+                    lsp_info.id,
+                    lsp_info.lsp_pubkey,
+                    webhook_url.clone(),
+                    sign_response.signature.clone(),
+                )
+                .await?;
```
JssDWt, line 1948: There's some magic we have to do for error handling here. I think it doesn't matter when we fail to register notifications for an old LSP, but it does matter whn we try to register for notifications on a new LSP. Perhaps the function should return a response with the LSPs it registered and the ones where registration failed.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1924,36 +1924,58 @@ impl BreezServices {
...
-            .await?;
+        for lsp_info in get_notification_lsps(
+            self.persister.clone(),
+            self.lsp_api.clone(),
+            self.node_api.clone(),
+        )
+        .await?
+        {
+            self.lsp_api
+                .unregister_payment_notifications(
```
JssDWt, line 1970: Unregistering an old LSP probably shouldn't fail unregistering other LSPs.

JssDWt (commented):
Left some comments, mostly about error handling when multiple LSPs are involved.

libs/sdk-core/src/greenlight/node_api.rs
ok300: Done: 1d1038149644b6c7c2718e5840d0f0c67e7f3b0a

libs/sdk-core/src/breez_services.rs
ok300: Done: 1d600f7ada79180684f3333fc31885565446be62

libs/sdk-core/src/breez_services.rs
ok300: Based on the original issue, IMO both cases are just as important: registering notifs for the active or a historical LSP.

However I did update the loop, such that it doesn't return immediately if a register call fails: dfca1fee5a3b51db6a9c3fb0326886e50be19c51

libs/sdk-core/src/lsp.rs
ok300: I tried to come up with other names, but they were either too long (`list_active_and_historical_lsps`) or not accurate enough (`list_linked_lsps`).

Hopefully the rustdoc in the trait is clear enough about the method's purpose:
```rust
    /// List all LSPs, active and historical, used by the given pubkey
    async fn list_used_lsps(&self, node_pubkey: String) -> SdkResult<Vec<LspInformation>>;
```

If you find a better name, LMK and I'll address it in a follow-up.

libs/sdk-core/src/breez_services.rs
ok300: On first `connect_lsp`, there are no historical LSPs, so the loop for historical LSPs won't find anything. On subsequent `connect_lsp` calls when changing LSP, I think it makes sense to register notifications for historical LSPs with open channels (that's the main goal of the PR).

So overall I think the behavior is as intended.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2645,6 +2667,41 @@ async fn get_lsp_by_id(
...
+    lsp_api: Arc<dyn LspAPI>,
+    node_api: Arc<dyn NodeAPI>,
+) -> SdkResult<Vec<LspInformation>> {
+    let node_pubkey = persister
+        .get_node_state()?
+        .ok_or(SdkError::generic("Node info not found"))?
+        .id;
+    let active_lsp_id = persister
+        .get_lsp_id()?
+        .ok_or(SdkError::generic("No active LSP ID found"))?;
```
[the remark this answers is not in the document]
ok300: Done: d38da32576738792f3bb47c4c45ccb3fc7e917bd

libs/sdk-core/src/breez_services.rs
```diff
@@ -1924,36 +1924,58 @@ impl BreezServices {
...
+        {
+            self.lsp_api
+                .register_payment_notifications(
+                    lsp_info.id,
+                    lsp_info.lsp_pubkey,
+                    webhook_url.clone(),
+                    sign_response.signature.clone(),
+                )
+                .await?;
+        }
```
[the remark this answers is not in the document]
ok300: I didn't find an elegant way to do this in parallel while keeping the error handling logic, but maybe it can be added in a follow-up PR if necessary.

libs/sdk-core/src/breez_services.rs
```diff
@@ -2645,6 +2695,39 @@ async fn get_lsp_by_id(
...
+    let node_pubkey = persister
+        .get_node_state()?
+        .ok_or(SdkError::generic("Node info not found"))?
+        .id;
+    let maybe_active_lsp_id = persister.get_lsp_id()?;
+    let open_peers = node_api.get_open_peers().await?;
+
+    let mut notification_lsps = vec![];
+    for lsp in lsp_api.list_used_lsps(node_pubkey).await? {
+        match matches!(maybe_active_lsp_id, Some(ref active_lsp_id) if active_lsp_id == &lsp.id) {
```
roeierez, line 2714: I think we don't need to fetch the current lsp and instead we can identify it by the fee params section. If it exists then we should register for notification (because it means it can open a channel to us).

roeierez (changes requested):

libs/sdk-core/src/breez_services.rs
ok300: Done: 0f0bed0eeb95f5a0612e4ff0e775d8adc9581802

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1650,27 +1676,16 @@ impl NodeAPI for Greenlight {
...
-                let is_private = c.private.unwrap_or_default();
-                has_public_channel |= !is_private;
-                is_private
-                    && c.state == Some(cln::ChannelState::ChanneldNormal as i32)
-                    && c.peer_id.is_some()
-            })
-            .map(|c| (c.peer_id.clone().unwrap(), c))
-            .collect();
+        let open_peer_channels = self.get_open_peer_channels_pb().await?;
+        let has_public_channel = open_peer_channels.values().any(|c| {
```
roeierez, line 1682: This seems wrong as the get_open_peer_channels_pb filters public channels.

roeierez (changes requested):

libs/sdk-core/src/greenlight/node_api.rs
ok300: Fixed: 47ae85e972215a87b02f238cbe82846f2a19991d

roeierez (approved):
LGTM

PR #1037, opened 2024-07-10
`onchain_payment_limits`: include `max_reverse_swap_amount` in result

### LINKED ISSUE #1036: On insufficient balance, `onchain_payment_limits` returns `0` limits
When the user has insufficient balance for a reverse swap, calling `onchain_payment_limits` returns the limits as `0`.

This makes it impossible for the caller to tell what the actual limits are and / or why a reverse swap can't take place.

This behavior was introduced by #850. To fix it, we'll have to either revert #850 and re-introduce `max_reverse_swap_amount`, or include `max_reverse_swap_amount` in the result of `onchain_payment_limits`.

ok300:
Fixes #1036

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -445,7 +445,7 @@ dictionary ReverseSwapFeesRequest {
 dictionary OnchainPaymentLimitsResponse {
     u64 min_sat;
     u64 max_sat;
-    u64 max_sat_with_current_channels;
+    u64 max_payable_sat;
 };
 
 enum SwapAmountType {
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -445,6 +445,7 @@ dictionary ReverseSwapFeesRequest {
 dictionary OnchainPaymentLimitsResponse {
     u64 min_sat;
     u64 max_sat;
+    u64 max_sat_with_current_channels;
 };
 
 enum SwapAmountType {
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1,4 +1,3 @@
-use std::cmp::min;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::str::FromStr;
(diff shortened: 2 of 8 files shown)


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -445,6 +445,7 @@ dictionary ReverseSwapFeesRequest {
 dictionary OnchainPaymentLimitsResponse {
     u64 min_sat;
     u64 max_sat;
+    u64 max_sat_with_current_channels;
```
roeierez, line 448: IMO we don't need it. User can take it from the node info.

roeierez (changes requested):
Looks good. dropped one comment

libs/sdk-bindings/src/breez_sdk.udl
ok300: Are you sure?

Sending onchain is already 3 steps:
- `onchain_payment_limits`
- `prepare_onchain_payment`
- `pay_onchain`

Keeping this out of the `onchain_payment_limits` result means sending onchain will need 4 steps.

roeierez: It seems weird to return mixture of onchain limits with user balance to me. Application developers usually already cache the user balance somewhere (and update it whenever they get a change event from the sdk) so I am not sure they will need to call it explicitly during the call.
IMO it is cleaner without it but happy to hear more opinions regarding this.

ok300: The solution discussed (app checking `max_reverse_swap_amount` separately) was already in place for a while, but was removed with #850 to simplify the steps.

How about this as an alternative solution: `onchain_payment_limits` doesn't return the `max_reverse_swap_amount`, but throws an error if that value is below the swapper minimum? This way, instead of returning limits `min = 0, max = 0` as it does now, it would result in an error. Would that work?

roeierez: I don't think so because the user wants to know on any case the swapper original limits. That was the issue from the first place.

ok300: Alright then I will add the separate method.

roeierez (approved):
LGTM

PR #1038, opened 2024-07-10
Use received amount to account for overpayment

### LINKED ISSUE #1026: Bug: Incorrect payment `amount_msat` for Bolt11 overpayments
After receiving an overpayment for a Bolt11 invoice, the corresponding `Payment` fetched using `list_payments` does not include the amount received, only the original amount from the invoice.

dangeross:
Changes Payment mapping to use the received amount instead of invoice amount to account for overpayments.

Example when sending in Phoenix 1500 instead of 1000 msats:
```
Received payment: amount_received_msat: Some(Amount { msat: 1500 }) amount_msat: Some(Amount { msat: 1000 })
```

Fixes #1026

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -1929,7 +1929,7 @@ impl TryFrom<gl_client::signer::model::greenlight::Invoice> for Payment {
             id: hex::encode(invoice.payment_hash.clone()),
             payment_type: PaymentType::Received,
             payment_time: invoice.payment_time as i64,
-            amount_msat: amount_to_msat(&invoice.amount.unwrap_or_default()),
+            amount_msat: amount_to_msat(&invoice.received.or(invoice.amount).unwrap_or_default()),
             fee_msat: 0,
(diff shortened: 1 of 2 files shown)


roeierez (approved):
LGTM

JssDWt:
Looks good.

What about payments that were already inserted with the `amount_msat` amount? Should we update those somehow?

dangeross:
> What about payments that were already inserted with the `amount_msat` amount? Should we update those somehow?

@JssDWt I think if drop the `last_backup_time` cached_item in the migration, then we would re-sync all the payments (paid invoices) from Greenlight. What do you think @roeierez?

JssDWt:
> I think if drop the `last_backup_time` cached_item in the migration, then we would re-sync all the payments (paid invoices) from Greenlight

I think it's the `last_sync_time` we would have to remove there.

dangeross:
> I think it's the `last_sync_time` we would have to drop there.

Oops, yeah

roeierez (approved):
LGTM

JssDWt (approved):
LGTM

PR #1039, opened 2024-07-11
Add `matches_callback_domain` flag to LNURL URL success action

### LINKED ISSUE #1028: LNURL Pay error "Success Action URL has different domain than the callback domain"
We've found that some LNURL providers are not strictly following [LUD-9](https://github.com/lnurl/luds/blob/luds/09.md). Specifically, they use different domains for the action and callback URLs. The SDK doesn't allow this, as defined in the LNURL spec, but it seems most other wallets do.

In out-of-band discussions, we agreed that the main reason for this limitation is trust. LNURL clients should display the domain name, but later, the successful action would take the client to a different domain.

An idea for a solution is to add a field to `UrlSuccessActionData` that warns about the mismatch but still allows the payment to be made.

dangeross:
This PR removes the validation of the success action URL to match the LNURL callback domain and instead sets `matches_callback_domain` on UrlSuccessActionData if there is a mismatch.

Fixes #1028

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -293,6 +293,7 @@ dictionary MessageSuccessActionData {
 dictionary UrlSuccessActionData {
     string description;
     string url;
+    boolean matches_callback_domain;
 };
 
 [Enum]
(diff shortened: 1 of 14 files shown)


roeierez (approved):
LGTM

ok300 (changes requested):
In principle looks good, but how can a caller make use of this?

From what I can tell, there is no way to tell `lnurl_pay` to not pay if callback domains don't match.

ok300:
Is the goal to remove the domain matching check?

If so, `matches_callback_domain` should be propagated to `SuccessActionProcessed`, which is what's returned by `lnurl_pay`. Otherwise the caller has no way of telling the domains matched or didn't match.

roeierez (approved):
LGTM

ok300 (approved):

PR #1042, opened 2024-07-14
Fix number of incoming transactions in swap

roeierez:


dangeross (approved):
LGTM,. Just to confirm, it links to here where it checks against 1 total_incoming_txs?
https://github.com/breez/breez-sdk/blob/d19390283200129f83f42192f01af273743e109f/libs/sdk-core/src/models.rs#L1427-L1430

roeierez:
True

ok300 (approved):

PR #1050, opened 2024-07-19
Fix failed payment response of completed payment

dangeross:
This PR fixes an issue were when completing a payment after having a failed payment of the same bolt11, the failed payment information is returned in the `send_payment` response. This is because when using CLN `list_pays()` filtering by the payment_hash and complete status, the response lists both failed and complete pays. Previously the first is picked from the list, which would be the failed pay.


ok300 (approved):

roeierez (approved):
LGTM

PR #1055, opened 2024-07-26
Fix reverse swap claim tx fee estimation

### LINKED ISSUE #1054: Reverse Swap: chosen claim tx feerate not applied correctly when paying to P2TR
The customer reported this claim tx: https://mempool.space/tx/3afcf2196c238021b99f77efdae078e42ab581ca279076601898ded5b38e3305

where the chosen claim feerate was 4 sat/vB:
```
    "onchain_amount_sat": 183058,
    "sat_per_vbyte": 4,
    "receive_amount_sat": null,
    "cache": {
      "status": "CompletedSeen",
      "lockup_txid": "1714461275fae9db53f47e597611c14d410fb62b7fe3a656730790d5bae45648",
      "claim_txid": "3afcf2196c238021b99f77efdae078e42ab581ca279076601898ded5b38e3305"
    }
```

However the actual feerate used was 3.09 sat/vB.

ok300:
When estimating the claim tx fee for reverse swaps (send onchain), we used a dummy tx with a simple input and a simple output. This was a non-segwit tx with a size of 60 Bytes.

However, this tx structure is simpler than the actual claim tx, which is a segwit tx where the input has a witness with 3 elements: a signature, the preimage and the redeem script. A claim tx has approx. 310 Bytes / 148 vBytes. The actual size varies within a few bytes, depending on the output type and on signature size fluctuations (+/-1 Byte).

This PR extracts the low-level creation of the claim tx in it's own `build_claim_tx_inner`, which it then re-uses
- in the actual claim tx: `create_claim_tx`
- in constructing a dummy tx for the purposes of fee estimation: `build_fake_claim_tx`

This way, the claim tx fee estimation is based on a realistic tx.

The PR is split in 2 commits, for easier review: extracting the low-level `build_claim_tx_inner` and adding `build_fake_claim_tx` on top.

Fixes #1054

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -229,7 +229,6 @@ jobs:
         working-directory: libs/sdk-react-native 
         run: |
           yarn global add tslint typescript
-          brew update
           brew install kotlin ktlint swiftformat
           make react-native
 
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -2015,6 +2015,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-conv"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+
+[[package]]
 name = "num-integer"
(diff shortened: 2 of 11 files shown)


roeierez (approved):
LGTM

PR #1058, opened 2024-07-29
SDK-Common: Add BIP21 URI parsing for Liquid

hydra-yse:
Necessary for breez/breez-sdk-liquid#405.

Ideally, we could create a fork of the `bip21` crate and add the liquid logic there, but that approach requires quite some time and may not necessarily be merged by the author. Instead, I added a rudimentary BIP21 parser, taking inspiration from Boltz client's crate, so that our `parse` method can now support all the necessary types.


libs/sdk-common/src/input_parser.rs
```diff
@@ -175,6 +175,27 @@ pub async fn parse(input: &str) -> Result<InputType> {
         };
     }
 
+    // Covers BIP 21 URIs and simple onchain Liquid addresses (which are valid BIP 21 with the 'liquid:' or 'liquidtestnet:' prefix)
+    if let Ok(liquid_address_data) = parse_liquid_bip21(input) {
+        // Special case of LN BOLT11 with onchain fallback
+        // Search for the `lightning=bolt11` param in the BIP21 URI and, if found, extract the bolt11
+        let mut invoice_param: Option<LNInvoice> = None;
+        if let Some(query) = input.split('?').collect::<Vec<_>>().get(1) {
```
erdemyerebasmaz, line 183: NIT:
I believe we can depend on string slicing here instead of collecting to a vector to avoid unnecessary allocations.
```rust
        if let Some(query) = input.find('?') {
            let invoice_params = &input[query + 1..];
            ...
        }
```

We can also return early instead of matching `invoice_param` later on.

Performance gains are may appear negligible but `parse` is one of the highest usage rate API's we have, both internally and externally, so even the smallest optimization adds up.

erdemyerebasmaz (commented):
Changes look good!

Have two NIT comments for performance & readability.

libs/sdk-common/src/input_parser.rs
```diff
@@ -228,6 +249,68 @@ fn prepend_if_missing(prefix: &str, input: &str) -> String {
...
+        });
+    }
+
+    Ok(AddressData {
+        address,
+        network,
+        amount_sat: None,
+        label: None,
+        message: None,
+    })
```
erdemyerebasmaz, line 312: NIT: One suggestion for readability.

We can define URI variables & assign their values directly after parsing, without the need for a `HashMap`.
```rust
    let mut amount_sat = None;
    let mut label = None;
    let mut message = None;
    
    if let Some(params) = parts.next() {
        for pair in params.split('&') {
            if let Some((key, val)) = pair.split_once('=') {
                match key {
                    "amount" => amount_sat = Some((val.parse::<f64>()? * 100_000_000.0) as u64),
                    "label" => label = Some(val.to_string()),
                    "message" => message = Some(val.to_string()),
                    _ => {}
                }
            }
        }
    }

    Ok(AddressData {
        address,
        network,
        amount_sat,
        label,
        message,
    })
```

One other improvement on earlier statements would be to use `ok_or`/`ok_or_else` & return error messages early, instead of pattern matching.

libs/sdk-common/src/input_parser.rs
```diff
@@ -175,6 +175,27 @@ pub async fn parse(input: &str) -> Result<InputType> {
...
+        let mut invoice_param: Option<LNInvoice> = None;
+        if let Some(query) = input.split('?').collect::<Vec<_>>().get(1) {
+            invoice_param = querystring::querify(query)
+                .iter()
+                .find(|(key, _)| key == &"lightning")
+                .map(|(_, value)| parse_invoice(value))
+                .transpose()?;
+        }
+
+        return match invoice_param {
```
roeierez, line 191: This is different than bip21 for the greenlight sdk. Here we truly want to return the parsed bip21 (both invoice and address). Perhaps also in the greenlight sdk we should have done it this way.
I am thinking of the other info that the sdk might use (label, etc...). Also the input parser is a generic parser that can be used by the user regardless of payments.
How about returning here LiquidBip21 structure?

libs/sdk-common/src/input_parser.rs
```diff
@@ -388,7 +471,15 @@ pub enum InputType {
...
     BitcoinAddress {
-        address: BitcoinAddressData,
+        address: AddressData,
+    },
+
+    /// # Supported standards
+    ///
+    /// - plain on-chain liquid address
+    /// - BIP21 on liquid/liquidtestnet
+    LiquidAddress {
```
roeierez, line 481: This will propagate to greenlight sdk which I am not sure we want...

libs/sdk-common/src/input_parser.rs
hydra-yse: Will do! I left the invoice as plain string, though we do parse it in the bip21 parser. Should we return `LNInvoice` struct directly instead?

libs/sdk-common/src/input_parser.rs
hydra-yse: Thanks for the suggestion Erdem, that looks very elegant!

hydra-yse:
@erdemyerebasmaz I read your comment about allocations. Currently the deserializer performs a lot of that as I avoided using lifetime specifiers in the trait implementation like the `bip21` crate does. Do you think it's worth adding it in this PR so we avoid re-allocating strings for every field in `LiquidURI`, or should it be addressed in the future?

libs/sdk-common/src/liquid/bip21.rs
```diff
@@ -0,0 +1,174 @@
...
+use std::{num::ParseFloatError, str::FromStr, string::FromUtf8Error};
+use urlencoding::decode;
+
+use crate::{
+    invoice::{parse_invoice, InvoiceError},
+    prelude::Network,
+};
+
+#[derive(Debug, Clone, Serialize)]
+pub struct LiquidAddressData {
```
roeierez, line 16: NIT: LiquidBit21 ? as it is not only an address.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -724,6 +734,7 @@ dictionary RecommendedFees {
 [Enum]
 interface InputType {
     BitcoinAddress(BitcoinAddressData address);
+    LiquidAddress(LiquidAddressData address);
```
roeierez, line 737: Still propagating to the sdk here. since we build the greenlight sdk I think we can remove it from here.

roeierez (changes requested):
Looks good, some final NIT comments and we are good to go.

libs/sdk-common/src/liquid/bip21.rs
roeierez: I see now that you are using it for plain liquid address also so perhaps that's fine. @hydra-yse ?

libs/sdk-common/Cargo.toml
```diff
@@ -29,6 +29,8 @@ tonic = { workspace = true, features = [
     "tls-webpki-roots",
 ] }
 url = "2.5.0"
+elements = "0.24.1"
```
roeierez, line 32: Better to make the elements dependency as optional under feature as well and list that feature as part of  "liquid" features list.

libs/sdk-common/Cargo.toml
```diff
@@ -37,4 +39,7 @@ tokio = { workspace = true }
 once_cell = { workspace = true }
 
 [build-dependencies]
-tonic-build = { workspace = true }
\ No newline at end of file
+tonic-build = { workspace = true }
+
+[features]
```
roeierez, line 44: By default all features are enabled so I suggest to change that to not include liquid. 
e.g:
`default = []`

roeierez (changes requested):

libs/sdk-common/Cargo.toml
erdemyerebasmaz: > By default all features are enabled so I suggest to change that to not include liquid. e.g: `default = []`

I don't know if it has changed but it's the opposite, [all features are disabled unless explicitly enabled](https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature).

roeierez: Correct, my mistake.

libs/sdk-common/src/input_parser.rs
roeierez: I think plain string is just fine. User an parse it if desired.

libs/sdk-common/Cargo.toml
erdemyerebasmaz: Done.

Build elements crate only as part of liquid feature 7debe215449fbca9eaa294cf3f969391c5c9897c

roeierez (approved):
LGTM

libs/sdk-common/Cargo.toml
erdemyerebasmaz: I'm going through the PR and I've also added `urlencoding` to the optional dependency list.

- Build urlencoding crate only as part of liquid feature 72ca6efb1650fa408dbb9b5f6db9b16664943047

erdemyerebasmaz (approved):
LGTM!

Notes for @hydra-yse:
- **Binding issue**: While it's possible to mirror structs based on config feature, `sdk-core` & `sdk-bindings` is only interested in non-liquid feature structs. So, I've only mirrored the non-liquid variants. 6f125a75f918913551d3fed07c17e636e4c67e6b
  - Removed `liquid` feature from `sdk-bindings` e5bd06b84099f5ed6350f851d3017c1fe77fd173
- Marked `elements` & `urlencoding` as optional crates, to be only built with liquid feature. 7debe215449fbca9eaa294cf3f969391c5c9897c 72ca6efb1650fa408dbb9b5f6db9b16664943047
- Added several syntactic refactor changes for readability.
  - Extract liquid address parsing into a method 777d45e84d039b1df86f95b2e2033c735460938e
    - In case it's expanded & it's error handling changes in the future
  - refactor: Use ok_or/ok_or_else for error handling 6dd4eb404bdbf1feab3c3f1fd3429f70a71e8cc1
  - refactor: Use pattern matching on elements_address.params to set network 5e8c0341e2e2270442e2e53a067e41ecdf1e1618

erdemyerebasmaz:
@roeierez  Hold off merging as my last refactoring change has errors.

- refactor: Use pattern matching on elements_address.params to set network 5e8c0341e2e2270442e2e53a067e41ecdf1e1618

erdemyerebasmaz:
> @roeierez Hold off merging as my last refactoring change has errors.
> 
> * refactor: Use pattern matching on elements_address.params to set network [5e8c034](https://github.com/breez/breez-sdk/commit/5e8c0341e2e2270442e2e53a067e41ecdf1e1618)

@roeierez Removed these changes as `AddressData` cannot be used in pattern matching. One of it's fields `bech32::Hrp` does not derive `PartialEq`.

ok300 (approved):

libs/sdk-common/src/input_parser.rs
```diff
@@ -577,6 +595,52 @@ pub struct BitcoinAddressData {
...
+            "".to_string()
+        } else {
+            optional_keys
+                .iter()
+                .map(|(key, value)| format!("{key}={value}"))
+                .collect::<Vec<String>>()
+                .join("&")
+        };
+
+        Ok(format!("{scheme}:{}{suffix_str}", self.address))
```
roeierez, line 640: In case there are no options shouldn't we return just the address (no schema)? Otherwise if we parse a pure address and then serialize it it will be serialized as bip21 (with bitcoin schema).

PR #1060, opened 2024-08-05
Return insufficient funds error when needed

### LINKED ISSUE #1059: Return a specific error when onchain funds are not enough to pay for onchain fees
At the moment, `prepare_redeem_onchain_funds` returns `NodeError::generic("Insufficient funds to pay fees")` when funds are not enough to pay for onchain fees. It would be useful if a specific error (e.g. `NodeError:InsufficientFundsForFees`) was returned so that SDK integrators could react without having to parse the error string.

roeierez:
fixed #1059

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -102,7 +102,7 @@ enum SendPaymentError {
 enum RedeemOnchainError {    
     "Generic",    
     "ServiceConnectivity",
-    "InsufficientFunds"
+    "InsufficientFunds",
 };
 
 enum EnvironmentType {
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -98,6 +98,13 @@ enum SendPaymentError {
     "ServiceConnectivity",
 };
 
+[Error]
+enum RedeemOnchainError {    
+    "Generic",    
+    "ServiceConnectivity",
+    "InsufficientFunds"
+};
+
 enum EnvironmentType {
     "Production",
     "Staging",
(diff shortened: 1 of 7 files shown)


libs/sdk-core/src/error.rs
```diff
@@ -201,6 +201,52 @@ impl From<PersistError> for ReceiveOnchainError {
...
+    /// This error is raised when a general error occurs not specific to other error variants
+    /// in this enum.
+    #[error("Generic: {err}")]
+    Generic { err: String },
+
+    /// This error is raised when a connection to an external service fails.
+    #[error("Service connectivity: {err}")]
+    ServiceConnectivity { err: String },
+
+    // This error is raised when the node does not have enough funds to redeem the onchain balance.
```
dangeross, line 217:
```suggestion
    /// This error is raised when the node does not have enough funds to redeem the onchain balance.
```

libs/sdk-core/src/error.rs
```diff
@@ -201,6 +201,52 @@ impl From<PersistError> for ReceiveOnchainError {
...
+    #[error("{err}")]
+    InsufficientFunds { err: String },
+}
+
+impl From<NodeError> for RedeemOnchainError {
+    fn from(value: NodeError) -> Self {
+        match value {
+            NodeError::InsufficientFunds(err) => Self::InsufficientFunds { err },
```
dangeross, line 225: Could convert ServiceConnectivity here also
```suggestion
            NodeError::InsufficientFunds(err) => Self::InsufficientFunds { err },
            NodeError::ServiceConnectivity(err) => Self::ServiceConnectivity { err },
```

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -98,6 +98,13 @@ enum SendPaymentError {
     "ServiceConnectivity",
 };
 
+[Error]
+enum RedeemOnchainError {    
+    "Generic",    
+    "ServiceConnectivity",
+    "InsufficientFunds"
```
dangeross, line 105:
```suggestion
    "InsufficientFunds",
```

dangeross (approved):

ok300 (approved):

PR #1061, opened 2024-08-05
Validate invoice expiration

### LINKED ISSUE #1057: sdk: Add pre-flight checks for payments
The Greenlight plugin implements a number of pre-flight checks:

 - Check that the sum of all available channel balances exceeds the amount to be sent, if not abort.
 - Wait for enough available channel balances to come online before calling `pay`. On a 30s timeout abort with an error
 - Check the invoice expiry. If the invoice is expired, abort.

All of these are considered to be user-errors (well, the slow channel reconnect could be indicative of several problems, so those trigger introspections from us), and as such they could and likely should be handled on the client side already.

Since the SDK has a very good idea of the channel state, thanks to the background sync of node metadata, the first and last check would be trivial to implement, and would result in an even faster feedback for users.

roeierez:
Partially fixes #1057 
Checking the balance is a bit tricky since the local cache might be outdated so we will have to query the node and if that's the case then it is better to leave it for the gl-plugin saving the round trip IMO.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -261,6 +261,12 @@ impl BreezServices {
     ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
+        let invoice_expiration = parsed_invoice.timestamp + parsed_invoice.expiry;
+        if invoice_expiration < SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs() {
+            return Err(SendPaymentError::InvoiceExpired {
+                err: format!("Invoice expired at {}", parsed_invoice.expiry).into(),
+            });
+        }
         let invoice_amount_msat = parsed_invoice.amount_msat.unwrap_or_default();
         let provided_amount_msat = req.amount_msat.unwrap_or_default();
 
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -262,7 +262,8 @@ impl BreezServices {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
         let invoice_expiration = parsed_invoice.timestamp + parsed_invoice.expiry;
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -261,6 +261,12 @@ impl BreezServices {
     ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
+        let invoice_expiration = parsed_invoice.timestamp + parsed_invoice.expiry;
+        if invoice_expiration < SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs() {
+            return Err(SendPaymentError::InvoiceExpired {
+                err: format!("Invoice expired at {}", parsed_invoice.expiry),
```
dangeross, line 267: Do you want to show the expiry timestamp here?
```suggestion
                err: format!("Invoice expired at {}", invoice_expiration),
```

dangeross (approved):

libs/sdk-core/src/breez_services.rs
```diff
@@ -261,6 +261,12 @@ impl BreezServices {
     ) -> Result<SendPaymentResponse, SendPaymentError> {
         self.start_node().await?;
         let parsed_invoice = parse_invoice(req.bolt11.as_str())?;
+        let invoice_expiration = parsed_invoice.timestamp + parsed_invoice.expiry;
+        if invoice_expiration < SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs() {
+            return Err(SendPaymentError::InvoiceExpired {
+                err: format!("Invoice expired at {}", invoice_expiration),
+            });
+        }
```
erdemyerebasmaz, line 269: One suggestion for readability.
```suggestion
        let invoice_expiration = parsed_invoice.timestamp + parsed_invoice.expiry;
        let current_time = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
        if invoice_expiration < current_time {
            return Err(SendPaymentError::InvoiceExpired {
                err: format!("Invoice expired at {}", invoice_expiration),
            });
        }
```

erdemyerebasmaz (approved):
LGTM

I have one readability suggestion.

erdemyerebasmaz (approved):

Issue #1018, opened 2024-06-20
`inbound_liquidity_msats` is confusing


JssDWt:
The `inbound_liquidity_msats` in `NodeState` currently shows the largest receivable amount on a single channel. For integrations like the btcpay server integration that's confusing when people want to maximize their inbound liquidity. `inbound_liquidity_msats` should probably contain the _total_ receivable amount on all your channels. 

We have to make sure that internally we keep using the max receivable amount on a single channel for invoices though, and also have a field that clients can use to calculate fees.

ok300:
Fixed in #1062

PR #1062, opened 2024-08-05
Fix `inbound_liquidity_msats`

danielgranhao:
Addresses #1018

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -188,7 +188,7 @@ dictionary NodeState {
     u64 max_chan_reserve_msats;
     sequence<string> connected_peers;
     u64 max_receivable_single_payment_amount_msat;
-    u64 inbound_liquidity_msats;
+    u64 total_inbound_liquidity_msats;
 };
 
 dictionary ConfigureNodeRequest {
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -187,6 +187,7 @@ dictionary NodeState {
     u64 max_single_payment_amount_msat;
     u64 max_chan_reserve_msats;
     sequence<string> connected_peers;
+    u64 max_receivable_single_payment_amount_msat;
     u64 inbound_liquidity_msats;
 };
 
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -634,7 +634,7 @@ pub struct NodeState {
     pub max_receivable_single_payment_amount_msat: u64,
 
     /// Total receivable on all available channels
-    pub inbound_liquidity_msats: u64,
+    pub total_inbound_liquidity_msats: u64,
 }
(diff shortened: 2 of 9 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -629,6 +629,7 @@ pub struct NodeState {
     pub max_single_payment_amount_msat: u64,
     pub max_chan_reserve_msats: u64,
     pub connected_peers: Vec<String>,
+    pub max_receivable_single_payment_amount_msat: u64,
```
ok300, line 632: I would add some rustdocs about the meaning of the changed `NodeState` fields, for example

```rust
pub struct NodeState {
    ...

    /// Maximum receivable in a single payment (single channel)
    pub max_receivable_single_payment_amount_msat: u64,

    /// Total receivable on all available channels
    pub inbound_liquidity_msats: u64,
}
```

ok300 (approved):
Looks good, added just a NIT on comments.

libs/sdk-core/src/models.rs
danielgranhao: Added the proposed docs in [e2d48a4](https://github.com/breez/breez-sdk/pull/1062/commits/e2d48a4f6a2f640877ea4b10c24fce1d4c8b019a)

libs/sdk-core/src/models.rs
```diff
@@ -629,6 +629,11 @@ pub struct NodeState {
     pub max_single_payment_amount_msat: u64,
     pub max_chan_reserve_msats: u64,
     pub connected_peers: Vec<String>,
+
+    /// Maximum receivable in a single payment (single channel)
```
JssDWt, line 633: I know this line was suggested by @ok300, but a little nit here still.

```suggestion
    /// Maximum receivable in a single payment without requiring a new channel open.
```

JssDWt (commented):
I just have one nit, looking good!

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -187,6 +187,7 @@ dictionary NodeState {
     u64 max_single_payment_amount_msat;
     u64 max_chan_reserve_msats;
     sequence<string> connected_peers;
+    u64 max_receivable_single_payment_amount_msat;
```
roeierez, line 190: Now that we add this field the former one (inbound_liquidity_msats) changes its semantic. 
My concern is that developers will upgrade and their logic will break without them knowing about it.
Can we rename the `inbound_liquidity_msats` so developers will get a compilation error on upgrade which will force them to change the code?
We can also leave it with the old semantic (and mark it as deprecated) and add another field (total_inbound_liquidity_msat) with the unified liquidity.
Anything that will not let this change go under the radar.

danielgranhao: That's a very good point @roeierez. I'll rename it to `total_inbound_liquidity_msat`.

roeierez (approved):

JssDWt (approved):
LGTM!

PR #1063, opened 2024-08-06
Diagnostic data: sanitize swap and reverse swap infos

ok300:
This PR ensures our diagnostic data doesn't include sensitive fields for swaps and reverse swaps.

### DIFF
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -1584,6 +1584,127 @@ impl PaymentPathEdge {
     }
 }
 
+pub(crate) mod sanitize {
+    use anyhow::Result;
+    use serde::Serialize;
+
+    use crate::{
+        FullReverseSwapInfo, OpeningFeeParams, ReverseSwapInfoCached, SwapInfo, SwapStatus,
+    };
+
+    pub(crate) trait Sanitize<R, S> {
+        /// Sanitizes a raw struct [R] to a sanitized version of it [S]
+        fn sanitize(self) -> S;
+    }
+
+    pub(crate) fn sanitize_vec<R, S>(vals: Vec<R>) -> Vec<S>
+    where
+        R: Sanitize<R, S>,
+    {
+        vals.into_iter()
+            .map(|val| val.sanitize())
+            .collect::<Vec<S>>()
+    }
+
+    pub(crate) fn sanitize_vec_pretty_print<R, S>(vals: Vec<R>) -> Result<String>
+    where
+        R: Sanitize<R, S>,
+        S: Serialize,
+    {
+        serde_json::to_string_pretty(&sanitize_vec(vals)).map_err(anyhow::Error::new)
+    }
+
+    /// Sanitized version of [FullReverseSwapInfo], without the following fields:
+    /// - `preimage`
+    /// - `private_key`
+    #[derive(Serialize)]
+    pub(crate) struct FullReverseSwapInfoSanitized {
+        pub id: String,
+        pub created_at_block_height: u32,
+        pub claim_pubkey: String,
+        pub timeout_block_height: u32,
+        pub invoice: String,
+        pub redeem_script: String,
+        pub onchain_amount_sat: u64,
+        pub sat_per_vbyte: Option<u32>,
+        pub receive_amount_sat: Option<u64>,
+        pub cache: ReverseSwapInfoCached,
+    }
+    impl Sanitize<FullReverseSwapInfo, FullReverseSwapInfoSanitized> for FullReverseSwapInfo {
+        fn sanitize(self) -> FullReverseSwapInfoSanitized {
+            FullReverseSwapInfoSanitized {
+                id: self.id,
+                created_at_block_height: self.created_at_block_height,
+                claim_pubkey: self.claim_pubkey,
+                timeout_block_height: self.timeout_block_height,
+                invoice: self.invoice,
+                redeem_script: self.redeem_script,
+                onchain_amount_sat: self.onchain_amount_sat,
+                sat_per_vbyte: self.sat_per_vbyte,
+                receive_amount_sat: self.receive_amount_sat,
+                cache: self.cache,
+            }
+        }
+    }
+
+    /// Sanitized version of [SwapInfo], without the following fields:
+    /// - `preimage`
+    /// - `private_key`
+    #[derive(Serialize)]
+    pub(crate) struct SwapInfoSanitized {
+        pub bitcoin_address: String,
+        pub created_at: i64,
+        pub lock_height: i64,
+        pub payment_hash: Vec<u8>,
+        pub public_key: Vec<u8>,
+        pub swapper_public_key: Vec<u8>,
+        pub script: Vec<u8>,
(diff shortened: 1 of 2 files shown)


libs/sdk-core/src/models.rs
```diff
@@ -1584,6 +1584,127 @@ impl PaymentPathEdge {
...
 
+pub(crate) mod sanitize {
+    use anyhow::Result;
+    use serde::Serialize;
+
+    use crate::{
+        FullReverseSwapInfo, OpeningFeeParams, ReverseSwapInfoCached, SwapInfo, SwapStatus,
+    };
+
+    pub(crate) trait Sanitize<R, S> {
```
roeierez, line 1595: Not sure if we need a trait here. I think we could save some code if we only implement the `From` trait for the FullReverseSwapInfoSanitized from FullReverseSwapInfo ?

roeierez (changes requested):

libs/sdk-core/src/models.rs
```diff
@@ -1584,6 +1584,127 @@ impl PaymentPathEdge {
...
+        S: Serialize,
+    {
+        serde_json::to_string_pretty(&sanitize_vec(vals)).map_err(anyhow::Error::new)
+    }
+
+    /// Sanitized version of [FullReverseSwapInfo], without the following fields:
+    /// - `preimage`
+    /// - `private_key`
+    #[derive(Serialize)]
+    pub(crate) struct FullReverseSwapInfoSanitized {
```
JssDWt, line 1621: You think it would be possible to do this conversion with a macro?

Have something like this on `FullReverseSwapInfo`:

```
#[secret]
pub preimage: Vec<u8>
```

And have the serializer ignore those 'secret' fields, or put `"redacted"` instead? 
That would help so we don't have to remember to update these structs when fields are added to them, and it would be easy to add this when other secrets are identified.

roeierez: If we are going this path then perhaps omitting from the json using `#[serde(skip_serializing)]` I don't think we are serializing to json anywhere else.

JssDWt: > If we are going this path then perhaps omitting from the json using `#[serde(skip_serializing)]` I don't think we are serializing to json anywhere else.

That would be fine too.

ok300: That's a good idea, but unfortunately we do serialize to json in `list_payments` > `LnPaymentDetails`, which has

```rust
    /// Only set for [PaymentType::Received] payments that were received in the context of a swap
    pub swap_info: Option<SwapInfo>,

    /// Only set for [PaymentType::Sent] payments that were sent in the context of a reverse swap
    pub reverse_swap_info: Option<ReverseSwapInfo>,
```

If we `#[serde(skip_serializing)]` any field of `SwapInfo`, serialization fails here.

JssDWt: @ok300 I don't understand what you mean, can you elaborate? Where does serialization fail?

ok300: It fails when trying to build the `LnPaymentDetails` as part of reconstructing the `Payment`:

https://github.com/breez/breez-sdk-greenlight/blob/a5657e49b6fa9bd9fa2da4dbdeff5af30cb20973/libs/sdk-core/src/persist/transactions.rs#L240-L250

The query has a `JOIN` on the swap table. From this data, it tries to reconstruct the `SwapInfo` present in the `Payment`'s `LnPaymentDetails`, which fails if some fields are marked as `#[serde(skip_serializing)`.

ok300:
@JssDWt @roeierez I simplified the PR based on the feedback above: c9dc87467aad22d73c121644a2ad5a19319e8d60

It now:
- allows for both sanitized serialization (used in `generate_sdk_diagnostic_data`) and preserves the default non-sanitized serialization (used elsewhere, like when reconstructing `SwapInfos` as part of `list_payments`)
- defining the sanitization logic doesn't need to be updated when new fields are added to the underlying struct (only the sanitized fields have to be explicitly stated)
- defining the sanitization logic is simple to do for new structs (just implement `Sanitize<T>`)

What do you think?

libs/sdk-core/src/models.rs
```diff
@@ -1584,15 +1584,67 @@ impl PaymentPathEdge {
     }
 }
 
+pub(crate) mod sanitize {
```
roeierez, line 1587: Not sure what is the benefit of a new module and a trait for sanitize. Wouldn't it be simpler if we just add a function to the structures, for example:

```
impl FullReverseSwapInfo {
 fn sanitize(self) -> FullReverseSwapInfo {
            FullReverseSwapInfo {
                preimage: vec![],
                private_key: vec![],
                ..self.clone()
            }
        }
}
```

And then just use it inline where needed?

ok300: The main benefit is it makes the calling code more readable:

```rust
let reverse_swaps = self
    .persister
    .list_reverse_swaps()
    .map(sanitize_vec_pretty_print)??;
```
vs
```rust
let reverse_swaps = serde_json::to_string_pretty(
    &self
        .persister
        .list_reverse_swaps()?
        .into_iter()
        .map(|rsi| rsi.sanitize())
        .collect::<Vec<FullReverseSwapInfo>>(),
)?;
```

I'm fine with either way, just prefer the first option because it's easier to read.

roeierez (approved):

libs/sdk-core/src/models.rs
```diff
@@ -1584,15 +1584,67 @@ impl PaymentPathEdge {
...
+    impl Sanitize<SwapInfo> for SwapInfo {
+        fn sanitize(self) -> SwapInfo {
+            SwapInfo {
+                preimage: vec![],
+                private_key: vec![],
+                ..self.clone()
+            }
+        }
+    }
+}
```
JssDWt, line 1633: A small suggestion to make the interface a little easier to implement

```suggestion
    pub(crate) trait Sanitize {
        /// Sanitizes a raw struct to a clone of itself where sensitive fields are blanked
        fn sanitize(self) -> Self;
    }

    pub(crate) fn sanitize_vec<T>(vals: Vec<T>) -> Vec<T>
    where
        T: Sanitize,
    {
        vals.into_iter()
            .map(|val| val.sanitize())
            .collect::<Vec<T>>()
    }

    pub(crate) fn sanitize_vec_pretty_print<T>(vals: Vec<T>) -> Result<String>
    where
        T: Sanitize + Serialize,
    {
        serde_json::to_string_pretty(&sanitize_vec(vals)).map_err(anyhow::Error::new)
    }

    impl Sanitize for FullReverseSwapInfo {
        fn sanitize(self) -> FullReverseSwapInfo {
            FullReverseSwapInfo {
                preimage: vec![],
                private_key: vec![],
                ..self.clone()
            }
        }
    }

    impl Sanitize for SwapInfo {
        fn sanitize(self) -> SwapInfo {
            SwapInfo {
                preimage: vec![],
                private_key: vec![],
                ..self.clone()
            }
        }
    }
}
```

JssDWt (commented):
Just a nit, otherwise LGTM

PR #1071, opened 2024-08-09
drop node state

JssDWt:
merge commit 74a820817bb44dacb1a1193264017877b77156c6 introduced commits c51d2125a9e476a72328d943a718c301921b49ef and
ee25097d55d97c0458c8ed973fadeb3bdfe0afc7, which modify the node state. When the node state is cached, however, deserialization now fails. This commit drops the current node state to ensure deserialization always works.


JssDWt:
Tagging @danielgranhao for review too, because I can't pick you from the list.

danielgranhao (approved):

dangeross (approved):
LGTM. I guess this should be added to a hotfix release

JssDWt:
> LGTM. I guess this should be added to a hotfix release

These commits were not in a release tag yet. So I don't think that's necessary.

PR #1076, opened 2024-08-20
Exclude `use_trampoline` field of `LnUrlPayRequest` on `liquid` feature of `sdk-common`

erdemyerebasmaz:
This PR excludes `use_trampoline` field of `LnUrlPayRequest` on `liquid` feature of `sdk-common`.

### Changelist:
- Exclude use_trampoline field of LnUrlPayRequest from liquid feature 12df2fed5ab63b23e983176fd4f0152d2fc069be
- chore: `flutter pub get` && `flutter pub upgrade` aaa54051c86babadf9cdb1ee001ed1b4fa8ccf41

### DIFF
--- a/libs/sdk-common/src/lnurl/specs/pay.rs
+++ b/libs/sdk-common/src/lnurl/specs/pay.rs
@@ -132,6 +132,7 @@ pub mod model {
         pub data: LnUrlPayRequestData,
         /// The amount in millisatoshis for this payment
         pub amount_msat: u64,
+        #[cfg(not(feature = "liquid"))]
         /// Trampoline payments outsource pathfinding to the LSP. Trampoline payments can improve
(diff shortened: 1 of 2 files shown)


libs/sdk-common/src/lnurl/specs/pay.rs
```diff
@@ -132,6 +132,7 @@ pub mod model {
         pub data: LnUrlPayRequestData,
         /// The amount in millisatoshis for this payment
         pub amount_msat: u64,
+        #[cfg(not(feature = "liquid"))]
```
ok300, line 135:
```suggestion
        #[cfg(not(feature = "liquid"))] // Only available for the Greenlight SDK
```

ok300 (approved):
One NIT, otherwise LGTM

PR #1080, opened 2024-08-28
Add file logger RN example

dangeross:
This PR adds to the React Native example additional code to log to file SDK log entries and events, which can then be shared via email. 

Added to the example are dependencies:
- [react-native-file-logger](https://github.com/BeTomorrow/react-native-file-logger)
- [react-navigation](https://reactnavigation.org)

To configure the file logger, we disable capturing console logs by default and set the max number of log files to 5 (5 days of logs). You can additionally set the `maximumFileSize` to limit the file size:
```javascript
FileLogger.configure({
    captureConsole: false,
    maximumNumberOfFiles: 5
})
```

We wrap the capturing of log entries to a utility class `Log` which can log both to console and optionally to file:
```javascript
import { Log } from "../utils/logging"

const log = new Log("ExampleApp")
log.debug(`Log a debug log entry to file`, true)
```

We update the SDK `logHandler` in the `HomeScreen` to use the `Log` class to capture logs:
```javascript
    const logHandler = (logEntry) => {
        switch (logEntry.level) {
            case "ERROR":
                log.error(`[ERROR]: ${logEntry.line}`, true)
                break
            case "TRACE":
                // Only log to file when in DEBUG mode
                log.debug(`[TRACE]: ${logEntry.line}`, DEBUG)
                break
            default:
                log.debug(`[${logEntry.level}]: ${logEntry.line}`, true)
                break
        }
    }
```

And finally add a button to trigger sending the log file via email:
```javascript
        try {
            await FileLogger.sendLogFilesByEmail({
                subject: "Logs",
                body: `OS: ${Platform.OS} - ${Platform.Version}\n\n` + JSON.stringify(nodeInfo, null, 2)
            })
        } catch (e) {
            log.error(JSON.stringify(e))
        }
```


roeierez (approved):
LGTM

PR #1082, opened 2024-09-09
Change short_channel_id to string

dangeross:
This PR changes the `short_channel_id` of the `RouteHintHop` model to string in the format:
```
<BLOCK_HEIGHT>x<TX_INDEX>x<OUTPUT_INDEX>
```

### DIFF
--- a/libs/sdk-common/src/invoice.rs
+++ b/libs/sdk-common/src/invoice.rs
@@ -85,7 +85,7 @@ impl From<SystemTimeError> for InvoiceError {
 fn parse_short_channel_id(id_str: &str) -> InvoiceResult<u64> {
     let parts: Vec<&str> = id_str.split('x').collect();
     if parts.len() != 3 {
-        return Ok(0);
+        return Err(InvoiceError::generic("Invalid short channel id"));
     }
     let block_num = parts[0].parse::<u64>()?;
     let tx_num = parts[1].parse::<u64>()?;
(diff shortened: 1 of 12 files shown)


libs/sdk-core/src/swap_out/boltzswap.rs
```diff
@@ -61,7 +61,7 @@ impl From<BoltzRouteHintHop> for RouteHintHop {
     fn from(value: BoltzRouteHintHop) -> Self {
         RouteHintHop {
             src_node_id: value.node_id,
-            short_channel_id: 0_u64,
+            short_channel_id: "0".to_string(),
```
JssDWt, line 64:
```suggestion
            short_channel_id: "0x0x0".to_string(),
```

libs/sdk-common/src/invoice.rs
```diff
@@ -75,6 +82,25 @@ impl From<SystemTimeError> for InvoiceError {
     }
 }
 
+fn parse_short_channel_id(id_str: &str) -> InvoiceResult<u64> {
+    let parts: Vec<&str> = id_str.split('x').collect();
+    if parts.len() != 3 {
+        return Ok(0);
```
JssDWt, line 88: I know this was already there, but shouldn't this return an error?

dangeross: Changed to throw an error, also added some tests

roeierez (approved):

JssDWt (approved):
LGTM

PR #1088, opened 2024-09-17
Implement lnurl auth signer

roeierez:
This PR changes the lnurl auth functionality to support signers that don't expose private keys.
Instead of passing the private key (linking key) to the lnurl module for signing the lnurl module expect a signer implementation with some basic building blocks for siginig and deriving public keys.
Also the caller now doesn't need to have a knowledge about the lnurl auth internals (constructing the linking key for example) as the lnurl module takes care of it.
This is needed in order also to support splitting the signer from the sdk implementation in breez-ask-liquid project.

### DIFF
--- a/libs/sdk-common/src/lnurl/specs/auth.rs
+++ b/libs/sdk-common/src/lnurl/specs/auth.rs
@@ -6,7 +6,7 @@ use reqwest::Url;
 
 use crate::prelude::*;
 
-pub trait LnurAuthSigner {
+pub trait LnurlAuthSigner {
     fn derive_bip32_pub_key(&self, derivation_path: &[ChildNumber]) -> LnUrlResult<Vec<u8>>;
     fn sign_ecdsa(&self, msg: &[u8], derivation_path: &[ChildNumber]) -> LnUrlResult<Vec<u8>>;
     fn hmac_sha256(
@@ -23,7 +23,7 @@ pub trait LnurAuthSigner {
 /// https://github.com/lnurl/luds/blob/luds/05.md
 ///
 /// See the [parse] docs for more detail on the full workflow.
-pub async fn perform_lnurl_auth<S: LnurAuthSigner>(
+pub async fn perform_lnurl_auth<S: LnurlAuthSigner>(
     req_data: &LnUrlAuthRequestData,
     signer: &S,
 ) -> LnUrlResult<LnUrlCallbackStatus> {
@@ -92,7 +92,7 @@ pub fn validate_request(
     })
 }
 
-pub fn get_derivation_path<S: LnurAuthSigner>(
+pub fn get_derivation_path<S: LnurlAuthSigner>(
     signer: &S,
     url: Url,
 ) -> LnUrlResult<Vec<ChildNumber>> {
--- a/libs/sdk-common/src/lnurl/specs/auth.rs
+++ b/libs/sdk-common/src/lnurl/specs/auth.rs
@@ -29,7 +29,11 @@ pub async fn perform_lnurl_auth<S: LnurAuthSigner>(
 ) -> LnUrlResult<LnUrlCallbackStatus> {
     let url = Url::from_str(&req_data.url).map_err(|e| LnUrlError::InvalidUri(e.to_string()))?;
     let derivation_path = get_derivation_path(signer, url)?;
-    let sig = signer.sign_ecdsa(req_data.k1.as_bytes(), &derivation_path)?;
+    let sig = signer.sign_ecdsa(
+        &hex::decode(&req_data.k1)
+            .map_err(|e| LnUrlError::Generic(format!("Error decoding k1: {e}")))?,
+        &derivation_path,
+    )?;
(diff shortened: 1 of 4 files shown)


libs/sdk-common/src/lnurl/specs/auth.rs
```diff
@@ -1,39 +1,46 @@
...
+    signer: &S,
 ) -> LnUrlResult<LnUrlCallbackStatus> {
-    let k1_to_sign = Message::from_slice(
-        &hex::decode(req_data.k1)
-            .map_err(|e| LnUrlError::Generic(format!("Error decoding k1: {e}")))?,
-    )?;
-    let sig = Secp256k1::new().sign_ecdsa(&k1_to_sign, &linking_keys.secret_key());
+    let url = Url::from_str(&req_data.url).map_err(|e| LnUrlError::InvalidUri(e.to_string()))?;
+    let derivation_path = get_derivation_path(signer, url)?;
+    let sig = signer.sign_ecdsa(req_data.k1.as_bytes(), &derivation_path)?;
```
dangeross, line 32: I think this `k1` needs to be decoded from hex

libs/sdk-common/src/lnurl/specs/auth.rs
```diff
@@ -1,39 +1,46 @@
...
+    let xpub = signer.derive_bip32_pub_key(&derivation_path)?;
 
     // <LNURL_hostname_and_path>?<LNURL_existing_query_parameters>&sig=<hex(sign(utf8ToBytes(k1), linkingPrivKey))>&key=<hex(linkingKey)>
```
dangeross, line 35: From [LUD-04](https://github.com/lnurl/luds/blob/luds/04.md):
```suggestion
    // <LNURL_hostname_and_path>?<LNURL_existing_query_parameters>&sig=<hex(sign(hexToBytes(k1), linkingPrivKey))>&key=<hex(linkingKey)>
```

libs/sdk-core/src/lnurl/auth.rs
```diff
@@ -0,0 +1,58 @@
...
+            .derive_bip32_key(derivation_path.to_vec())?
+            .to_keypair(&Secp256k1::new())
+            .public_key()
+            .encode())
+    }
+
+    fn sign_ecdsa(&self, msg: &[u8], derivation_path: &[ChildNumber]) -> LnUrlResult<Vec<u8>> {
+        let xpriv = self.node_api.derive_bip32_key(derivation_path.to_vec())?;
+        let sig = Secp256k1::new().sign_ecdsa(
+            &Message::from_slice(msg).map_err(|_| LnUrlError::generic("failed to sign"))?,
```
dangeross, line 38:
```suggestion
            &Message::from_slice(msg).map_err(|_| LnUrlError::generic("Failed to sign"))?,
```

libs/sdk-core/src/lnurl/auth.rs
```diff
@@ -0,0 +1,58 @@
...
+        secp256k1::{Message, Secp256k1},
+        util::bip32::ChildNumber,
+    },
+    lightning::util::ser::Writeable,
+};
+use sdk_common::prelude::{LnUrlError, LnUrlResult, LnurAuthSigner};
+
+use crate::node_api::NodeAPI;
+
+pub(crate) struct SDKLnurlAuthSigner {
```
dangeross, line 15: Small nit about naming, `Sdk` is used throughout for structs and enums (correct or not, just for consistency)

libs/sdk-common/src/lnurl/specs/auth.rs
roeierez: Indeed, I wonder how it is working also with this code...

roeierez: Done

libs/sdk-core/src/lnurl/auth.rs
roeierez: Done

libs/sdk-core/src/lnurl/auth.rs
roeierez: Done

libs/sdk-common/src/lnurl/specs/auth.rs
```diff
@@ -1,39 +1,50 @@
...
-use bitcoin::secp256k1::{Message, Secp256k1};
-use bitcoin::util::bip32::{ChildNumber, ExtendedPrivKey};
-use bitcoin::KeyPair;
+use bitcoin::hashes::hex::ToHex;
+use bitcoin::util::bip32::ChildNumber;
 use reqwest::Url;
 
 use crate::prelude::*;
 
+pub trait LnurAuthSigner {
```
dangeross, line 9: Sorry, just spotted:
```suggestion
pub trait LnurlAuthSigner {
```

roeierez: Thank you. Done.

ok300 (approved):
It worked before and after the last 2 commits, but it's more readable now. Tested on stacker news.

PR #1091, opened 2024-09-28
json diagnostic

JssDWt:
Serialize the diagnostic data into a single json document.

In this PR also an attempt to serialize `Vec<u8>` as hex. It works, but became a bit of a giant cumbersome project. See the commit message for more information.

### DIFF
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -647,6 +647,7 @@ dependencies = [
  "ripemd",
  "rusqlite",
  "rusqlite_migration",
+ "ryu",
  "sdk-common",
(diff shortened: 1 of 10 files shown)


ok300 (approved):
Until someone finds a better way, this is IMO good enough.

It's very useful to have byte arrays pretty-printed as hex.

roeierez (approved):
As a temporary solution I think we can go with it. We can also open an issue in cln grpc to add an attribute for the binary vectors to be serialized as hex and then get rid of the custom serialization.

dangeross (approved):

PR #1093, opened 2024-10-03
greenlight: wait for channel reestablished

JssDWt:
The calls to `send_pay` are not covered by greenlight's awaitable channel futures. Therefore wait for channels to be reestablished before doing any calls to `send_pay`.

Alternative to https://github.com/Blockstream/greenlight/issues/522


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -684,6 +684,65 @@ impl Greenlight {
         _ = self.inprogress_payments.fetch_sub(1, Ordering::Relaxed);
         res
     }
+
+    async fn wait_channel_reestablished(&self, path: &PaymentPath) -> NodeResult<()> {
+        let deadline = match Instant::now().checked_add(Duration::from_secs(20)) {
+            Some(deadline) => deadline,
+            None => return Err(NodeError::generic("failed to add 20 seconds to now")),
```
dangeross, line 691: Bit more informative
```suggestion
            None => return Err(NodeError::generic("Failed to set channel establishment deadline")),
```

dangeross (commented):
Looking good

small nit, can you capitalise the error strings?

JssDWt:
Capitalized error strings and use `ok_or` rather than a verbose `match` statement.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -684,6 +684,67 @@ impl Greenlight {
...
+                .checked_add(Duration::from_secs(20))
+                .ok_or(NodeError::generic(
+                    "Failed to set channel establishment deadline",
+                ))?;
+
+        while Instant::now().le(&deadline) && !self.poll_channel_reestablished(path).await? {
+            tokio::time::sleep(Duration::from_millis(50)).await
+        }
+
+        Ok(())
```
dangeross, line 700: It looks like if we pass the deadline without channel reestablish the send_pay is still attempted, is that intended?

JssDWt: No, that was not intended, good catch!
However, now that you mention it, it might actually be fine to attempt that payment here.

dangeross: I don't have a problem with that

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -960,7 +1021,7 @@ impl NodeAPI for Greenlight {
                 "send_pay route to pay: {:?}, received_amount = {}",
                 route, to_pay_msat
             );
-
+            self.wait_channel_reestablished(&max.path).await?;
```
roeierez, line 1024: As far as I know greenlight has a mechanism to wait for channel re-establishment before attempting to pay, what is the benefit of adding such mechanism on the client too?

roeierez: Or perhaps this mechanism doesn't exist in `send_pay`  and that's why it is added here?

JssDWt: > Or perhaps this mechanism doesn't exist in `send_pay` and that's why it is added here?

exactly: https://github.com/Blockstream/greenlight/issues/522

roeierez:
@JssDWt I talked to Christian today and:
1. You are right the mechanism is not there.
2. The get_route that we execute before doesn't care about channel activeness so your analysis is correct.

The one thing he noted is that peer_connected is not enough for channel re-establishment but we need to check the channel state. I believe this is the check in greenlight: https://github.com/Blockstream/greenlight/blob/main/libs/gl-plugin/src/awaitables.rs#L177
@cdecker can you verify that?

JssDWt:
> The one thing he noted is that peer_connected is not enough for channel re-establishment but we need to check the channel state. I believe this is the check in greenlight: https://github.com/Blockstream/greenlight/blob/main/libs/gl-plugin/src/awaitables.rs#L177

I included that check in this PR: https://github.com/breez/breez-sdk-greenlight/pull/1093/files#diff-7466a53546c6cded460ef5502a9168b13cc930d94fdf7d0ea8166bb10c75fe7eR741

roeierez:
> > The one thing he noted is that peer_connected is not enough for channel re-establishment but we need to check the channel state. I believe this is the check in greenlight: https://github.com/Blockstream/greenlight/blob/main/libs/gl-plugin/src/awaitables.rs#L177
> 
> I included that check in this PR: https://github.com/breez/breez-sdk-greenlight/pull/1093/files#diff-7466a53546c6cded460ef5502a9168b13cc930d94fdf7d0ea8166bb10c75fe7eR741

Oh perfect. How did I miss that?

roeierez (approved):
LGTM

dangeross (approved):

PR #1100, opened 2024-10-08
Bump gl-client to latest released `v0.3.0`

### LINKED ISSUE #1090: ReceivePaymentException between network changes
When the SDK moves between networks (different sims or wifi <> sim), the ability receivePayment is borked.  The timeout takes over 10 minutes to respond meaning the app needs to be killed and restarted to gain back and network activity.
The received error after this time is:
```
java.lang.Exception: breez_sdk.ReceivePaymentException$Generic: Generic: status: Internal, message: "converting invoice response to grpc: empty result from RPC", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Thu, 26 Sep 2024 17:54:11 GMT", "content-length": "0"} }, source: None
```

Suspect that this is due to the reuse of grpc for performance? Does grpc need to be connection aware or bindings accessible so it can be restarted at will?

ok300:
This PR bumps `gl-client` to the latest publicly released[^2] `v0.3.0`.

According to the changelog[^1] this brings the fix for #1090 among others.

Fixes #1090

[^1]: https://github.com/Blockstream/greenlight/blob/main/CHANGELOG.md
[^2]: https://crates.io/crates/gl-client/versions

### DIFF
--- a/libs/Cargo.lock
+++ b/libs/Cargo.lock
@@ -1470,8 +1470,8 @@ checksum = "4271d37baee1b8c7e4b708028c57d816cf9d2434acb33a549475f78c181f6253"
 
 [[package]]
 name = "gl-client"
-version = "0.2.0"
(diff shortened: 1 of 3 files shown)


roeierez (approved):
LGTM

PR #1102, opened 2024-10-08
Optimize fetching claim tx by querying the lockup address

roeierez:
Instead of querying the claim address (final user address) for the claim tx it is better to query the lockup address.
Since the claim is not in our control I have seen a case where it had thousands of transactions resulted in 80m of transactions associated with that address.
The lockup address should have max 2 transactions associated with.

### DIFF
--- a/libs/sdk-core/src/swap_out/reverseswap.rs
+++ b/libs/sdk-core/src/swap_out/reverseswap.rs
@@ -533,13 +533,17 @@ impl BTCSendSwap {
         let lockup_addr = rsi.get_lockup_address(self.config.network)?;
         Ok(self
             .chain_service
-            .address_transactions(rsi.claim_pubkey.clone())
+            .address_transactions(lockup_addr.to_string())
             .await?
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/swap_out/reverseswap.rs
```diff
@@ -646,6 +650,12 @@ impl BTCSendSwap {
             let claim_tx = self.get_claim_tx(&rsi).await?;
             let claim_tx_status = TxStatus::from(&claim_tx);
 
+            if let Some(tx) = &claim_tx {
+                info!(
+                    "Cound claim tx for reverse swap {:?}: {:?}, status: {:?}",
```
dangeross, line 655:
```suggestion
                    "Found claim tx for reverse swap {:?}: {:?}, status: {:?}",
```

roeierez: fixed

ok300 (approved):

PR #1107, opened 2024-10-14
use invoice destination for trampoline

JssDWt:
For trampoline, the payment destination is the LSP.
Use the pubkey in the invoice as destination if an invoice is available.


erdemyerebasmaz (commented):
LGTM

PR #1109, opened 2024-10-14
Remove unnecessary get_info calls

roeierez:
We don't need anymore to explicitly start the node (call get_info) as geenlight reverse proxy does it automatically. 
Instead of getting the pubkey from get_info, we get it from the signer locally. 
This saves 200-300ms on my desktop (potentially more on mobile) from some sdk functions including sync.


dangeross (approved):

JssDWt (approved):
utACK!

PR #1115, opened 2024-10-29
refresh after hibernation

JssDWt:
As an alternative to #1112, reconnect only the services using persistent grpc channels or keepalives after hibernation.


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -170,24 +180,61 @@ impl Greenlight {
...
             gl_client: Mutex::new(None),
             node_client: Mutex::new(None),
             persister,
             inprogress_payments: AtomicU16::new(0),
+            internal_shutdown: Mutex::new(Arc::new(tx)),
+            external_shutdown: Mutex::new(None),
         })
     }
 
+    async fn do_start_signer(&self) {
```
roeierez, line 201: Previously the caller has started the signer and was responsible for running it in a task so the greenlight module doesn't need to handle any async tasks. 
I think we can simplify by keeping it as it was and on reconnect only replace the internal signer, then the caller (who calls reconnect) will have to run the signer again (which we are doing already by restarting the background task).
As a result I think we also won't need also the internal shutdown lock at this context so everything related to the hibernation is solely coded on the `reconnect` function.

libs/sdk-core/src/breez_services.rs
```diff
@@ -1510,11 +1517,49 @@ impl BreezServices {
...
+                        continue;
+                    }
+                };
+
+                if elapsed
+                    .saturating_sub(DETECT_HIBERNATE_SLEEP_DURATION)
+                    .ge(&DETECT_HIBERNATE_MAX_OFFSET)
+                {
+                    debug!("Hibernation detected, notifying services.");
+                    let _ = cloned.hibernation_sender.send(());
```
roeierez, line 1539: Perhaps we can here, just before the send, call reconnect on both breez server and geenlight?
It will save two tokio tasks that only handles these reconnections and also ensures these reconnects before the background tasks.

libs/sdk-common/src/lnurl/specs/auth.rs
```diff
@@ -6,10 +6,12 @@ use reqwest::Url;
...
 
+#[tonic::async_trait]
 pub trait LnurlAuthSigner {
-    fn derive_bip32_pub_key(&self, derivation_path: &[ChildNumber]) -> LnUrlResult<Vec<u8>>;
-    fn sign_ecdsa(&self, msg: &[u8], derivation_path: &[ChildNumber]) -> LnUrlResult<Vec<u8>>;
-    fn hmac_sha256(
+    async fn derive_bip32_pub_key(&self, derivation_path: &[ChildNumber]) -> LnUrlResult<Vec<u8>>;
+    async fn sign_ecdsa(&self, msg: &[u8], derivation_path: &[ChildNumber])
+        -> LnUrlResult<Vec<u8>>;
+    async fn hmac_sha256(
```
roeierez, line 14: If we make these async we will have to follow up with a change in breez_sdk_liquid that also uses this trait.

roeierez (changes requested):
Looks very good IMO. I dropped some comments that may simplify a bit.

libs/sdk-common/src/lnurl/specs/auth.rs
JssDWt: This comes from the mutex for the signer. Since these functions use the signer, and the signer is behind a async mutex, this function needs to be async too.

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: This is now implemented, let me know what you think.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -987,10 +996,37 @@ struct SyncState {
...
+            self.sdk_config.network.into(),
+            self.device.clone(),
+        ) {
+            Ok(new_signer) => new_signer,
+            Err(e) => {
+                error!(
+                    "Reconnect hibernation: failed to create new signer after reconnect request: {:?}",
+                    e
+                );
+                return;
```
dangeross, line 1019: It seems at this point if there's an error the previous signer loop is shutdown and we don't attempt to reconnect again (maybe it will if the hibernation reconnect is triggered again)? Is there a nice way to handle this?

JssDWt: I pushed some changes after review comments from @roeierez. Now the `reconnect` function no longer starts the signer. Starting the signer is done in a loop, if starting the signer exits, it will try to start again after a sleep. It still won't recreate the signer, but it could still be recreated after a new hibernation.

I'm not sure if there's a better way to handle that. Retrying to create the signer I don't think is an option. Keeping the old signer is probably the best option if recreating it fails.

roeierez (approved):
LGTM

PR #1120, opened 2024-11-08
Add list_swaps function

### LINKED ISSUE #1077: Discussion: Enhancing the "Swap" lifecycle
As of now the Breez SDK allows us to fetch only currently ongoing swaps. However in the past we experienced an issue with our users where an ongoing swap wasn't returned anymore from the SDK method (considering it as complete) but the associated LN payment to the user’s wallet either hasn’t occurred or has potentially failed. At this stage we are basically clueless as to what happened, eventually a user will be able to claim a refund but in the meanwhile there is just nothing for us to tell them.

To improve the user experience, we want to ensure that our users can always see the status of their swaps. After a swap is fully completed, it would also be beneficial to retain the swap data for display in the wallet history (e.g. via SDK method to get the list of past swaps). This data should include various timestamps throughout the swap’s lifecycle (e.g., swap requested, on-chain transaction confirmed, LN payment sent, LN payment received by the recipient).

Please share your thoughts in this thread.

JssDWt:
Adds a `list_swaps` function to the SDK, to query current and historical swaps. Swaps can be filtered by created timestamp and status. They can be paged with `offset` and `limit` fields.

Fixes #1077

### DIFF
--- a/libs/sdk-bindings/src/breez_sdk.udl
+++ b/libs/sdk-bindings/src/breez_sdk.udl
@@ -579,6 +579,14 @@ dictionary SwapInfo {
     u32? confirmed_at;
 };
 
+dictionary ListSwapsRequest {
+    sequence<SwapStatus>? status = null;
+    i64? from_timestamp = null;
+    i64? to_timestamp = null;
+    u32? offset = null;
+    u32? limit = null;
+};
+
 dictionary ReverseSwapPairInfo {
     u64 min;
     u64 max;
@@ -940,6 +948,9 @@ interface BlockingBreezServices {
 
    [Throws=SdkError]
    RefundResponse refund(RefundRequest req);
+
+   [Throws=SdkError]
+   sequence<SwapInfo> list_swaps(ListSwapsRequest req);
 
    [Throws=SdkError]
    ReverseSwapPairInfo fetch_reverse_swap_fees(ReverseSwapFeesRequest req);
--- a/libs/sdk-core/src/models.rs
+++ b/libs/sdk-core/src/models.rs
@@ -959,6 +959,17 @@ pub struct ReceiveOnchainRequest {
     pub opening_fee_params: Option<OpeningFeeParams>,
 }
 
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct ListSwapsRequest {
+    pub status: Option<Vec<SwapStatus>>,
+    /// Epoch time, in seconds. If set, acts as filter for minimum swap creation time, inclusive.
+    pub from_timestamp: Option<i64>,
+    /// Epoch time, in seconds. If set, acts as filter for maximum swap creation time, exclusive.
+    pub to_timestamp: Option<i64>,
+    pub offset: Option<u32>,
+    pub limit: Option<u32>,
+}
+
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct BuyBitcoinRequest {
     pub provider: BuyBitcoinProvider,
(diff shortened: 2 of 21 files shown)


libs/sdk-core/src/persist/swap.rs
```diff
@@ -329,24 +331,54 @@ impl SqliteStorage {
...
-        let mut stmt = con.prepare(&self.select_swap_query("true", ""))?;
+        if where_clause.is_empty() {
+            where_clause.push_str("true");
+        }
+
+        let mut query = self.select_swap_query(&where_clause, "");
+
+        match req.limit {
+            Some(limit) => query.push_str(&format!("LIMIT {}\n", limit)),
+            None => query.push_str("LIMIT -1\n"),
```
roeierez, line 374: NIT: do we have to put LMIT in this case?

libs/sdk-core/src/persist/swap.rs
```diff
@@ -416,24 +448,34 @@ impl SqliteStorage {
     }
 }
 
+fn append_and_clause(s: &mut String, clause: &str) {
```
roeierez, line 451: Perhaps it is better to use a slice of conditions and then when formatting the query use string join with "AND" ?

roeierez (changes requested):
Looks good! two small comments.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -579,6 +579,14 @@ dictionary SwapInfo {
     u32? confirmed_at;
 };
 
+dictionary ListSwapsRequest {
+    sequence<SwapStatus>? status = null;
+    i64? from_timestamp = null;
+    i64? to_timestamp = null;
```
ok300, line 585: Maybe `u64` is better, since epoch time cannot be negative?

libs/sdk-core/src/models.rs
```diff
@@ -959,6 +959,17 @@ pub struct ReceiveOnchainRequest {
     pub opening_fee_params: Option<OpeningFeeParams>,
 }
 
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct ListSwapsRequest {
+    pub status: Option<Vec<SwapStatus>>,
+    /// Epoch time, in seconds. Filters swap creation time.
```
ok300, line 965:
```suggestion
    /// Epoch time, in seconds. If set, acts as filter for minimum swap creation time, inclusive.
```

libs/sdk-core/src/models.rs
```diff
@@ -959,6 +959,17 @@ pub struct ReceiveOnchainRequest {
...
+    pub from_timestamp: Option<i64>,
+    /// Epoch time, in seconds. Filters swap creation time.
```
ok300, line 967:
```suggestion
    /// Epoch time, in seconds. If set, acts as filter for maximum swap creation time, exclusive.
```

libs/sdk-core/src/persist/swap.rs
JssDWt: You can't have OFFSET only, so this is defense against the client putting only OFFSET

libs/sdk-bindings/src/breez_sdk.udl
JssDWt: Makes sense, I copied this from `list_payments`

JssDWt: Actually it probably makes more sense to keep the api equivalent between list_payments and this one. So we have the option of making this an i64, or change list_payments to u64. I prefer i64 here, because it doesn't cause breaking changes, what do you think?

libs/sdk-core/src/models.rs
JssDWt: Done

libs/sdk-core/src/models.rs
JssDWt: Done

libs/sdk-core/src/persist/swap.rs
JssDWt: Done

roeierez (approved):
LGTM

libs/sdk-bindings/src/breez_sdk.udl
ok300: I see, sounds good.

ok300 (approved):

PR #1121, opened 2024-11-10
Expand `InputType` with new variant: Bolt12 offer

ok300:
This PR extends the input parsed in `sdk-common` to recognize Bolt12 offers.


ok300:
@hydra-yse your older PR #598 parses the offer to more detailed fields:

https://github.com/breez/breez-sdk-greenlight/blob/ad384b106bfcee3e225625b879b153f398b6ee7f/libs/sdk-bindings/src/breez_sdk.udl#L735-L755

As that PR is still in review, I'm not sure if that structure is final (`LNOffer`, `InputType::Bolt12Offer(LNOffer offer)`).

So would you say we need those fields as defined there? If so, I can integrate them in this PR.

dangeross:
> So would you say we need those fields as defined there? If so, I can integrate them in this PR

I think they are needed, we need to parse the offer for the user. I also think our PR struct is out of date and I would suggest bumping the `lightning`/`lightning-invoice` dependencies. The current version doesn't parse offers according to the latest spec (e.g. the description field is now optional), which I think is not compatible with Phoenix

JssDWt:
> As that PR is still in review, I'm not sure if that structure is final (`LNOffer`, `InputType::Bolt12Offer(LNOffer offer)`).
> 
> So would you say we need those fields as defined there? If so, I can integrate them in this PR.

Also keep in mind there are invoice requests and invoices in bolt12, next to offers. We don't have to support them, but let's make the interface in such a way that we don't have to break anything to support them in the future.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -734,6 +734,7 @@ dictionary RecommendedFees {
 interface InputType {
     BitcoinAddress(BitcoinAddressData address);
     Bolt11(LNInvoice invoice);
+    Bolt12(string offer);
```
roeierez, line 737: I think we should have here the parsed LNOffer (same as we have for other types

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -738,10 +738,27 @@ dictionary RecommendedFees {
...
+    Amount? amount;
+    u64? absolute_expiry;
+    string? issuer;
+};
+
 [Enum]
 interface InputType {
     BitcoinAddress(BitcoinAddressData address);
     Bolt11(LNInvoice invoice);
+    Bolt12(LNOffer offer);
```
JssDWt, line 761: Let's call this Bolt12Offer?
https://github.com/breez/breez-sdk-greenlight/pull/1121#issuecomment-2468030319

ok300: Done: a8a7057320aca4f5b03708a11de0ea66dc931d80

ok300:
As discussed, this is adding a simple `Bolt12Offer { offer: String }` variant, until `gl-client` supports the newer `lightning` and `lightning-invoice` crates.

In addition, Liquid SDK will attempt to parse the offer locally.

libs/sdk-bindings/src/breez_sdk.udl
ok300: Addressed: https://github.com/breez/breez-sdk-greenlight/pull/1121#issuecomment-2472855649

libs/sdk-common/src/invoice.rs
```diff
@@ -133,8 +133,10 @@ impl From<lightning::offers::offer::Amount> for Amount {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct LNOffer {
+    /// String representation of the Bolt12 offer
     pub bolt12: String,
```
dangeross, line 137: Maybe `offer` is more understandable?

libs/sdk-common/src/invoice.rs
```diff
@@ -101,6 +101,50 @@ fn format_short_channel_id(id: u64) -> String {
...
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum Amount {
+    Bitcoin {
+        amount_msat: u64,
+    },
+    Currency {
+        // Reference to [FiatCurrency.id]
+        iso4217_code: String,
```
dangeross, line 111: Maybe a name that more obviously links it to FiatCurrency? something like `currency_id`

libs/sdk-common/src/invoice.rs
```diff
@@ -101,6 +101,50 @@ fn format_short_channel_id(id: u64) -> String {
...
+        fractional_amount: u64,
```
dangeross, line 112: Would this be like Euro cents? and you can use the FiatCurrency info to calculate the decimal equivalent

libs/sdk-common/src/invoice.rs
```diff
@@ -101,6 +101,50 @@ fn format_short_channel_id(id: u64) -> String {
...
+        match amount {
+            lightning::offers::offer::Amount::Bitcoin { amount_msats } => Amount::Bitcoin {
+                amount_msat: amount_msats,
+            },
+            lightning::offers::offer::Amount::Currency {
+                iso4217_code,
+                amount,
+            } => Amount::Currency {
+                iso4217_code: String::from_utf8(iso4217_code.to_vec())
+                    .expect("Expecting a valid ISO 4217 character sequence"),
```
dangeross, line 127: Better to do a TryFrom?

libs/sdk-common/src/invoice.rs
ok300: This doesn't necessarily link it to `FiatCurrency`. It just happens they both use the same type of ID. `currency_id` might be obvious for us, but maybe not for others who aren't aware of `FiatCurrency`.

 IMO `iso4217_code` is clearer to the caller.

libs/sdk-common/src/invoice.rs
ok300: Good idea, done: caf483b6eebf860e079c239858f2900c82302955

libs/sdk-common/src/invoice.rs
ok300: Done: 643928e4cab2aa0a77f9df5b1c202706181fe293

dangeross (approved):
LGTM

hydra-yse (commented):
Should we also add parsing for `recurrence-` and `quantity` based fields, or are those outside the scope of what will be used? Perhaps they can be added later as they seem useless for now. Besides that, looking good! :+1:

ok300:
IMO it's best to add them when they will also be tested and used.

hydra-yse (approved):

PR #1122, opened 2024-11-11
Android: Replaced with "$rootDir" with "$projectDir"

larcho:
This handles relative paths better. For instance, we use a Monorepo where our node_modules is located somewhere else.


dangeross (approved):
LGTM, based on the test here: https://github.com/breez/breez-sdk-liquid/pull/558

PR #1123, opened 2024-11-14
Clarify `receive_onchain` docs

danielgranhao:
I adjusted the docs according to my understanding of the method. As an SDK user, the word "try" made me think that it would be on a best-effort basis, and redeeming would still work even if the provided params were not valid when the swap is ready to be redeemed.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -807,7 +807,8 @@ impl BreezServices {
     /// Onchain receive swap API
     ///
     /// Create and start a new swap. A user-selected [OpeningFeeParams] can be optionally set in the argument.
-    /// If set, and the operation requires a new channel, the SDK will try to use the given fee params.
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -807,7 +807,8 @@ impl BreezServices {
     /// Onchain receive swap API
     ///
     /// Create and start a new swap. A user-selected [OpeningFeeParams] can be optionally set in the argument.
-    /// If set, and the operation requires a new channel, the SDK will try to use the given fee params.
+    /// If set, and the operation requires a new channel, the SDK will enforce the use the given fee params.
```
JssDWt, line 810: Nit: There's 2 times `the` in the sentence. Suggestion:
```suggestion
    /// If set, and the operation requires a new channel, the SDK will use the given fee params.
```

JssDWt (approved):
ACK 08579a965fe72493c782cc4b33120ef9059466d2

dangeross (approved):

PR #1126, opened 2024-11-16
allow mpp in open channel invoices

JssDWt:
Invoices for channel opens didn't have the feature flag 'basic mpp' set. This means senders would not attempt to pay the open channel invoice if their balance was spread out over multiple channels, but not one of the channels contains the full amount to pay the invoice.

This PR sets the basic mpp feature in open channel invoices, so they can be paid with MPP.


dangeross:
Effectively covers any receive payment where the SDK has to recreate an invoice right? (Like empty/missing LSP route hint)

JssDWt:
> Effectively covers any receive payment where the SDK has to recreate an invoice right? (Like empty/missing LSP route hint)

Yes, that's right

dangeross (approved):

roeierez (approved):

PR #1127, opened 2024-11-18
expose reverse swap fees

### LINKED ISSUE #1111: Exposing fees in `ReverseSwapInfo`
Sending an onchain payment using a reverse swap results in a `Payment` that contains a `ReverseSwapInfo`. It would be useful if `ReverseSwapInfo` included the fees that were spent for the lockup and claim txs, as well as the swap provider fees. Right now we can only find out how much was spent on fees in total by comparing the sent LN funds and the resulting sent onchain funds.

JssDWt:
Expose the lockup/claim/service fees from reverse swaps.

Historical fees are currently not backfilled. So this currently only works for new swaps.
The fees are synced with the remote.

If the fees cannot be found, they're set to 0.

Fixes #1111


libs/sdk-core/src/swap_out/reverseswap.rs
```diff
@@ -343,47 +312,47 @@ impl BTCSendSwap {
...
+                    claim_pubkey: req.recipient_address,
                     invoice: response.invoice,
                     preimage: reverse_swap_keys.preimage,
                     private_key: reverse_swap_keys.priv_key,
                     timeout_block_height: response.timeout_block_height,
                     id: response.id,
                     onchain_amount_sat: response.onchain_amount,
-                    sat_per_vbyte,
-                    receive_amount_sat,
+                    sat_per_vbyte: None,
```
roeierez, line 341: Can we remove now the sat_per_vbyte from the struct?

roeierez (approved):
One small comment, otherwise LGTM.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -602,7 +602,7 @@ dictionary PrepareOnchainPaymentResponse {
     f64 fees_percentage;
     u64 fees_lockup;
     u64 fees_claim;
-
+    u64 fees_service;
```
ok300, line 605: `fees_percentage` above and the new `fees_service` refer to the same thing.

libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -615,6 +615,10 @@ dictionary ReverseSwapInfo {
     string? claim_txid;
     u64 onchain_amount_sat;
     ReverseSwapStatus status;
+    u64 fees_lockup;
+    u64 fees_claim;
+    u64 fees_service;
+    u64 total_fees;
```
ok300, line 621: To keep with the other 3 new fields, maybe `fees_total` is better?

libs/sdk-core/src/persist/migrations.rs
```diff
@@ -642,5 +642,11 @@ pub(crate) fn current_sync_migrations() -> Vec<&'static str> {
         UPDATE swaps SET max_swapper_payable = max_allowed_deposit;
         ",
         "ALTER TABLE payments_external_info ADD COLUMN lnurl_pay_comment TEXT;",
+        "
+        ALTER TABLE reverse_swaps ADD column fees_lockup INTEGER DEFAULT 0;
+        ALTER TABLE reverse_swaps ADD column fees_claim INTEGER DEFAULT 0;
+        ALTER TABLE reverse_swaps ADD column fees_service INTEGER DEFAULT 0;
+        ALTER TABLE reverse_swaps ADD column total_fees INTEGER DEFAULT 0;
```
ok300, line 649: Since the `udl` doesn't treat them as optional, maybe `INTEGER NOT NULL DEFAULT 0` is more appropriate?

libs/sdk-core/src/swap_out/reverseswap.rs
Patti88: When running cargo build, I noticed that the openssl-sys crate failed to build, citing a missing Perl module, Locale::Maketext::Simple. This module is needed for OpenSSL's configuration. It seems plausible that this missing module is contributing to the overall build failure.

PR #1129, opened 2024-11-26
Add namespace to gradle files on Flutter plugin to be compatible with AGP 8.+

erdemyerebasmaz:
This PR addresses compatibility with Android Gradle Plugin (AGP) 8.+, the second error seen on issue https://github.com/breez/breez-sdk-greenlight/issues/1128

### Changelist
- Add namespace to gradle files on Flutter plugin
  - AGP 8.+ requires each module to declare a namespace. This change ensures the plugin is compliant with the latest AGP requirements.
- Removing unnecessary `AndroidManifest.xml`
  - As the namespace is now declared in the Gradle files, the redundant AndroidManifest.xml has been removed.  
- Explicitly specifying types for decoded values
  - Addresses type inference failures.

### DIFF
--- a/libs/sdk-flutter/android/build.gradle
+++ b/libs/sdk-flutter/android/build.gradle
@@ -1,36 +1,23 @@
+plugins {
+    id 'com.android.library'
+    id 'org.jetbrains.kotlin.android'
+    id 'org.jetbrains.kotlin.plugin.serialization'
+}
+
 group 'com.breez.breez_sdk'
 version '1.0-SNAPSHOT'
 
-buildscript {
-    ext.kotlin_version = '1.8.20'
-    repositories {
-        google()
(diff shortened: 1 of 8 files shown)


libs/sdk-flutter/android/build.gradle
dangeross, line 1: Should we update the compileSdk/java version in `libs/sdk-bindings/bindings-android` and `libs/sdk-react-native/android` also or is it fine to leave them as is?

erdemyerebasmaz: Ideally yes, but cope of this PR only concerns Flutter plugin. Other bindings can be addressed in separate PRs if needed.

erdemyerebasmaz:
Superseded by #1206

PR #1132, opened 2024-11-28
configure grpc connection for breez server

roeierez:
This PR allows Breez grpc connection to recover from a stale connection in no more than 5 seconds. Previously it could take up to 20 seconds.

### DIFF
--- a/libs/sdk-common/src/breez_server.rs
+++ b/libs/sdk-common/src/breez_server.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use anyhow::Result;
 use log::trace;
 use tokio::sync::Mutex;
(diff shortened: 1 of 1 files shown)


libs/sdk-common/src/breez_server.rs
```diff
@@ -29,18 +31,26 @@ pub struct BreezServer {
...
-            Endpoint::from_shared(self.server_url.clone())?.connect_lazy();
+        *self.grpc_channel.lock().await = Self::create_endpoint(&self.server_url)?.connect_lazy();
         Ok(())
     }
 
+    fn create_endpoint(server_url: &str) -> Result<Endpoint> {
+        Ok(Endpoint::from_shared(server_url.to_string())?
+            .http2_keep_alive_interval(Duration::new(5, 0))
+            .tcp_keepalive(Some(Duration::from_secs(5)))
+            .http2_keep_alive_interval(Duration::from_secs(5))
```
JssDWt, line 49: This is duplicate

libs/sdk-common/src/breez_server.rs
```diff
@@ -29,18 +31,26 @@ pub struct BreezServer {
...
-        *self.grpc_channel.lock().await =
-            Endpoint::from_shared(self.server_url.clone())?.connect_lazy();
+        *self.grpc_channel.lock().await = Self::create_endpoint(&self.server_url)?.connect_lazy();
         Ok(())
     }
 
+    fn create_endpoint(server_url: &str) -> Result<Endpoint> {
+        Ok(Endpoint::from_shared(server_url.to_string())?
+            .http2_keep_alive_interval(Duration::new(5, 0))
+            .tcp_keepalive(Some(Duration::from_secs(5)))
```
JssDWt, line 48: The default keepalive interval is 20 seconds. Is it necessary to put this to 5 seconds?

roeierez: Yes I think it is better and more tailor made for mobile

JssDWt: Ok sounds good.

JssDWt:
For clippy rebase to main

JssDWt (approved):
LGTM
It might be that gl-client gets different keepalive settings: https://github.com/Blockstream/greenlight/pull/548

roeierez:
> LGTM It might be that gl-client gets different keepalive settings: [Blockstream/greenlight#548](https://github.com/Blockstream/greenlight/pull/548)

I added a comment there: https://github.com/Blockstream/greenlight/pull/548/files#r1863998844
I will wait for Christian to respond and ensure we both using the same settings.

PR #1133, opened 2024-11-29
detect and act on network change

JssDWt:
When there is a network change, tonic behaves as follows:
- After the keepalive timeout, reconnect automatically.
- Before the keepalive timeout, any grpc call will time out. After the timeout, the connection is reestablished.

This commit adds a mechanism to reconnect all grpc clients after one of the clients detects a network change. Initially it was attempted to only retry based on a keepalive timeout error, but a network change affects all grpc clients, so subsequent requests to other grpc endpoints would still fail with a timeout. Ofcourse those grpc clients can also add a retry-on-timeout, but since it is known at this point the grpc clients are temporarily dead, reconnect them immediately. This ensures subsequent calls to other endpoints won't add additional time by waiting for a timeout.

Currently this PR wraps the `trampoline_pay`, `pay`, and `invoice` calls in the greenlight client.

tokio was bumped to allow cloning watch::Sender

### DIFF
--- a/libs/sdk-common/src/tonic_wrap/mod.rs
+++ b/libs/sdk-common/src/tonic_wrap/mod.rs
@@ -20,9 +20,10 @@ impl Display for Status {
 
 pub struct TransportError(pub tonic::transport::Error);
 
-const BROKEN_CONNECTION_STRINGS: [&str; 2] = [
+const BROKEN_CONNECTION_STRINGS: [&str; 3] = [
     "http2 error: keep-alive timed out",
     "connection error: address not available",
+    "connection error: timed out",
 ];
 
 impl Display for TransportError {
@@ -77,7 +78,7 @@ where
     // It's a bit of a guess which errors can occur here. hyper Io errors start
     // with 'connection error'. These are some of the errors seen before.
     if !BROKEN_CONNECTION_STRINGS.contains(&source.to_string().as_str()) {
-        debug!("transport error string is: {}", source.to_string());
+        debug!("transport error string is: '{}'", source.to_string());
         return Err(status);
     }
 
--- a/libs/sdk-common/src/tonic_wrap/mod.rs
+++ b/libs/sdk-common/src/tonic_wrap/mod.rs
@@ -1,5 +1,7 @@
 use std::{error::Error, fmt::Display, future::Future};
 
+use log::debug;
+
 pub struct Status(pub tonic::Status);
 
 impl Display for Status {
@@ -75,6 +77,6 @@ where
         "with_connection_fallback: initial call failed due to keepalive 
         timeout. Retrying fallback."
     );
-    let res = fallback().await;
-    res
+
+    fallback().await
 }
--- a/libs/sdk-common/src/tonic_wrap/mod.rs
+++ b/libs/sdk-common/src/tonic_wrap/mod.rs
@@ -20,10 +20,11 @@ impl Display for Status {
 
 pub struct TransportError(pub tonic::transport::Error);
 
-const BROKEN_CONNECTION_STRINGS: [&str; 3] = [
+const BROKEN_CONNECTION_STRINGS: [&str; 4] = [
     "http2 error: keep-alive timed out",
     "connection error: address not available",
     "connection error: timed out",
+    "connection error: unexpected end of file",
 ];
 
 impl Display for TransportError {
--- a/libs/sdk-common/src/tonic_wrap/mod.rs
+++ b/libs/sdk-common/src/tonic_wrap/mod.rs
@@ -1,6 +1,4 @@
-use std::{error::Error, fmt::Display, future::Future};
-
-use log::debug;
+use std::{error::Error, fmt::Display};
 
 pub struct Status(pub tonic::Status);
 
@@ -20,13 +18,6 @@ impl Display for Status {
 
 pub struct TransportError(pub tonic::transport::Error);
 
-const BROKEN_CONNECTION_STRINGS: [&str; 4] = [
-    "http2 error: keep-alive timed out",
-    "connection error: address not available",
-    "connection error: timed out",
-    "connection error: unexpected end of file",
-];
-
 impl Display for TransportError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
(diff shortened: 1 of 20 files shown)


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -704,6 +715,61 @@ impl Greenlight {
...
+        if !source.to_string().contains("keep-alive timed out") {
+            return Err(status);
+        }
+
+        info!("AAAAAAAAAAAAAAA is connection error");
+        if self.network_change_detector.send(()).is_err() {
+            warn!("greenlight network change detector died.");
+        }
+
+        let res = fallback().await;
```
dangeross, line 767: Do we need to wait at all to allow time for the GRPC clients to reconnect?

JssDWt: No it looks like the connection is immediately reusable.

You can test this with the sdk-cli by:
- connecting the sdk
- changing your laptop to another network
- calling send_payment or lnurl_pay with or without --use_trampoline

It will hang for a long time (until this PR https://github.com/Blockstream/greenlight/pull/548 is included) and then continue with the fallback successfully.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -958,6 +1024,16 @@ impl Greenlight {
...
+
+    // async fn trampoline_retry(&self, request: TrampolinePayRequest, payment_hash: Vec<u8>) -> Result<tonic::Response<TrampolinePayResponse>, tonic::Status> {
+    //     let node_client = self.get_node_client().await?;
+    //     let wait_resp = node_client.wait_send_pay(WaitsendpayRequest{
+    //         payment_hash,
+    //         ..Default::default()
+    //     }).await?;
+    //     wait_resp.
+    //     todo!()
+    // }
```
dangeross, line 1036: Want to remove this?

dangeross (approved):
LGTM, with a nit

libs/sdk-core/src/breez_services.rs
```diff
@@ -1491,6 +1493,34 @@ impl BreezServices {
         Ok(())
     }
 
+    fn detect_network_change(self: &Arc<BreezServices>) {
```
roeierez, line 1496: with_connection_fallback looks great on node_api. Not sure propagating this to other clients to reconnect is necessary? if we set keep alive to 5 seconds  all other connections will recover more or less at the same time independently.

libs/sdk-core/src/greenlight/node_api.rs
JssDWt: Done

libs/sdk-core/src/breez_services.rs
JssDWt: That then depends on the keepalive pings being sent around the same time. When testing this I got an error from `lsp_info`, which was called in `do_sync` after the payment completed. That hanged and timed out. I think reying on those keepalive pings is too unreliable.

roeierez: I agree it won't be at the same time but I think also we can't avoid entirely from transport errors even from the network change detection.
The case you mentioned is indeed interesting because based on my tests if all connections are set to 5 seconds keep alive timeout  and assuming payment takes more than 5 seconds it shouldn't happen because the sync that comes afterward should be already with recovered connections. Are you sure the tests were done with this PR that configures correctly the breez server timeouts: https://github.com/breez/breez-sdk-greenlight/pull/1133

JssDWt: No, they weren't done with those breez server timeouts. Greenlight may potentially get different timeouts here: https://github.com/Blockstream/greenlight/pull/548. Also a little calculation is quickly made:
keepalive pings between clients can be 5 seconds apart. Consider the worst case:

keepalive ping from gl-client timed out now, keepalive ping from breez server client times out in 5 seconds. gl-client retries and finishes the call within 2 seconds. There's 3 seconds left to fall into the breez server client timing out.

roeierez: That's right. So basically you are trying to solve the main use cases (payment, invoice) that starts with greenlight and make sure breez server connection that is involved later won't break it. Did I get it right?
I think it is a good goal indeed and I am not against the proposed solution just looking to ensure there is not simpler one.
How about minimizing the breez server keep alive interval and timeout to 1-2 seconds? keep alive is just a small 1 tcp packet which seems very negligible.
No payment/invoice API will be finished in 2 seconds (I think only reconnecting takes more than a second).
If you have a strong opinion that we need the network change detection from greenlight then let's do it.

JssDWt: Maybe wrap all (most) grpc calls in with_connection_fallback?

roeierez: Sounds good to me!

JssDWt:
Updated to wrap most grpc calls in `with_connection_fallback`

roeierez (approved):
LGTM with one comment to check if we can pass only the called function to the `with_connection_fallback`

libs/sdk-common/src/tonic_wrap/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+            "description: {:?}, source: {:?}",
+            self.0.to_string(),
+            self.0.source(),
+        )
+    }
+}
+
+pub async fn with_connection_fallback<T, M, F>(
+    main: M,
+    fallback: impl FnOnce() -> F,
```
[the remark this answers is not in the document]
JssDWt: I tried this, if you want you can try yourself too.
I tried by passing in a client to the function and having a `Fn` that operates on the client. I got stuck on lifetimes. It would be the cleaner way though.

JssDWt:
Payment was still getting stuck. Added more wrappers.

libs/sdk-common/src/tonic_wrap/mod.rs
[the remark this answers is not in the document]
roeierez: I think we can use a macro here so we can use it like this:

```
retry! {
 client.lsp_list(request.clone()).await
}
```

I will give it a try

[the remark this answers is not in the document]
roeierez: It seems that this macro should do it:

```
#[macro_export]
macro_rules! retry_grpc {
    ($f:expr) => {{
      use std::error::Error;
      use log::debug;
      const BROKEN_CONNECTION_STRINGS: [&str; 3] = [
          "http2 error: keep-alive timed out",
          "connection error: address not available",
          "connection error: timed out",
      ];

      let res = $f;
      let result = match res {
          Ok(t) => Ok(t),
          Err(status) => {
            let mut retruned = Err(status.clone());
            if let Some(source) = status.source() {
              if let Some(error) = source.downcast_ref::<tonic::transport::Error>() {
                if error.to_string() == "transport error" {
                  if let Some(source) = error.source() {
                    if BROKEN_CONNECTION_STRINGS.contains(&source.to_string().as_str()) {
                      debug!("retry_grpc: initial call failed due to broken connection. Retrying.");
                      retruned = $f
                    }
                  }
                }
              }
            }
            retruned
          },
      };
      result
   }};
}
```

Then it can be used as:

```
retry_grpc! {
 client.lsp_list(request.clone()).await
}
```

JssDWt:
@roeierez Found another way to make the compiler happy by passing a single function, what do you think?

```
pub async fn with_connection_retry<C, RQ, RS>(
    client: &mut C, 
    req: RQ,
    f: impl for<'c> Fn(&'c mut C, RQ) -> Pin<Box<dyn Future<Output = Result<tonic::Response<RS>, tonic::Status>> + Send + 'c>>,
) -> Result<tonic::Response<RS>, tonic::Status>
where
    RQ: std::fmt::Debug + Clone,
    RS: std::fmt::Debug,
{
    let res = f(client, req.clone()).await;
    ...
    f(client, req).await
}
```

```
        let chain_api_servers = with_connection_retry(
            &mut client,
            ChainApiServersRequest {},
            |client, req| Box::pin(client.chain_api_servers(req)),
        )
            .await
```

JssDWt:
Now the retry logic is wrapped in a macro. Posting the commit message here:

A macro makes for a little less code duplication. The macro returns a
single awaitable future. So the result of the macro can be used in join!
calls. I was unable to do this without consuming the grpc client object.
Therefore the grpc client object has to be cloned if it's used later
again.

roeierez (approved):
LGTM

PR #1135, opened 2024-12-02
retry on any transport error with a source

JssDWt:
It appears many more error strings are seen than the ones specified in `BROKEN_CONNECTION_STRINGS`. Among others:
`http2 error: connection error received: not a result of an error`
`connection error: connection aborted`
`error trying to connect: dns error: failed to lookup address information: No address associated with hostname`
`error trying to connect: tcp connect error: Software caused connection abort (os error 103)`
`error trying to connect: Software caused connection abort (os error 103)`

It appears there's a vast amount of different errors that can occur in the situation where the grpc channel would repair itself. Therefore retry on any transport error.


roeierez (approved):
LGTM

PR #1136, opened 2024-12-02
remove hibernation logic

JssDWt:
The reconnect after hibernation does not fix the root cause of errors originating from lost grpc connections after hibernation. When the app comes back from hibernation, any already outgoing grpc request will fail with a transport error. The only way to get around that is by retrying the request. Also, grpc connections are automatically reestablished after such a failure occurs. If the grpc client gets a timeout or broken connection failure on one request, the next request will succeed with a fresh underlying connection.

retrying requests is implemented in #1133


roeierez (approved):
LGTM

dangeross (approved):

PR #1099, opened 2024-10-07
improve payment sync

JssDWt:
- make payment sync call all invoice/send_pay endpoints simultaneously.
- fix missing amounts in payments by taking amount from the invoice like previous
- fix wrong description for payments, now take the invoice description
- fix issue where pending payments were wrongfully removed from the db, removed the pending payment deletion. They should be updated by sync automatically.

Hopefully fixes https://github.com/breez/c-breez/issues/909

PR #1140, opened 2024-12-05
delete pending payments unknown to the node

JssDWt:
Before a payment is made, a pending payment is inserted in the database [here](https://github.com/breez/breez-sdk-greenlight/blob/45c941144dc05f8901b4e042e7f6b488d875cbc5/libs/sdk-core/src/breez_services.rs#L310). It can happen that there is never a corresponding payment made on the node. Like with connection errors. In that case we no longer remove the pending payment since the sync performance improvements in #1099.

In this PR, the pending payments are passed to the node api on sync. If there are any pending payments that are not synced in that sync round, the node api will check whether they exist on the node. If not, they will be deleted.

Tested this with existing pending payments on my local database. Also tested by making send_payment return an error immediately locally.

I have added the `array` extension to sqlite in order to delete payments from the database based on the list of payment hashes. Loading that extension took 0.000001 to 0.000012s on my machine.


libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -729,6 +731,68 @@ impl Greenlight {
...
+        if pending_map.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let client = self.get_node_client().await?;
+
+        // If there's any pending payments remaining, check whether they exist
+        // in parallel.
+        let mut futures = Vec::new();
+        for hash in pending_map {
```
roeierez, line 768: Do you think it is better to initiate one query for all pending payments here instead a query per payment?

JssDWt: You mean fetching all payments and checking whether our pending payment is there?
I don't think so, no, because usually this will be a single payment.

JssDWt:
Updated to treat the payment inserted before sending the actual payment as a 'pseudo' payment.

roeierez (approved):

dangeross (approved):

PR #1143, opened 2024-12-11
Support external input parsing in sdk-common

danielgranhao:
Needed for https://github.com/breez/breez-sdk-liquid/issues/535.

Adds support for external parsers to sdk-common. The immediate need is to have this in the liquid SDK, but it would be nice to have it in the common parse implementation to make it easy to use in the greenlight version later.


libs/sdk-common/src/input_parser.rs
```diff
@@ -227,6 +273,63 @@ pub async fn parse(input: &str) -> Result<InputType> {
...
 }
 
+/// Parse input using provided external parsers.
+async fn parse_external(
+    input: &str,
+    external_input_parsers: &[ExternalInputParser],
+) -> Result<InputType> {
+    for parser in external_input_parsers {
+        // Check regex
+        let re = Regex::new(&parser.input_regex).context("Couldn't parse regex")?;
```
roeierez, line 283: NIT: perhaps also it is useful to add the provider_id to the error?

roeierez (approved):
LGTM! added a small NIT.

dangeross (approved):
LGTM

libs/sdk-common/src/input_parser.rs
danielgranhao: Added in [86ee6e0](https://github.com/breez/breez-sdk-greenlight/pull/1143/commits/86ee6e0f9670116239d543b98576c493c5008870)

PR #1149, opened 2024-12-20
Common BIP353 address parser

lorenzoronzani:
feat: I implemented within the sdk-common the function to parse a BIP353 address and extract the corresponding bolt12 or lnurl-pay offer. To achieve that I needed a dns-resolver and I decided to use hickory-resolver crate.

feat: edited also the cli code to integrate and inject from outside the dns_resolver dependency

This PR is the continuing part of this one https://github.com/breez/breez-sdk-liquid/pull/619 in the breez-sdk-liquid repo.

I tried to implement the advice, but for the decoding part of the TXT data I decided at the moment to keep things separate from the parse_core, because of the time required to better analyze how messages are processed within the parse core.

### DIFF
--- a/libs/sdk-common/src/input_parser.rs
+++ b/libs/sdk-common/src/input_parser.rs
@@ -20,11 +20,10 @@ use LnUrlRequestData::*;
 use crate::prelude::*;
 
 const USER_BITCOIN_PAYMENT_PREFIX: &str = "user._bitcoin-payment";
-#[cfg(feature = "liquid")]
-const BOLT12_PREFIX: &str = "lno=";
-const LNURL_PAY_PREFIX: &str = "lnurl=";
+// #[cfg(feature = "liquid")]
+const BOLT12_PREFIX: &str = "lno";
+const LNURL_PAY_PREFIX: &str = "lnurl";
 const BIP353_PREFIX: &str = "bitcoin:";
-const QUERY_PARAMS_SEPARATOR: &str = "&";
 
 lazy_static! {
     static ref DNS_RESOLVER: TokioAsyncResolver = {
@@ -226,6 +225,13 @@ pub async fn parse(
     Err(anyhow!("Unrecognized input type"))
 }
 
+fn get_by_key(tuple_vector: &Vec<(&str, &str)>, key: &str) -> Option<String> {
+    tuple_vector
+        .iter()
+        .find(|(k, _)| *k == key)
+        .map(|(_, v)| v.to_string())
+}
+
 async fn bip353_parse(
     input: &str,
     dns_resolver: &AsyncResolver<GenericConnector<TokioRuntimeProvider>>,
@@ -257,7 +263,25 @@ async fn bip353_parse(
                     return None;
                 }
 
-                return Some(decoded);
+                println!("Decoded: {}", decoded);
+
+                if let Some((_, query_part)) = decoded.split_once(&format!("{}?", BIP353_PREFIX)) {
+                    println!("Query_part: {}", query_part);
+
+                    let query_params = querystring::querify(query_part);
+
+                    println!("Query_params: {:?}", query_params);
+
+                    if let Some(bolt12_address) = get_by_key(&query_params, BOLT12_PREFIX) {
+                        println!("BOLT 12: {:?}", get_by_key(&query_params, BOLT12_PREFIX));
+                        return Some(bolt12_address);
+                    }
+
+                    if let Some(lnurl) = get_by_key(&query_params, LNURL_PAY_PREFIX) {
+                        println!("LNURL: {:?}", get_by_key(&query_params, LNURL_PAY_PREFIX));
+                        return Some(lnurl);
+                    }
+                }
             }
             Err(e) => {
                 error!("Failed to decode TXT data: {}", e);
@@ -268,26 +292,26 @@ async fn bip353_parse(
     None
 }
 
-#[cfg(feature = "liquid")]
-fn extract_bolt12_offer(input: &str) -> Option<String> {
-    return extract_by_label(input, BOLT12_PREFIX);
-}
+// // #[cfg(feature = "liquid")]
+// fn extract_bolt12_offer(input: &str) -> Option<String> {
+//     return extract_by_label(input, BOLT12_PREFIX);
+// }
 
-fn extract_lnurl(input: &str) -> Option<String> {
-    return extract_by_label(input, LNURL_PAY_PREFIX);
-}
+// fn extract_lnurl(input: &str) -> Option<String> {
+//     return extract_by_label(input, LNURL_PAY_PREFIX);
+// }
 
-fn extract_by_label(input: &str, label: &str) -> Option<String> {
-    if let Some((_, value)) = input.split_once(label) {
-        let value = value
-            .split_once(QUERY_PARAMS_SEPARATOR)
-            .map_or(value, |(first, _)| first);
+// fn extract_by_label(input: &str, label: &str) -> Option<String> {
+//     if let Some((_, value)) = input.split_once(label) {
+//         let value = value
+//             .split_once(QUERY_PARAMS_SEPARATOR)
+//             .map_or(value, |(first, _)| first);
 
-        return Some(value.to_string());
-    }
+//         return Some(value.to_string());
+//     }
 
-    None
-}
+//     None
+// }
 
 /// Core parse implementation
 async fn parse_core(input: &str) -> Result<InputType> {
@@ -312,11 +336,6 @@ async fn parse_core(input: &str) -> Result<InputType> {
             }),
             Some(invoice) => Ok(InputType::Bolt11 { invoice }),
         };
-    }
-
-    #[cfg(feature = "liquid")]
-    if let Some(bolt12_offer) = extract_bolt12_offer(input) {
-        // TODO: Add the section to manage a bolt12_offer
     }
 
     #[cfg(feature = "liquid")]
@@ -357,13 +376,6 @@ async fn parse_core(input: &str) -> Result<InputType> {
             return Ok(InputType::Url { url: input.into() });
         }
     }
-
-    let lnurl = match extract_lnurl(input) {
-        Some(value) => value,
-        None => input.to_string(),
-    };
-
-    let input = lnurl.as_str();
 
     // Try to strip the "lightning:" prefix from possible lnurl string. If prefix is not there, default to original input
     let input = input
--- a/libs/sdk-common/src/input_parser.rs
+++ b/libs/sdk-common/src/input_parser.rs
@@ -23,6 +23,7 @@ const USER_BITCOIN_PAYMENT_PREFIX: &str = "user._bitcoin-payment";
 const BOLT12_PREFIX: &str = "lno";
 const LNURL_PAY_PREFIX: &str = "lnurl";
 const BIP353_PREFIX: &str = "bitcoin:";
+const BIP353_PREFIX_COUNT_CONSTRAINT: usize = 1;
 
 lazy_static! {
     static ref DNS_RESOLVER: TokioAsyncResolver = {
@@ -224,6 +225,28 @@ pub async fn parse(
     Err(anyhow!("Unrecognized input type"))
 }
 
+fn validate_txt_record(decoded: &str) -> Option<bool> {
+    if !decoded.to_lowercase().starts_with(BIP353_PREFIX) {
+        error!(
+            "Invalid decoded TXT data (doesn't begin with: {})",
+            BIP353_PREFIX
+        );
+
+        return None;
+    }
+
+    if !decoded.to_lowercase().matches(BIP353_PREFIX).count() == BIP353_PREFIX_COUNT_CONSTRAINT {
+        error!(
+            "Invalid decoded TXT data. Number of {} != {}",
+            BIP353_PREFIX, BIP353_PREFIX_COUNT_CONSTRAINT
+        );
+
+        return None;
+    }
+
+    Some(true)
+}
+
 fn get_by_key(tuple_vector: &[(&str, &str)], key: &str) -> Option<String> {
     tuple_vector
         .iter()
@@ -238,17 +261,9 @@ fn decode_txt_record(txt_data: Vec<u8>) -> Option<String> {
         })
         .ok()?;
 
-    if !decoded.to_lowercase().starts_with(BIP353_PREFIX) {
-        error!(
-            "Invalid decoded TXT data (doesn't begin with: {})",
-            BIP353_PREFIX
-        );
+    validate_txt_record(&decoded)?;
 
-        return None;
-    }
-
-    let prefix_query = format!("{}?", BIP353_PREFIX);
-    let (_, query_part) = decoded.split_once(&prefix_query)?;
+    let (_, query_part) = decoded.split_once("?")?;
 
     let query_params = querystring::querify(query_part);
 
--- a/libs/sdk-common/src/input_parser.rs
+++ b/libs/sdk-common/src/input_parser.rs
@@ -6,18 +6,20 @@ use ::bip21::Uri;
 use anyhow::{anyhow, Context, Result};
 use bitcoin::bech32;
 use bitcoin::bech32::FromBase32;
+use hickory_resolver::name_server::GenericConnector;
+use hickory_resolver::AsyncResolver;
 use log::error;
 use percent_encoding::NON_ALPHANUMERIC;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 use LnUrlRequestData::*;
 
-use trust_dns_resolver::config::*;
-use trust_dns_resolver::TokioAsyncResolver;
+use hickory_resolver::name_server::*;
 
 use crate::prelude::*;
 
 const USER_BITCOIN_PAYMENT_PREFIX: &str = "user._bitcoin-payment";
+const BOLT12_PREFIX: &str = "lno=";
 const LNURL_PAY_PREFIX: &str = "lnurl=";
 
 /// Parses generic user input, typically pasted from clipboard or scanned from a QR.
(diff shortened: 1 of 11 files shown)


libs/sdk-common/src/input_parser.rs
```diff
@@ -183,8 +191,21 @@ use crate::prelude::*;
 pub async fn parse(
     input: &str,
     external_input_parsers: Option<&[ExternalInputParser]>,
+    dns_resolver: Option<&AsyncResolver<GenericConnector<TokioRuntimeProvider>>>,
```
roeierez, line 194: Any reason to have this as an optional parameter? It won't be exposed to the bindings anyway so why not use our own inside the function and simplify the interface?

danielgranhao, line 194: Do you see a need to provide a "custom" dns resolver to `parse`? 

The interface would be leaner if a `parse` caller didn't need to provide one. We can create a new one on every call.

libs/sdk-common/src/input_parser.rs
```diff
@@ -183,8 +191,21 @@ use crate::prelude::*;
...
 ) -> Result<InputType> {
-    let input = input.trim();
+    let mut input = input.trim();
+
+    // Try to parse the destination as a bip353 address.
+    let input_str = if let Some(resolver) = dns_resolver {
```
danielgranhao, line 199: NIT: `str` in a name leads one to believe it's a string slice (`&str`) but in this case it's a `String`

libs/sdk-common/src/input_parser.rs
```diff
@@ -183,8 +191,21 @@ use crate::prelude::*;
 pub async fn parse(
     input: &str,
     external_input_parsers: Option<&[ExternalInputParser]>,
+    dns_resolver: Option<&AsyncResolver<GenericConnector<TokioRuntimeProvider>>>,
 ) -> Result<InputType> {
-    let input = input.trim();
+    let mut input = input.trim();
```
danielgranhao, line 196: NIT: it's generally better to avoid mut when possible. Logic is easier to follow when we know values cannot change. In this specific case, I usually prefer to use shadowing (creating a new variable with the same name).

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +218,47 @@ pub async fn parse(
...
+            Ok(decoded) => {
+                // The idea is to return from this parse only the BIP21 address and do logic on it inside the parse_core function.
+                // return Some(decoded);
+                if let Some((_, bolt12_address)) = decoded.split_once(BOLT12_PREFIX) {
+                    return Some(bolt12_address.to_string());
+                }
+
+                if let Some((_, lnurl)) = decoded.split_once(LNURL_PAY_PREFIX) {
+                    return Some(lnurl.to_string());
+                }
```
roeierez, line 251: The spec states that any value of the txt record should start with "bitcoin:" prefix so we should validate that before attempting to extract the relevant part.
If we are dealing here with either ln offer or lnurl then IMO we should validate it starts with either:
bitcoin:?lno=
bitcoin:?lnurl=

roeierez (changes requested):
Thank you @lorenzoronzani for the PR

libs/sdk-common/src/input_parser.rs
lorenzoronzani: I agree with you. I think that I misunderstood this comment to the liquid PR (https://github.com/breez/breez-sdk-liquid/pull/619/files/51cf0e48605d5df88f26a3b12a21e2332c89e607)

lorenzoronzani: I agree with you. I think that I misunderstood this comment to the liquid PR (https://github.com/breez/breez-sdk-liquid/pull/619/files/51cf0e48605d5df88f26a3b12a21e2332c89e607)

libs/sdk-common/src/input_parser.rs
JssDWt: @roeierez It's a BIP21 string, which could technically include both. I would say use the existing logic for BIP21 parsing here (with the `querystring::querify` part). Then check if the querystring contains bolt12. If it does: return bolt12. Then check whether it contains lnurl. If it does, return that.

libs/sdk-common/src/input_parser.rs
```diff
@@ -186,6 +195,19 @@ pub async fn parse(
 ) -> Result<InputType> {
     let input = input.trim();
 
+    let mut dns_resolvers_opts = ResolverOpts::default();
+    dns_resolvers_opts.validate = true;
```
JssDWt, line 199: I'm getting a line in the log:
`validate option is only available with 'dnssec' feature`

Tip:
Add the following to cargo.toml:
```
hickory-proto = { version = "0.24.2", features = ["dnssec"] }
hickory-resolver = { version = "0.24.2", features = ["dnssec"] }
```

However, when I try that on some BIP-353 domains I get: `proto error: self-signed dnskey is invalid`

Perhaps this is useful? https://github.com/TheBlueMatt/dnssec-prover

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +219,58 @@ pub async fn parse(
...
+                if !decoded.to_lowercase().starts_with(BIP353_PREFIX) {
+                    error!(
+                        "Invalid decoded TXT data (doesn't begin with: {})",
+                        BIP353_PREFIX
+                    );
+
+                    return None;
+                }
+
+                if let Some((_, bolt12_address)) =
```
JssDWt, line 253: I would suggest to return the full BIP21 string here. And have the rest of the input parser handle the resulting string. For example, I found this string in the wild: `bitcoin:?lno=lno1zrxq8pjw7qjlm68mtp7e3yvxee4y5xrgjhhyf2fxhlphpckrvevh50u0qffqdtuhyd77jpju7prz0qvkcm0xfzfpz844z9dwkaexuapm3e66vqsrxkhw43wz8wfa349nyl2wcsdp2u62ykz4lzuke37p75f4x286v6eqqva25lyzapjs7h424f7v65hv3k539558aqdztn24tjtkyqrrnr3f4vxwxhz0hhrkmzt4mz228sgs3tjl2vzhq0y3s4g8c65nvl35gyy2r833d6qaf64e2qwkey8zkm2q4yln60xgkqqsz8u6nrn984hru24yunnk0p093c&sp=sp1qqvwqe7yvuk9hwy6keczg2h0xajp6zeqzamzpcpfm0nrpgjr42yyq2q6hzkuldhpj5frduwjlzndf26rk9atzt49qwufun7drfw3c8yxypvk57ux5`

libs/sdk-common/src/input_parser.rs
roeierez: That makese sense, I didn't realize it is a bip21 format...
So the validation should be just parse using bip21::Uri and then extract the relevant part as @JssDWt suggested. I agree.

libs/sdk-common/src/input_parser.rs
roeierez: The problem I see here is that this doesn't look like a valid bip21 as it misses the address part (before the query params).

libs/sdk-common/src/input_parser.rs
roeierez: As I commented bellow it seems that in the spec the example is missing the bitcoin address (only contains the query params which should serve as fallback).
I suspect the bip21::Uri will fail to parse this string. Perhaps we should attempt to parse it using the crate and if fails try to extract the query params in case it starts with "bitcoin:?" ?

libs/sdk-common/src/input_parser.rs
hydra-yse: @roeierez The address field is not mandatory and most parsers support it, so it should be alright

libs/sdk-common/src/input_parser.rs
hydra-yse: Not sure about re-creating it on every call, since it's something that we expect to be frequently evoked from the user's side. 
I'm not against instantiating the resolver internally altogether, you could try taking a look at the `lazy_static` crate and starting it there.

hydra-yse: See https://github.com/breez/breez-sdk-greenlight/pull/1149#discussion_r1894510251

libs/sdk-common/src/input_parser.rs
JssDWt: @roeierez The mandatory address is a disputed property. I think it shouldn't be necessary to enforce that, especially as a client.

roeierez: I agree, just stated that our current bip21 parser (the crate we use) doesn't seem to support that.

libs/sdk-common/src/input_parser.rs
lorenzoronzani: I am not able to reproduce your log, but reading inside the documentation I found this message `To enable DNSSEC, one of the features dnssec-openssl or dnssec-ring must be enabled.` (https://crates.io/crates/hickory-resolver#:~:text=To%20enable%20DNSSEC%2C%20one%20of%20the%20features%20dnssec%2Dopenssl%20or%20dnssec%2Dring%20must%20be%20enabled.)

I tried to do that. Can you explain me better where did you find this log?

libs/sdk-common/src/input_parser.rs
lorenzoronzani: I did another test and confirm that the current bip21 parser returns that the provided bip21 is not a valid bip21uri.

I checked if the latest version of the crate still require the address and the answer is yes unfortunatly. Do you have some advice to proceed?

roeierez: I would go for the simple (and practical) approach now .Means parse the uri ourselves and the query params.
If contains offer return the offer string.
If contains lnurl return the lnurl string.

libs/sdk-common/src/input_parser.rs
JssDWt: The log `proto error: self-signed dnskey is invalid` I no longer get with `dnssec-ring`.

The log `validate option is only available with 'dnssec' feature` is also no longer there with `dnssec-ring`.

I think that resolves it!

libs/sdk-common/src/input_parser.rs
lorenzoronzani: I extracted the parsing functions to be easy reusable when we want to move the logic. I added the feature = "liquid" for the bolt12 part.

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,68 @@ pub async fn parse(
...
+                // return Some(decoded);
+
+                #[cfg(feature = "liquid")]
+                if let Some(bolt12_offer) = extract_bolt12_offer(&decoded) {
+                    return Some(bolt12_offer.to_string());
+                }
+
+                if let Some(lnurl) = extract_lnurl(&decoded) {
+                    return Some(lnurl.to_string());
+                }
```
roeierez, line 260: This doesn't take into account that there might be two query params one for offer and one for lnurl for example:
`bitcoin:?lno=<offer>&lnurl=<lnurl>`
I suggest:
1. First validate and parse the url which should be ended with the query params parsed in hand.
2. Then look according to priorities for the lno in the parsed params and then for the lnurl.

roeierez (changes requested):
Looks good.

libs/sdk-common/src/input_parser.rs
```diff
@@ -222,6 +314,11 @@ async fn parse_core(input: &str) -> Result<InputType> {
         };
     }
 
+    #[cfg(feature = "liquid")]
```
roeierez, line 317: I don't think we should use the `extract_bolt12_offer` & `extract_lnurl` inside `parse_core`.
Currently this specific bip21 format (without the bitcoin) is only intended to be used as part of the bip 353 spec so I feel more comfortable of using this logic inside the `bip353_parse` function and return from there either the offer or the lnurl.

Also for the internal parsing (and extracting the either offer or lnurl) in `bip353_parse` we can use a logic similar to this one: https://github.com/FGRibreau/querystring/blob/master/src/lib.rs#L38

1. First extract the query part (after the ? )
2. Then use that logic to create a hash table of key/value
3. Look for the "lno" first and after that for the "lnurl" and return the value from  the bip353_parse function.

What do you think about this suggestion @lorenzoronzani ?

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +226,69 @@ pub async fn parse(
...
 
+async fn bip353_parse(
+    input: &str,
+    dns_resolver: &AsyncResolver<GenericConnector<TokioRuntimeProvider>>,
+) -> Option<String> {
+    if let Some((local_part, domain)) = input.split_once('@') {
+        let dns_name = format!("{}.{}.{}", local_part, USER_BITCOIN_PAYMENT_PREFIX, domain);
+
+        // Query for TXT records of a domain
+        let txt_data = match dns_resolver.txt_lookup(dns_name).await {
```
roeierez, line 237: From the spec:

`Clients resolving Bitcoin payment instructions MUST ignore any TXT records at the same label which do not begin with (ignoring case) "bitcoin:". Resolvers encountering multiple "bitcoin:"-matching TXT records at the same label MUST treat the records as invalid and refuse to use any payment instructions therein.`

We need to make sure we validate that.

roeierez (changes requested):

libs/sdk-common/src/input_parser.rs
lorenzoronzani: Looks good

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,74 @@ pub async fn parse(
...
+) -> Option<String> {
+    let (local_part, domain) = input.split_once('@')?;
+    let dns_name = format!("{}.{}.{}", local_part, USER_BITCOIN_PAYMENT_PREFIX, domain);
+
+    // Query for TXT records of a domain
+    let txt_data = match dns_resolver.txt_lookup(dns_name).await {
+        Ok(records) => records
+            .iter()
+            .flat_map(|record| record.to_string().into_bytes())
+            .collect::<Vec<u8>>(),
```
danielgranhao, line 286: Is there a reason not to treat each record separately instead of combining all potential matches into a single byte array?

Validation would then consist of checking there is only a single one starting with `bitcoin:` and that record would be the one to proceed with.

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,74 @@ pub async fn parse(
...
+    if !decoded.to_lowercase().starts_with(BIP353_PREFIX) {
+        error!(
+            "Invalid decoded TXT data (doesn't begin with: {})",
+            BIP353_PREFIX
+        );
+
+        return None;
+    }
+
+    if !decoded.to_lowercase().matches(BIP353_PREFIX).count() == BIP353_PREFIX_COUNT_CONSTRAINT {
```
danielgranhao, line 238:
```suggestion
    if decoded.to_lowercase().matches(BIP353_PREFIX).count() != BIP353_PREFIX_COUNT_CONSTRAINT {
```

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,74 @@ pub async fn parse(
     Err(anyhow!("Unrecognized input type"))
 }
 
+fn validate_txt_record(decoded: &str) -> Option<bool> {
```
danielgranhao, line 228: This seems to always return either `None` or `Some(true)`. Probably can be replaced with `fn is_valid_txt_record(decoded: &str) -> bool`

danielgranhao (changes requested):

libs/sdk-common/src/input_parser.rs
lorenzoronzani: The idea behind this choice was to keep things separate to be easier to debug.

danielgranhao: I may be missing something, but this approach may lead to issues. Imagine there two records with the same label:

- `bitcoin:?lno=lno1qs...nai13`
- `some_random_text`

The spec states that `clients MUST ignore any TXT records at the same label which do not begin with (ignoring case) "bitcoin:"`, so in the example I provided, the 2nd record would be ignored, and the 1st would be parsed correctly into a BOLT12 offer. 

By concatenating every record into a byte array, which ends up being used as a single string, the above scenario would not be correctly parsed, as the decode logic would take in `bitcoin:?lno=lno1qs...nai13some_random_text`.

lorenzoronzani: Yes, exactly, you are right.
I am resolving now.

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,80 @@ pub async fn parse(
...
+    let (local_part, domain) = input.split_once('@')?;
+    let dns_name = format!("{}.{}.{}", local_part, USER_BITCOIN_PAYMENT_PREFIX, domain);
+
+    // Query for TXT records of a domain
+    let bip353_record = match dns_resolver.txt_lookup(dns_name).await {
+        Ok(records) => {
+            let decoded_records: Vec<String> = records
+                .iter()
+                .filter_map(|record| String::from_utf8(record.to_string().into_bytes()).ok())
+                .collect();
```
danielgranhao, line 289: Can be simplified

```suggestion
            let decoded_records: Vec<String> = records.iter().map(|r| r.to_string()).collect();
```

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,80 @@ pub async fn parse(
...
+        error!(
+            "Invalid decoded TXT data. Number of {} != {}",
+            BIP353_PREFIX, BIP353_PREFIX_COUNT_CONSTRAINT
+        );
+
+        return None;
+    }
+
+    return bip353_record.first().map(|record| record.to_string());
+}
```
danielgranhao, line 274: Consider the following alternative, which IMO is simpler

- Drops intermediate value `bip353_prefix_counter`
- Drops prefix count constant. It isn't foreseeable that this value changes and it isn't referenced in several places, so I don't think it adds value
- Uses into_iter to avoid the map at the end

```suggestion
fn extract_bip353_record(records: Vec<String>) -> Option<String> {
    let bip353_record = records
        .into_iter()
        .filter(|record| validate_bip353_record(record))
        .collect::<Vec<String>>();

    if bip353_record.len() > 1 {
        error!(
            "Invalid decoded TXT data. Multiple records found ({})",
            bip353_record.len()
        );
        return None;
    }

    bip353_record.first().cloned()
}
```

libs/sdk-common/src/lib.rs
```diff
@@ -18,6 +18,7 @@ mod utils;
 // use a different version, the Into traits defined here would not be usable by them
 // (e.g. impl From<bip32::Error> for LnUrlError)
 pub use bitcoin;
+pub use hickory_resolver;
```
danielgranhao, line 21: I don't think we need this. This would mainly be useful if callers needed to handle `hickory_resolver`-specific constructs, which isn't the case.

libs/sdk-core/src/lib.rs
```diff
@@ -196,5 +196,6 @@ pub use breez_services::{
 pub use chain::RecommendedFees;
 pub use lsp::LspInformation;
 pub use models::*;
+pub use sdk_common::hickory_resolver;
```
danielgranhao, line 199: Same, I don't think we need this

libs/sdk-common/src/input_parser.rs
```diff
@@ -197,6 +225,80 @@ pub async fn parse(
...
+fn parse_bip353_record(bip353_record: String) -> Option<String> {
+    let (_, query_part) = bip353_record.split_once("?")?;
+
+    let query_params = querystring::querify(query_part);
+
+    return get_by_key(&query_params, BOLT12_PREFIX)
+        .or_else(|| get_by_key(&query_params, LNURL_PAY_PREFIX));
+}
+
+fn validate_bip353_record(decoded: &str) -> bool {
```
danielgranhao, line 244: The following name is more idiomatic and lets callers more easily understand the function's behavior.

```suggestion
fn is_valid_bip353_record(decoded: &str) -> bool {
```

danielgranhao (commented):
Sorry for nitpicking 😄 

Also, please take into account the failed clippy check. There's some unnecessary `returns` that it complains about.

lorenzoronzani:
Oh no, thanks for nitpicking, I'm here to improve my skills and having someone to give me advice is a nice pro 😀.

danielgranhao (approved):
LGTM - we should just keep in mind BOLT12 offers behind BIP353 addresses will only be successfully parsed when we move BOLT12 parsing into the sdk-common.

hydra-yse (approved):
LGTM

roeierez (approved):
LGTM

PR #1153, opened 2025-01-06
Update expected txs in test

dangeross:
In `test_address_transactions` change the expected result. Looks like the mempool.space results are now block height ascending


JssDWt (approved):
We should really not be calling remote endpoints in unit tests. But for now this is good. LGTM.

roeierez:
> In `test_address_transactions` change the expected result. Looks like the mempool.space results are now block height ascending

In their docs they mention that the sorting order is the newest first: https://mempool.space/docs/api/rest#get-address-transactions

danielgranhao:
I noticed the test failing as well. The order can change. Meaning that the test sometimes fails and other times succeeds. Not sure if we are ok with addition trait implementations just for tests, but if so, the solution could be like proposed [here](https://github.com/breez/breez-sdk-greenlight/commit/acbcd3cdfd78cb27dd6c922b12f7bd527f6c7c0c).

roeierez (approved):

danielgranhao:
Closing in favor of #1160

PR #1156, opened 2025-01-09
Expose fn to get invoice pubkey

dangeross:


JssDWt:
Note after a rust-lightning update to 0.0.123, we can use `get_payee_pub_key` on `Bolt11Invoice` instead.

roeierez (approved):

PR #1170, opened 2025-02-18
WASM: sdk-common

dangeross:
This PR updates the `sdk-common` crate to be compilable to the `wasm32-unknown-unknown` target. 

- Exchange the use of the `tonic` transport feature in WASM for `tonic_web_wasm_client`
- Update async traits to be WASM compatible
- Update tests to be WASM compatible including adding a `mockito` alternative for WASM server mocking
- Add CI tests for WASM and `sdk-common` as a WASM dependency check
- Add `sdk-macros` crate to simplify WASM/non-WASM use
- Add DNS-over-HTTPS lookup for WASM

Fixes: https://github.com/breez/breez-sdk-liquid/issues/745


roeierez:
> * BIP353: Would a Breez API to do DNS TXT lookups be acceptable?

There is DNS over http protocol and services. https://developers.google.com/speed/public-dns/docs/doh
I think google has a public endpoint also.

roeierez (commented):
Thanks Ross for this PR! It really allows us to see what are the main issues and requires us to discuss an approach before we continue with the higher levels.
One thing we should think and discuss is how to minimize the wasm specific code changes and how to isolate it as much as possible (IMO even if we have to add more "duplicate" code).
I still don't have a silver bullet here but one think worth to consider is if we treat the SDK common as internal (not exposing structures to the browser) we might avoid wasmify all the structures.

dangeross:
Can test by cloning my [WASM playground](https://github.com/dangeross/wasm-playground)

libs/sdk-common/src/fiat.rs
```diff
@@ -1,14 +1,16 @@
...
 use serde::{Deserialize, Serialize};
 
 use crate::grpc::RatesRequest;
 use crate::prelude::BreezServer;
 use crate::with_connection_retry;
 
 /// Trait covering fiat-related functionality
-#[tonic::async_trait]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```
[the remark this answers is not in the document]
dangeross: Yes, tested with both CLIs

libs/sdk-common/src/buy/mod.rs
```diff
@@ -1,8 +1,10 @@
 use anyhow::Result;
+use async_trait::async_trait;
 
 pub mod moonpay;
 
-#[tonic::async_trait]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```
JssDWt, line 7: How does this work? Why does the future need `Send` in non-wasm, but not in wasm?

libs/sdk-common/src/utils/rest_client.rs
```diff
@@ -1,15 +1,20 @@
...
 pub fn get_reqwest_client() -> Result<reqwest::Client, ServiceConnectivityError> {
-    reqwest::Client::builder()
+    #[cfg(not(target_arch = "wasm32"))]
+    let client = reqwest::Client::builder()
         .timeout(Duration::from_secs(30))
         .build()
-        .map_err(Into::into)
+        .map_err(Into::into);
+    #[cfg(target_arch = "wasm32")]
+    let client = reqwest::Client::builder().build().map_err(Into::into);
```
JssDWt, line 16: wasm doesn't have request timeouts?

libs/sdk-common/justfile
```diff
@@ -0,0 +1,18 @@
+init:
+    cargo install wasm-pack
+
+[linux]
+wasm-clippy:
+    cargo clippy --target=wasm32-unknown-unknown -- -D warnings -A clippy::uninlined-format-args
+
+[macos]
+wasm-clippy:
+    AR=/usr/local/opt/llvm/bin/llvm-ar CC=/usr/local/opt/llvm/bin/clang cargo clippy --target=wasm32-unknown-unknown -- -D warnings -A clippy::uninlined-format-args
```
JssDWt, line 10: I think instead of using a justfile we could also do something like this in the `Makefile`


```
UNAME := $(shell uname)

ifeq ($(UNAME), Darwin)
    CLANG_PREFIX += AR=/usr/local/opt/llvm/bin/llvm-ar CC=/usr/local/opt/llvm/bin/clang
endif
```

And then

```
wasm-clippy:
    $(CLANG_PREFIX) cargo clippy --target=wasm32-unknown-unknown -- -D warnings -A clippy::uninlined-format-args
```

Depending on preference to install `just` or not.

JssDWt (commented):
Great milestone this one!

In general I think we should try keeping the wasm feature flags as compact as possible. Like moving things that need it to their own file. Or creating our own macros for testing or model derivations for example.

.github/workflows/main.yml
```diff
@@ -195,7 +200,9 @@ jobs:
 
       - name: Run sdk-common tests
         working-directory: libs/sdk-common
-        run: cargo test --features liquid
+        run: |
+          cargo test
+          wasm-pack test --headless --firefox
```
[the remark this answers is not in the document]
dangeross: Done [68dc4a1](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/68dc4a1b1568ae4906edcc626d4d95ab06752e9b)

libs/sdk-common/Cargo.toml
```diff
@@ -6,44 +6,74 @@ version.workspace = true
...
-url = "2.5.0"
-elements = { version = "0.25.0", optional = true }
-urlencoding = { version = "2.1.3" }
-percent-encoding = "2.3.1"
-hickory-resolver = { version = "0.24.2", features = ["dnssec-ring"] }
-lazy_static = "1.5.0"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2.14", features = ["js"] }
+prost = "^0.13"
```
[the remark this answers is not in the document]
dangeross: This caused dependency issues building sdk-core for me as we match the gl_client version

libs/sdk-common/Cargo.toml
```diff
@@ -6,44 +6,74 @@ version.workspace = true
...
 thiserror = { workspace = true }
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["sync"] }
+url = "2.5.0"
+urlencoding = { version = "2.1.3" }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+hickory-resolver = { version = "0.24.2", features = ["dnssec-ring"] }
+prost = { workspace = true }
+mockito = { workspace = true }
```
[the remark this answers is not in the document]
dangeross: Yeah, removed [68dc4a1](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/68dc4a1b1568ae4906edcc626d4d95ab06752e9b)

Makefile
```diff
@@ -6,10 +6,15 @@ fmt:
 
 clippy:
 	# Explicitly allow clippy::uninlined-format-args lint because it's present in the generated breez_sdk.uniffi.rs
+	cd libs/sdk-common && just wasm-clippy
 	cd libs && cargo clippy -- -D warnings -A clippy::uninlined-format-args
 	cd libs && cargo clippy --tests -- -D warnings -A clippy::uninlined-format-args
 	cd tools/sdk-cli && cargo clippy -- -D warnings
 
+test:
+	cd libs/sdk-common && just wasm-test
```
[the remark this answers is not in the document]
dangeross: Done [68dc4a1](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/68dc4a1b1568ae4906edcc626d4d95ab06752e9b)

libs/sdk-common/Cargo.toml
```diff
@@ -6,44 +6,74 @@ version.workspace = true
...
 querystring = "1"
-prost = { workspace = true }
 regex = { workspace = true }
 reqwest = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum_macros = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["sync"] }
```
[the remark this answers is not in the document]
dangeross: It's all sdk-common needs yes

libs/sdk-common/src/buy/mod.rs
dangeross: Its a WASMism: https://github.com/rustwasm/wasm-bindgen/issues/2833

libs/sdk-common/src/utils/rest_client.rs
dangeross: Not in reqwest's wasm client

libs/sdk-common/justfile
dangeross: Done [68dc4a1](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/68dc4a1b1568ae4906edcc626d4d95ab06752e9b)

libs/sdk-common/src/utils/rest_client.rs
JssDWt: I think a `reqwest` update can fix that
https://github.com/seanmonstar/reqwest/pull/1760

libs/sdk-common/mock_service_worker.js
```diff
@@ -0,0 +1,93 @@
...
+            const req = indexedDB.open("http_mock", 1);
+            req.onsuccess = (event) => rs(event.target.result);
+            req.onerror = rj;
+            req.onupgradeneeded = (event) => {
+                db = event.target.result;
+                db.createObjectStore("mocks", { keyPath: "nonce" });
+            };
+        })
+    }
+}
```
danielgranhao, line 17: When running the tests in non-headless mode, I'm getting an error after running the tests for the first time:

On Chrome:
```
DataError: Failed to execute 'get' on 'IDBObjectStore': No key or key range specified.
```
On Firefox:
```
DataError: No valid key or key range specified.
```

If I delete browser caches, cookies, etc, then it works once again.

I assume the culprit is failing to the get the db here.

danielgranhao (commented):
Awesome work! Found a couple of issues with tests, otherwise is looking good

libs/sdk-common/Makefile
```diff
@@ -0,0 +1,24 @@
+UNAME := $(shell uname)
+
+ifeq ($(UNAME), Darwin)
+	CLANG_PREFIX += AR=/usr/local/opt/llvm/bin/llvm-ar CC=/usr/local/opt/llvm/bin/clang
+endif
```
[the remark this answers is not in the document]
dangeross: Fixed [17c1765](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/17c17655d55aad750193c96d879ecddf49cec48c)

libs/sdk-common/Makefile
```diff
@@ -0,0 +1,24 @@
...
+	cargo clippy -- -D warnings -A clippy::uninlined-format-args
+
+cargo-test:
+	cargo test
+
+wasm-clippy:
+	$(CLANG_PREFIX) cargo clippy --target=wasm32-unknown-unknown -- -D warnings -A clippy::uninlined-format-args
+
+wasm-test:
+	$(CLANG_PREFIX) wasm-pack test --headless --firefox
```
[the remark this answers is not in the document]
dangeross: Added docs and fixed tests [17c1765](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/17c17655d55aad750193c96d879ecddf49cec48c)

Can you pull and test safari for me? I get errors starting the tests. I added `make wasm-test-safari`
```
Running headless tests in Safari on `http://127.0.0.1:54724/`
Try find `webdriver.json` for configure browser's capabilities:
Ok
driver status: signal: 9 (SIGKILL)                
Error: http://127.0.0.1:54724/session: status code 500
```

libs/sdk-common/mock_service_worker.js
dangeross: I can't reproduce this on chrome/firefox, but will keep testing

libs/sdk-common/Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+wasm-test-chrome:
+	$(CLANG_PREFIX) wasm-pack test --headless --chrome
```
danielgranhao, line 27: Annoyingly, to get chrome to work in headless mode, on Mac I also had to `brew install chromedriver`. Also, after install it gets flagged by the OS (shows a popup asking to delete a file the first time it's used), and one needs to manually go to Settings > Privacy & Security > "Allow use of chromedriver".

I didn't mention this before because I thought the error I saw was ocurring due to the tests failing.

For reference, this was the error:
```
Running headless tests in Chrome on `http://127.0.0.1:64837/`
Try find `webdriver.json` for configure browser's capabilities:
Ok
driver status: signal: 9 (SIGKILL)                
driver stdout:
    Starting ChromeDriver 114.0.5735.90 (386bc09e8f4f2e025eddae123f36f6263096ae49-refs/branch-heads/5735@{#1052}) on port 64837
    Only local connections are allowed.
    Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
    ChromeDriver was started successfully.

Error: http://127.0.0.1:64837/session/77bd929c9d22f9a33d5b6fe3492cf811/url: status code 404
error: test failed, to rerun pass `--lib`

```

libs/sdk-common/Makefile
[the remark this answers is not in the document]
danielgranhao: I get the same error when trying to run headless. Running on the browser works fine.

libs/sdk-common/Makefile
danielgranhao: Would be nice to briefly mention this prerequisite in the readme as well.

danielgranhao: I already had chrome installed, but according to some online reports the error was related to a mismatch between the chromedriver and Chrome itself? Anyways, installing chromedriver through brew did solve the issue

libs/sdk-common/mock_service_worker.js
danielgranhao: Thanks. On my machine the behavior is the same regardless of the browser. I spent some time trying to fix it but didn't manage to. IMO it isn't a blocker as it should be caused by this mock and it doesn't affect headless runs.

libs/sdk-common/src/test_utils/mock_server_wasm.rs
```diff
@@ -0,0 +1,101 @@
...
+    url_map: Mutex<HashMap<String, String>>,
+}
+
+#[allow(dead_code)]
+pub(crate) struct ServerOpts {
+    pub host: &'static str,
+    pub port: u16,
+}
+
+impl MockServer {
```
roeierez, line 34: Something to think of regarding the MockServer. Currently we are using a Mock server for lnurl testing. 
We are also dynamically replacingthe urls with a `maybe_replace_host_with_mock_test_host` function tha is only available for testing and registering the expected result on the mock server.
It also forces us to implement this service worker for wasm.
Since all lnurl requests are going through the function `get_parse_and_log_response` perhaps we should just create a testing version of it that returns the expected result and get rid of the mock server?
If we are able to do it it will remove a lot of specific WASM code and also the tests will use a lot of shared code.

libs/sdk-common/src/utils/rest_client.rs
```diff
@@ -18,7 +15,9 @@ pub async fn post_and_log_response(
 ) -> Result<String, ServiceConnectivityError> {
     debug!("Making POST request to: {url}");
 
-    let mut req = get_reqwest_client()?.post(url);
+    let mut req = get_reqwest_client()?
+        .post(url)
+        .timeout(Duration::from_secs(30));
```
danielgranhao, line 20: Nit: timeout duration could be a const

libs/sdk-common/webdriver.json
```diff
@@ -0,0 +1,8 @@
+{
+  "goog:chromeOptions": {
+    "args": [
+      "--use-fake-device-for-media-stream",
+      "--use-fake-ui-for-media-stream"
+    ]
+  }
+}
```
danielgranhao, line 8: Why is this needed? Is it related to the mock server?

libs/sdk-macros/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "sdk-macros"
+edition = "2021"
+version.workspace = true
```
danielgranhao, line 4: Maybe we should publish this to crates.io? It will be pretty useful for the boltz client crate as well :)

danielgranhao (commented):
Looking good! Left some comments. Also, as @roeierez proposed, I agree it would be nice to get rid of the mock server.

libs/sdk-common/src/breez_server.rs
```diff
@@ -16,34 +13,30 @@ use crate::grpc::payment_notifier_client::PaymentNotifierClient;
 use crate::grpc::signer_client::SignerClient;
 use crate::grpc::support_client::SupportClient;
 use crate::grpc::swapper_client::SwapperClient;
+use crate::grpc::transport::{GrpcClient, Transport};
 use crate::grpc::{ChainApiServersRequest, PingRequest};
 use crate::prelude::{ServiceConnectivityError, ServiceConnectivityErrorKind};
 use crate::with_connection_retry;
 
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
```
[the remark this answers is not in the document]
dangeross: Seems `proc_macro_attribute` can not be applied to statements/expressions. We could configure something like `#[cfg(is_wasm)]` in the build, but I think it only applies to the crate

libs/Cargo.toml
```diff
@@ -29,8 +31,8 @@ mockito = "1"
 once_cell = "1"
 prost = "^0.11"
 regex = "1.8.1"
-# Pin the reqwest dependency until macOS linker issue is fixed: https://github.com/seanmonstar/reqwest/issues/2006
-reqwest = { version = "=0.11.20", features = ["json"] }
+# Pin to commit until 0.12.12 is released
```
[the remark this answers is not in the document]
dangeross: Updated [a9a1c18](https://github.com/breez/breez-sdk-greenlight/pull/1170/commits/a9a1c18b4f003f2c0341cd77f7950a265ce709e4)

roeierez (approved):
LGTM with the additional replaceable rest client PR.

danielgranhao (approved):
Same, approving with #1175

PR #1175, opened 2025-02-28
Introduce a replaceable RestClient

dangeross:
This PR:
- Introduces a RestClient trait (with ReqwestRestClient and MockRestClient implementations)
- Allows external REST services (e.g. mempool.space, LNURL services) to be mocked for testing
- Removes the mockito dependency, simplifies WASM testing and removes URL injecting for test purposes


libs/sdk-common/src/input_parser.rs
```diff
@@ -33,6 +33,7 @@ const BIP353_PREFIX: &str = "bitcoin:";
 ///
 /// #[tokio::main]
 /// async fn main() {
+//
```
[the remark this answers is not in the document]
dangeross: Fixed [b3fff92](https://github.com/breez/breez-sdk-greenlight/pull/1175/commits/b3fff92987a3b1033193b567e897e9c9ab9432a9)

libs/sdk-common/src/input_parser.rs
```diff
@@ -188,6 +186,15 @@ const BIP353_PREFIX: &str = "bitcoin:";
 pub async fn parse(
     input: &str,
     external_input_parsers: Option<&[ExternalInputParser]>,
+) -> Result<InputType> {
+    let rest_client: Arc<dyn RestClient> = Arc::new(ReqwestRestClient::new()?);
+    parse_with_rest_client(rest_client, input, external_input_parsers).await
+}
+
+pub async fn parse_with_rest_client(
+    rest_client: Arc<dyn RestClient>,
```
JssDWt, line 195: The docs for the reqwest client say
> You do not have to wrap the Client in an [Rc] or [Arc](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) to reuse it, because it already uses an [Arc](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) internally.

So nit: I think functions can use a borrowed rest client rather than an Arc. That saves clones on passing it to the next function.

libs/sdk-common/src/input_parser.rs
```diff
@@ -469,6 +472,9 @@ fn ln_address_decode(ln_address: &str) -> Result<(String, String, String)> {
         // It is safe to downcase the domains since they are case-insensitive.
         // https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2
         let domain = split[1].to_lowercase();
+        if !domain.contains('.') {
```
JssDWt, line 475: What caused you to add this here in this PR?

libs/sdk-common/src/lnurl/specs/withdraw.rs
```diff
@@ -34,9 +36,11 @@ pub async fn validate_lnurl_withdraw(
 
     // Send invoice to the LNURL-w endpoint via the callback
     let callback_url = build_withdraw_callback_url(&req_data, &invoice)?;
-    let callback_url = maybe_replace_host_with_mock_test_host(callback_url)?;
 
-    let withdraw_status = match get_parse_and_log_response(&callback_url, false).await {
+    let response = rest_client
+        .get_and_log_response(&callback_url, false)
+        .await?;
+    let withdraw_status = match parse_json(&response) {
```
JssDWt, line 43: The timeout branch in this match statement is no longer correct, since a timeout would come from the `get_and_log_response` call above.

libs/sdk-core/src/chain.rs
```diff
@@ -398,44 +433,110 @@ mod tests {
...
 
-        let ms = RedundantChainService::from_base_urls(vec![
-            "https://mempool-url-unreachable.space/api/".into(),
-            "https://another-mempool-url-unreachable.space/api/".into(),
-            "https://mempool.emzy.de/api/".into(),
-        ]);
+        let ms = RedundantChainService::from_base_urls(
+            rest_client,
+            vec![
+                "https://mempool-url-unreachable.space/api/".into(),
```
JssDWt, line 503: great that this test now doesn't do actua network requests!

JssDWt (commented):
Looking good!
I like the way the responses can be set in the mock client. I very much like that we're not calling mempool.space in the tests anymore.

Some nits.

libs/sdk-common/src/input_parser.rs
dangeross: The MockRestClient was picking up a request to get LNURL info for `03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3zzz@sdfsffs` (in the node_id parsing trst). I think TLDs need a dot right?

libs/sdk-common/src/input_parser.rs
dangeross: Is this what you had in mind? [0f4633d](https://github.com/breez/breez-sdk-greenlight/pull/1175/commits/0f4633d245bdc594347cb9b3f9304955bab4104e)

libs/sdk-common/src/input_parser.rs
JssDWt: Technically it could be `localhost`, or other domains in the local network. So this failed before because the network call failed?

dangeross: > So this failed before because the network call failed?

Yes. Ok, it maybe makes sense to remove this then and add a mock response for this case

libs/sdk-common/src/input_parser.rs
JssDWt: Not really. I had in mind 
```
pub async fn validate_lnurl_withdraw(rest_client: &dyn RestClient)

... 

let mock_rest_client = MockRestClient::new();
validate_lnurl_withdraw(&mock_rest_client, withdraw_req, invoice);
```

But coding it up it would require generics, to support cloning the rest client. It would get quite some feet in the earth. So let's stick with the arcs.

libs/sdk-common/src/input_parser.rs
```diff
@@ -188,6 +186,15 @@ const BIP353_PREFIX: &str = "bitcoin:";
 pub async fn parse(
     input: &str,
     external_input_parsers: Option<&[ExternalInputParser]>,
+) -> Result<InputType> {
+    let rest_client: Arc<dyn RestClient> = Arc::new(ReqwestRestClient::new()?);
+    parse_with_rest_client(rest_client.as_ref(), input, external_input_parsers).await
+}
+
+pub async fn parse_with_rest_client<C: RestClient + ?Sized>(
```
roeierez, line 194: It seems that this function only exists so we can test the input parser by passing a rest client?
Perhaps we can make it with pub(crate) visibility ?

libs/sdk-common/src/utils/rest_client.rs
```diff
@@ -1,84 +1,109 @@
...
-    url: &str,
-    body: Option<String>,
-) -> Result<String, ServiceConnectivityError> {
-    debug!("Making POST request to: {url}");
+    /// Makes a POST request, and logs on DEBUG.
+    /// ### Arguments
+    /// - `url`: the URL on which POST will be called
+    /// - `headers`: the optional POST headers
+    /// - `body`: the optional POST body
+    async fn post_and_log_response(
```
roeierez, line 29: At the trait level I think it is better to make the function naming/definitions more simplified and less verbose.
IMO simply get & post are good options and we can keep the logging in the implementation without the need to express it in the function name.
I also think the result should be body and status code, letting the caller the option to enforce_status_check or not. I have no objection to add another function or two that gets the rest client as parameter and have more options just feel that the interface level (especially  because it is also exposed out of this trait) should be clean and minimal.

roeierez (changes requested):
Looks good. A few suggestions to improve the rest client interface

libs/sdk-common/src/input_parser.rs
dangeross: It will be used also in the Nodeless SDK, so needs to be pub

libs/sdk-common/src/utils/rest_client.rs
dangeross: I've renamed the functions and included the status code in the response [a395257](https://github.com/breez/breez-sdk-greenlight/pull/1175/commits/a395257c1544b989ec7ecb2b0c8487f3bd910229). 

To expose the client internals we probably need to add another trait, I don't see any use at the moment. Should we add it @roeierez ?

roeierez (approved):
LGTM

JssDWt (approved):

danielgranhao (approved):
LGTM - left just a couple of nits

PR #1180, opened 2025-03-13
fix refund tx id conflicts

JssDWt:
If a swap refund tx id is already in the database, the insert can be ignored. Fixes https://github.com/breez/breez-sdk-relai/issues/159

### DIFF
--- a/libs/sdk-core/src/persist/sync.rs
+++ b/libs/sdk-core/src/persist/sync.rs
@@ -151,12 +151,11 @@ impl SqliteStorage {
         // sync remote swap_refunds table
         tx.execute(
             "
-        INSERT INTO sync.swap_refunds
+        INSERT OR IGNORE INTO sync.swap_refunds
          SELECT
           bitcoin_address,
           refund_tx_id
-         FROM remote_sync.swap_refunds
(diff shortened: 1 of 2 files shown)


erdemyerebasmaz (approved):
LGTM!

libs/sdk-core/src/persist/sync.rs
```diff
@@ -156,7 +156,8 @@ impl SqliteStorage {
           bitcoin_address,
           refund_tx_id
          FROM remote_sync.swap_refunds
-         WHERE bitcoin_address NOT IN (SELECT bitcoin_address FROM sync.swap_refunds);",
+         WHERE bitcoin_address NOT IN (SELECT bitcoin_address FROM sync.swap_refunds)
+         ON CONFLICT DO NOTHING;",
```
erdemyerebasmaz, line 160: Actually, I have one comment: Isn't the `WHERE ... NOT IN` clause redundant now?

JssDWt: I agree. Removed that line in c80cb9f8dd3a19bd2a0114cbb6a8e9cbea8b616f

PR #1181, opened 2025-03-16
Taproot swaps (V2)

JssDWt:
This is an alternative to #1176

There is a single interface to swaps: `BTCReceiveSwap`. The main logic for swap execution is inside this struct, with some divergent paths for either segwit and taproot swaps where applicable. Most of the logic for the segwit and taproot swaps are the same, so they can be bundled together like this.

The data model is the same as before, only additional chain information has been added to the SwapInfo table. This is mainly because the current data model only considered utxos, but it's useful to also consider spent outputs in circumstances where multiple outputs exist for the same swap address.

The invoice creation logic is different from the original version. Invoices are deleted when the amount no longer matches, or when the local liquidity no longer matches. A missing piece of the puzzle there is to allow re-registration of a payment in lspd.

### DIFF
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -13,13 +13,14 @@ use futures::{Future, Stream};
 use gl_client::credentials::{Device, Nobody};
 use gl_client::node;
 use gl_client::node::ClnClient;
+use gl_client::pb::cln::delinvoice_request::DelinvoiceStatus;
 use gl_client::pb::cln::listinvoices_invoices::ListinvoicesInvoicesStatus;
 use gl_client::pb::cln::listinvoices_request::ListinvoicesIndex;
 use gl_client::pb::cln::listpays_pays::ListpaysPaysStatus;
 use gl_client::pb::cln::listpeerchannels_channels::ListpeerchannelsChannelsState::*;
 use gl_client::pb::cln::listsendpays_request::ListsendpaysIndex;
 use gl_client::pb::cln::{
-    self, Amount, GetrouteRequest, GetrouteRoute, ListchannelsRequest,
+    self, Amount, DelinvoiceRequest, GetrouteRequest, GetrouteRoute, ListchannelsRequest,
     ListclosedchannelsClosedchannels, ListpaysPays, ListpeerchannelsChannels, ListsendpaysPayments,
     PreapproveinvoiceRequest, SendpayRequest, SendpayRoute, WaitsendpayRequest,
 };
@@ -1082,6 +1083,35 @@ impl NodeAPI for Greenlight {
             .await?
             .into_inner();
         Ok(res.bolt11)
+    }
+
+    async fn delete_invoice(&self, bolt11: String) -> NodeResult<()> {
+        let mut client = self.get_node_client().await?;
+        let invoice_request = cln::ListinvoicesRequest {
+            invstring: Some(bolt11),
+            ..Default::default()
+        };
+        let invoice_result = with_connection_retry!(client.list_invoices(invoice_request.clone()))
+            .await?
+            .into_inner();
+        let invoice_result = invoice_result.invoices.first();
+        let result = match invoice_result {
+            Some(result) => result,
+            None => return Ok(()),
+        };
+
+        let status = match result.status() {
+            ListinvoicesInvoicesStatus::Unpaid => DelinvoiceStatus::Unpaid,
+            ListinvoicesInvoicesStatus::Paid => DelinvoiceStatus::Paid,
+            ListinvoicesInvoicesStatus::Expired => DelinvoiceStatus::Expired,
+        };
+        with_connection_retry!(client.del_invoice(DelinvoiceRequest {
+            label: result.label.clone(),
+            status: status.into(),
+            desconly: Some(false),
+        }))
+        .await?;
+        Ok(())
     }
 
     async fn fetch_bolt11(&self, payment_hash: Vec<u8>) -> NodeResult<Option<FetchBolt11Result>> {
--- a/libs/sdk-core/src/greenlight/node_api.rs
+++ b/libs/sdk-core/src/greenlight/node_api.rs
@@ -51,6 +51,7 @@ use crate::bitcoin::{
 use crate::lightning::util::message_signing::verify;
 use crate::lightning_invoice::{RawBolt11Invoice, SignedRawBolt11Invoice};
 use crate::node_api::{CreateInvoiceRequest, FetchBolt11Result, NodeAPI, NodeError, NodeResult};
+use crate::persist::cache::NodeStateStorage;
 use crate::persist::db::SqliteStorage;
 use crate::persist::send_pays::{SendPay, SendPayStatus};
 use crate::{models::*, LspInformation};
@@ -1875,11 +1876,11 @@ impl NodeAPI for Greenlight {
         }
     }
 
-    async fn max_sendable_amount(
+    async fn max_sendable_amount<'a>(
         &self,
         payee_node_id: Option<Vec<u8>>,
         max_hops: u32,
-        last_hop_hint: Option<&RouteHintHop>,
+        last_hop_hint: Option<&'a RouteHintHop>,
     ) -> NodeResult<Vec<MaxChannelAmount>> {
         let mut client = self.get_node_client().await?;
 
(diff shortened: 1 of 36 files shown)


libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -194,384 +234,283 @@ impl BTCReceiveSwap {
...
             status: Some(SwapStatus::monitored()),
             ..Default::default()
-        })?)
-    }
-
-    pub(crate) fn list_refundables(&self) -> Result<Vec<SwapInfo>> {
-        Ok(self.persister.list_swaps(ListSwapsRequest {
-            status: Some(SwapStatus::refundable()),
+        })?;
+        let recent = self.list_swaps(ListSwapsRequest {
```
roeierez, line 289: So we are rescanning anyway all swaps until timeout? regardless of their status?  Just curious what have we been missing before this implementation.

roeierez (commented):
Looks really good! dropped some small comments and I still want to review the state resolution as it has caused issues in the past.

libs/sdk-core/src/swap_in/swap.rs
JssDWt: This is just to make sure we catch additional funds sent to the same address within a reasonable time.

libs/sdk-core/src/greenlight/node_api.rs
```diff
@@ -1065,6 +1066,35 @@ impl NodeAPI for Greenlight {
...
+            .into_inner();
+        let invoice_result = invoice_result.invoices.first();
+        let result = match invoice_result {
+            Some(result) => result,
+            None => return Ok(()),
+        };
+
+        let status = match result.status() {
+            ListinvoicesInvoicesStatus::Unpaid => DelinvoiceStatus::Unpaid,
+            ListinvoicesInvoicesStatus::Paid => DelinvoiceStatus::Paid,
```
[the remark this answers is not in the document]
JssDWt: Right! Done in 3a9f1a430ebd3f323a31d36eed5068684dc1a27d

libs/sdk-core/src/persist/swap.rs
```diff
@@ -442,214 +483,3 @@ impl SqliteStorage {
         })
     }
 }
-
-#[cfg(test)]
```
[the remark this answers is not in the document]
JssDWt: I will write new tests. It's the tests for the state transitions that no longer work, because state transitions are now handled inside `BTCReceiveSwap`, rather than on `SwapInfo` itself.

libs/sdk-core/src/swap_in/segwit.rs
```diff
@@ -0,0 +1,178 @@
...
+            let w = Witness::from_vec(witness);
+            input.witness = w;
+        }
+
+        Ok(tx)
+    }
+
+    pub async fn get_swap_payment(
+        &self,
+        _swap_info: &SwapInfo,
```
danielgranhao, line 144: Nit: can we get rid of this parameter?

libs/sdk-core/src/swap_in/segwit.rs
```diff
@@ -0,0 +1,178 @@
...
+            let witness: Vec<Vec<u8>> = vec![sigvec, vec![], serialize(&input_script)];
+
+            let w = Witness::from_vec(witness);
+            input.witness = w;
+        }
+
+        Ok(tx)
+    }
+
+    pub async fn get_swap_payment(
```
danielgranhao, line 142: Nit: this name makes it seem like we're getting some "payment" struct (like reading from db). Maybe `complete_swap_payment` or `complete_payment`?

danielgranhao (commented):
Did a first pass and looks good! Left some minor comments

libs/sdk-core/src/persist/transactions.rs
```diff
@@ -317,6 +317,7 @@ impl SqliteStorage {
     }
 
     /// Look up a modified open channel bolt11 by hash.
+    #[allow(unused)]
     pub(crate) fn get_open_channel_bolt11_by_hash(
```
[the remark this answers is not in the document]
JssDWt: It's still used in a test. Also I felt it should still be used. I'll look at where I can get the open channel bolt11 from the database instead of recreating it. Otherwise I might remove that functionality.

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -581,892 +520,641 @@ impl BTCReceiveSwap {
...
+            })
+            .collect();
 
-        // compare the expected and created swap address
-        let address = crate::bitcoin::Address::p2wsh(&script, crate::bitcoin::Network::Bitcoin);
-        let address_str = address.to_string();
-        assert_eq!(address_str, expected_address);
+        if refundable_utxos.is_empty() {
+            return SwapStatus::Completed;
+        }
```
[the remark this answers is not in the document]
JssDWt: Looking at the current implementation, there's 2 options for that scenario:
- if the timelock passed for the first confirmed output, it returns 'Completed'
- if the timelock has not yet passed for the first confirmed output, it returns 'Redeemed'

I think Redeemed is the right option here, what do you think?

libs/sdk-core/src/swap_in/segwit.rs
```diff
@@ -0,0 +1,178 @@
...
+            version: 2,
+            lock_time: PackedLockTime::ZERO,
+            input: utxos
+                .iter()
+                .map(|utxo| {
+                    Ok(TxIn {
+                        witness: Witness::from_vec(vec![
+                            [1; 73].to_vec(),
+                            Vec::new(),
+                            [1; 100].to_vec(),
```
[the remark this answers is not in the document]
JssDWt: Done in f247c006b595eecf0c324b3e72ba507e799ba31d

libs/sdk-core/src/swap_in/segwit.rs
```diff
@@ -0,0 +1,178 @@
...
+        min_confirmation: u32,
+        current_tip: u32,
+    ) -> u32 {
+        let confirmations = current_tip.saturating_sub(min_confirmation);
+        (swap_info.lock_height as u32).saturating_sub(confirmations)
+    }
+
+    pub fn create_fake_refund_tx(
+        &self,
+        _swap_info: &SwapInfo,
```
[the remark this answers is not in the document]
JssDWt: Done in 4d90fae76c825d677d009a4c3b2f2d53513d70c3

libs/sdk-core/src/swap_in/segwit.rs
JssDWt: I went for `payout_swap`: 9fec848ab67f58da6bf6ec7678355509e28ab60c

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -581,892 +545,1100 @@ impl BTCReceiveSwap {
...
-}
+        // There are refundable utxos.
+        SwapStatus::Refundable
+    }
 
-#[cfg(test)]
-mod tests {
-    use std::time::{SystemTime, UNIX_EPOCH};
-    use std::{sync::Arc, vec};
+    fn calculate_status_without_chain_data(
```
roeierez, line 699: In case use sent above the limit, should we consider the swap as refundable immediately? Currently it seems to be refundable only after the timeout?

libs/sdk-core/src/swap_in/swap.rs
```diff
@@ -581,892 +545,1100 @@ impl BTCReceiveSwap {
...
 }
 
-pub(crate) fn create_swap_keys() -> Result<SwapKeys> {
-    let priv_key = rand::thread_rng().gen::<[u8; 32]>().to_vec();
-    let preimage = rand::thread_rng().gen::<[u8; 32]>().to_vec();
-    Ok(SwapKeys { priv_key, preimage })
-}
+/// ReceiveSwapper private functions
+impl BTCReceiveSwap {
+    fn calculate_status(
```
roeierez, line 623: Very Nice! I went over the logic several times and didn't find any edge cases.

roeierez (commented):
Looks really good. Some small comments.

JssDWt:
Things fixed in the last few days:
- the `script` parameter is set to the refund script. In fact this is not enough to spend. It should probably be both scripts. I didn't yet find a good way to serialize the entire taptree.
- added some tests
- Fixed the cooperative refund transaction creation
- Fixed fee calculation in prepare_refund
- Return a proper error when utxos are still timelocked in the unilateral case
- Ensure refund transactions can be RBF'ed
- Ensure the status is still `Refundable` when the refund tx is unconfirmed

libs/sdk-core/src/swap_in/swap.rs
JssDWt: That makes sense for taproot swaps, yes. Will do.

JssDWt: Done in fdb402b4538b269ece53ed115e6748b4d8e808d6

PR #1184, opened 2025-03-21
Fix payment timeout not being applied to RPC calls

### LINKED ISSUE #838: Unresponsive node bypasses timeout setting
Customer reported that calling `sendPayment()` returned only 5 minutes later with

```
"RpcError { code: None, message: \"no response from lightningd\",
```

Our `Config::payment_timeout_sec`

https://github.com/breez/breez-sdk/blob/77f0b65ba340767de6e290b1a579ea5baf044a16/libs/sdk-core/src/models.rs#L457

should have kicked it.

We should apply this timeout also as a connection timeout when calling GL methods.

Sidharth-Singh10:
closes #838 
## Description
This PR addresses the issue where the configured payment_timeout_sec wasn't being properly applied to RPC calls to the Lightning node, causing calls to hang indefinitely rather than timing out as expected.

## Changes

Modified `send_payment` and `send_spontaneous_payment` methods in breez_services.rs to:

- Apply the configured timeout using tokio::time::timeout
- Return a specific SendPaymentError::PaymentTimeout error when operations exceed the timeout
- Ensure timeout handling for both regular and trampoline payment methods

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -11,7 +11,6 @@ use bitcoin::hashes::hex::ToHex;
 use bitcoin::hashes::{sha256, Hash};
 use bitcoin::util::bip32::ChildNumber;
 use chrono::Local;
-use futures::TryFutureExt;
 use gl_client::pb::incoming_payment;
 use log::{LevelFilter, Metadata, Record};
 use sdk_common::grpc;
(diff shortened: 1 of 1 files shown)


JssDWt:
Thanks for the PR!

The issue with lightning payments is
- if a payment is stuck, it may still succeed after a long time.
- you cannot retry them while another one for the same payment hash is in-flight

Returning an error for payment timeouts while the payment is in-flight can lead to 'strange' errors for users. Like retrying a payment, but getting an error that there is insufficient balance. Retrying a payment to an lnurl, and eventually having paid twice, because the first succeeds after a long time. 

So generally the payment timeout is more a 'pathfinding' timeout. After the timeout you will stop making new attempts. But there could be attempts that well exceed the payment timeout.

What do you think?

Sidharth-Singh10:
Rather than treating a timeout as a payment failure, we could present it to the user as "Payment processing in background" and provide a way to check the status later. 

we could keep the timeout for UI responsiveness, but change how we communicate this to users:
- Add persistent tracking of in-progress payments
- Implement a background monitoring service that periodically checks payment status
- Add proper notification when timed-out payments eventually succeed or fail

This would prevent the confusion of users attempting to pay again while a payment is still potentially in flight.

Sidharth-Singh10:
Rather than treating a timeout as a payment failure, we could present it to the user as "Payment processing in background" and provide a way to check the status later. 

we could keep the timeout for UI responsiveness, but change how we communicate this to users:
- Add persistent tracking of in-progress payments
- Implement a background monitoring service that periodically checks payment status
- Add proper notification when timed-out payments eventually succeed or fail

This would prevent the confusion of users attempting to pay again while a payment is still potentially in flight.

JssDWt:
> * Add persistent tracking of in-progress payments
> * Implement a background monitoring service that periodically checks payment status
> * Add proper notification when timed-out payments eventually succeed or fail

Is there anything you need in the SDK to make that happen?

Sidharth-Singh10:
apologies, I am a bit confused, could you please clarify the question?
I'm not sure if you're asking whether I need specific SDK features to implement this approach, or if you're asking whether this functionality should be built into the SDK itself.

JssDWt:
> apologies, I am a bit confused, could you please clarify the question? I'm not sure if you're asking whether I need specific SDK features to implement this approach, or if you're asking whether this functionality should be built into the SDK itself.

I'm asking whether you need changes in the SDK to support your use case?

You could call the send_payment function and have it execute in the background to keep your UI responsive. When it returns you can update your UI state based on that. 

There are events you can subscribe to as well: `PaymentSucceed`, `PaymentFailed`.
Meantime in `list_payments` the payment's status will be `Pending`.

I would suggest to keep the payment as `pending` until it is resolved fully by either the function call returning or receiving an event.

Sidharth-Singh10:
Got it!

JssDWt:
@Sidharth-Singh10 I'm closing this PR, let me know if you want to reopen it.

PR #1189, opened 2025-03-27
sdk-common: drop Send and Sync on Wasm

danielgranhao:
This PR drops the use of Send and Sync on Wasm through the use of `MaybeSend` and `MaybeSync`. It also switches all uses of Arc to Rc.


dangeross (approved):
Looks good

hydra-yse (approved):
LGTM

PR #1191, opened 2025-03-31
Strip debug info from cli app to improve compilation time

andrei-21:


JssDWt:
### without optimization:
full build: 49.54s
changing a line in sdk-core: 7.46s

### with optimization
full build: 43.43s
changing a line in sdk-core: 6.49s

JssDWt (approved):
Seems fine to me. You can always remove this if you need it.

tools/sdk-cli/Cargo.toml
```diff
@@ -16,4 +14,8 @@ serde_json = "1.0"
 tiny-bip39 = "1"
 tokio = { version = "1", features = ["rt-multi-thread"] }
 serde = { version = "1.0", features = ["derive"] }
-clap = { version = "4", features = ["derive"] }
\ No newline at end of file
+clap = { version = "4", features = ["derive"] }
+
+[profile.dev]
+debug = 0
+strip = "debuginfo"
```
roeierez, line 21: Intuitively it seems to me that debuginfo belongs to dev profile (othewise where elsewhere?).
A simple solution is to come up with a custom profile that a developer would have to include in the cargo build command?
Saying all that, I have no strong opinion here so only listed my thoughts.

roeierez (approved):

tools/sdk-cli/Cargo.toml
JssDWt: @andrei-21 what are your thoughts on this?

JssDWt (approved):

roeierez (approved):

PR #1192, opened 2025-04-01
Explicitly drop the shutdown sender to notify the receiver

andrei-21:
Note: The `Sender::send()` method is asynchronous and returns a future. Creating a future and not awaiting it does nothing. The code actually worked because the sender itself was later dropped. I suggest dropping it explicitly here to express intention.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -1513,7 +1513,7 @@ impl BreezServices {
                 };
 
                 debug!("shutting down signer");
-                _ = tx.send(());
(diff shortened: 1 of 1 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -1513,7 +1513,7 @@ impl BreezServices {
                 };
 
                 debug!("shutting down signer");
-                _ = tx.send(());
+                drop(tx); // Dropping the sender explicitly to notify the receiver.
```
roeierez, line 1516: Nice catch.

JssDWt (approved):

PR #1193, opened 2025-04-03
Update nodeless lightning dependency to 0.1.2

dangeross:


libs/sdk-common/Cargo.toml
```diff
@@ -15,8 +15,7 @@ elements = { version = "0.25.0", optional = true }
 hex = { workspace = true }
 lazy_static = "1.5.0"
 lightning = { workspace = true }
-# Update to 0.1.2 when it is released
-lightning-with-bolt12 = { package = "lightning", git = "https://github.com/lightningdevkit/rust-lightning", rev = "f80d82e835987e045928955d2c83baba86c445e9", default-features = false, features = [
+lightning-with-bolt12 = { package = "lightning", version = "0.1.2", default-features = false, features = [
```
roeierez, line 18: Just checking if this is compatible with greenlight?

roeierez: Or it is excluded anyway?

dangeross: It's only enabled when using the `liquid` feature (used by Nodeless SDK)

roeierez (approved):

danielgranhao (approved):

PR #1203, opened 2025-04-17
improve error on insufficient balance for trampoline payments

JssDWt:
The current flow when you send a trampoline payment and the user's balance is insufficient is:
- we attempt a trampoline payment
- it will return an error like `missing balance 43276289msat<50254000msat`
- we attempt a regular payment
- it will return `Ran out of routes to try after 1 attempt`

This PR improves this flow by introducing an 'InsufficientBalance' error.
If the trampoline payment fails with 'missing balance', return immediately an InsufficientBalance error to the user.

### DIFF
--- a/libs/sdk-core/src/breez_services.rs
+++ b/libs/sdk-core/src/breez_services.rs
@@ -325,6 +325,16 @@ impl BreezServices {
             {
                 Ok(res) => Some(res),
                 Err(e) => {
+                    if e.to_string().contains("missing balance") {
+                        debug!(
+                            "trampoline payment failed due to insufficient balance: {:?}",
+                            e
+                        );
+                        return Err(SendPaymentError::InsufficientBalance {
+                            err: "Trampoline payment failed".into(),
+                        });
+                    }
+
                     warn!("trampoline payment failed: {:?}", e);
                     None
                 }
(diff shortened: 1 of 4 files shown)


libs/sdk-core/src/breez_services.rs
```diff
@@ -325,6 +325,16 @@ impl BreezServices {
             {
                 Ok(res) => Some(res),
                 Err(e) => {
+                    if e.to_string().contains("missing balance") {
+                        debug!(
+                            "trampoline payment failed due to insufficient balance: {:?}",
+                            e
+                        );
+                        return Err(SendPaymentError::InsufficientBalance {
+                            err: "Insufficient balance for trampoline payment".into(),
```
dangeross, line 334: My only nit here would be it produces an error "Insufficient balance: Insufficient balance for trampoline payment". Maybe?
```suggestion
                            err: "Tampoline payment failed".into(),
```

JssDWt: Agreed, done!

dangeross (approved):
LGTM

erdemyerebasmaz (approved):

roeierez (approved):
LGTM

PR #1204, opened 2025-04-17
CI: update minimal ubuntu versions

JssDWt:
ubuntu 20.04 is deprecated

This will affect the glibc versions used in our distributions, meaning we support less linux machines.


dangeross (approved):
LGTM. I ran a test of the publish CI on this branch also: https://github.com/breez/breez-sdk-greenlight/actions/runs/14523229065

Issue #1128, opened 2024-11-25
flutter: Plugin with id 'kotlinx-serialization' not found.


JssDWt:
When adding breez sdk dependency to a new flutter project, I get the following error:

```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/Users/jessedewit/.pub-cache/git/breez-sdk-flutter-c50967946b5d09145ca256f094922e98dfb52c31/android/build.gradle' line: 27

* What went wrong:
A problem occurred evaluating project ':breez_sdk'.
> Plugin with id 'kotlinx-serialization' not found.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
A problem occurred configuring project ':breez_sdk'.
> Failed to notify project evaluation listener.
   > Cannot invoke method substring() on null object
   > compileSdkVersion is not specified. Please add it to build.gradle

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
==============================================================================

BUILD FAILED in 8s
Running Gradle task 'assembleDebug'...                              8,8s
Error: Gradle task assembleDebug failed with exit code 1
```

So then you have to manually add the kotlin serialization plugin:

`id 'org.jetbrains.kotlin.plugin.serialization' version "1.8.22" apply false`

Which leads to the following error:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':breez_sdk'.
> Could not create an instance of type com.android.build.api.variant.impl.LibraryVariantBuilderImpl.
   > Namespace not specified. Specify a namespace in the module's build file. See https://d.android.com/r/tools/upgrade-assistant/set-namespace for information about setting the namespace.

     If you've specified the package attribute in the source AndroidManifest.xml, you can use the AGP Upgrade Assistant to migrate to the namespace value in the build file. Refer to https://d.android.com/r/tools/upgrade-assistant/agp-upgrade-assistant for general information about using the AGP Upgrade Assistant.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
[...]

PR #1206, opened 2025-04-21
build(android): update Gradle config for Flutter plugin compatibility

erdemyerebasmaz:
Fixes https://github.com/breez/breez-sdk-greenlight/issues/1128

---

Supersedes 
- https://github.com/breez/breez-sdk-greenlight/pull/1129

---

This PR updates the Android Gradle configuration to align with AGP 8 and recent tooling requirements:
- Add `namespace` to module-level `build.gradle` to comply with AGP requirements.
  - Remove obsolete `AndroidManifest.xml` no longer needed after namespace migration.
- Update `compileSdk` and `targetSdk` versions to 35.
-  Add missing `kotlinx-serialization` plugin dependency to fix: 
   - #1128 
   - `Plugin with id 'kotlinx-serialization' not found.`
- Add local `.so` artifacts to `jniLibs` via `sourceSets` to avoid build cache conflicts.
- Fix duplicate JNA class issue by excluding one of the clashing JARs:
   ```
    Duplicate class com.sun.jna.**** found in modules jna-5.14.0.aar -> jetified-jna-5.14.0-runtime (net.java.dev.jna:jna:5.14.0) and jna-5.14.0.jar -> jetified-jna-5.14.0 (net.java.dev.jna:jna:5.14.0)
    ```

These changes resolve build issues and ensure compatibility with the latest Flutter plugin, Android toolchain, and both local and remote dependencies.

---

There will be a continuation PR for addressing issues Flutter plugin issues on iOS.

### DIFF
--- a/libs/sdk-flutter/android/build.gradle.production
+++ b/libs/sdk-flutter/android/build.gradle.production
@@ -55,8 +55,9 @@ android {
 
 dependencies {
     api "breez_sdk:bindings-android:$version"
-    implementation "net.java.dev.jna:jna:5.14.0@aar"
+    implementation("net.java.dev.jna:jna:5.14.0@aar") {
+        exclude group: 'net.java.dev.jna', module: 'jna'
+    }
     /* JSON serialization */
     implementation 'org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3'
-
 }
--- a/libs/sdk-flutter/android/build.gradle.production
+++ b/libs/sdk-flutter/android/build.gradle.production
@@ -49,6 +49,7 @@ android {
     }
 
     defaultConfig {
+        ndkVersion = flutter.ndkVersion
         minSdk = 24
     }
 }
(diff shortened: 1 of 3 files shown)


libs/sdk-flutter/android/build.gradle.production
```diff
@@ -49,8 +55,9 @@ android {
 
 dependencies {
     api "breez_sdk:bindings-android:$version"
-    implementation "net.java.dev.jna:jna:5.14.0@aar"
+    implementation("net.java.dev.jna:jna:5.14.0@aar") {
+        exclude group: 'net.java.dev.jna', module: 'jna'
+    }
```
dangeross, line 60: Does this work? This should exclude transitive dependencies, which looks like you're trying to exclude itself. If anything the JNA should be excluded from the SDK bindings, the reason being jitpack doesn't include a complete JNA transitive dependency

erdemyerebasmaz: > If anything the JNA should be excluded from the SDK bindings, the reason being jitpack doesn't include a complete JNA transitive dependency

**Note:** I’ve only tested this using our SDK from our Maven repository, not via JitPack, so the transitive dependency resolution might differ there.

JssDWt (approved):
Tested with a brand new flutter project. All I had to do was bump the minSdk to 24.

libs/sdk-flutter/android/build.gradle.production
dangeross: Good to know. Thanks for the explanation

dangeross (approved):

PR #1207, opened 2025-04-21
ci(ios): unify iOS 12.0 deployment target across CI and podspecs

erdemyerebasmaz:
This PR is a continuation of 
- #1154 &
- #1206

to address missing architecture artifacts on iOS due target version being too low, as Rust require [Xcode 12 or higher to build the ARM64 targets](https://doc.rust-lang.org/rustc/platform-support/apple-ios.html#requirements).

This resolves online build & publishing Swift package issues, which consequently fix Flutter & React Native package issue as they depend on Swift package.


dangeross (approved):
LGTM

PR #1210, opened 2025-04-30
A fix to parse another varirant of cln error

### LINKED ISSUE #1209: No idea how to redeem BTC
Hello,

I've got my funds locked after the channel was inactive for too long.

Both with `inProgressSwap` and `listRefundables` I get an order which has status `reedemable`.

How do I redeem it?

I've both tried `redeemSwap(swapInfo.btcAddress)` and

```ts
await prepareRedeemOnchainFunds({
    toAddress: address,
    satPerVbyte: fees.hourFee,
  });

  const redeemOnChainFundsResponse = await redeemOnchainFunds({
    toAddress: address,
    satPerVbyte: fees.hourFee,
  });

  const txid = redeemOnChainFundsResponse.txid;
  // convert txid to hex representation
  const hexTxid = Buffer.from(txid).toString('hex');
```

But it fails.

`redeemSwap` fails with

Image

while the `redeemOnchainFunds` fails with 

Image

I have no clue what the correct procedure is.

For reference these are the locked funds:

https://mempool.space/tx/69ac87a96dfa270023b78de27a85f52d2c1285f901ea8c93fd10514f0217345a

Thank you in advance

roeierez:
Seems that there is another variant of cln error that we should parse.
Before this change since we didn't detect this type of error we didn't fail back to using the node invoice.

Fixed #1209


dangeross (approved):
LGTM

JssDWt (approved):
LGTM!

PR #1211, opened 2025-05-02
Remove std feature from lightning for Wasm

dangeross:
When validating the BOLT12 invoice request the lightning crate checks if the BOLT12 offer has expired which causes a panic in Wasm. Removing the `std` feature avoids this. It is our offer so we know it's valid.

### DIFF
--- a/libs/sdk-common/Cargo.toml
+++ b/libs/sdk-common/Cargo.toml
@@ -15,9 +15,6 @@ elements = { version = "0.25.0", optional = true }
 hex = { workspace = true }
 lazy_static = "1.5.0"
 lightning = { workspace = true }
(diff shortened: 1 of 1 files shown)


JssDWt:
You would need some kind of time provider to do the check, how does that work with no-std?

dangeross:
> You would need some kind of time provider to do the check, how does that work with no-std?

You can use the `web_time` crate which shims SystemTime with a javascript time alternative, then validate the expiry yourself

JssDWt:
> You can use the `web_time` crate which shims SystemTime with a javascript time alternative, then validate the expiry yourself

sounds good

danielgranhao (approved):

PR #1213, opened 2025-05-12
Refactor CLI app to handle `BreezServices` lifetime

andrei-21:
* Stop using the static variable since they are not dropped: https://doc.rust-lang.org/reference/items/static-items.html#r-items.static.lifetime
* Add an `abort` command to mimic crash behaviors (useful for testing)

### DIFF
--- a/tools/sdk-cli/src/command_handlers.rs
+++ b/tools/sdk-cli/src/command_handlers.rs
@@ -1,8 +1,9 @@
+use std::borrow::Cow::{self, Owned};
 use std::fs;
 use std::sync::Arc;
 use std::time::SystemTime;
 
-use anyhow::{anyhow, Context, Error, Result};
+use anyhow::{anyhow, ensure, Context, Error, Result};
 use breez_sdk_core::InputType::{LnUrlAuth, LnUrlPay, LnUrlWithdraw};
 use breez_sdk_core::{
     parse, BreezEvent, BreezServices, BuyBitcoinRequest, CheckMessageRequest, ConnectRequest,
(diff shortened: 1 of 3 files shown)


tools/sdk-cli/src/command_handlers.rs
```diff
@@ -68,549 +47,609 @@ impl Highlighter for CliHelper {
...
-        }
-        Commands::UnregisterWebhook { url } => {
-            sdk()?.unregister_webhook(url).await?;
-            Ok("Url unregistered successfully".into())
-        }
+    }
+
+    async fn connect(&mut self, req: ConnectRequest) -> Result<()> {
+        let service = BreezServices::connect(req, Box::new(CliEventListener {})).await?;
+        ensure!(self.sdk.is_none(), "Breez Services already initialized");
```
roeierez, line 643: Perhaps better to validate before we connect?

tools/sdk-cli/src/command_handlers.rs
```diff
@@ -68,549 +47,609 @@ impl Highlighter for CliHelper {
...
+            Commands::FetchFiatRates {} => {
+                serde_json::to_string_pretty(&self.sdk()?.fetch_fiat_rates().await?)
+                    .map_err(|e| e.into())
+            }
+            Commands::CloseLSPChannels {} => {
+                let tx_ids = self.sdk()?.close_lsp_channels().await?;
+                Ok(format!("Closing transaction ids:\n{tx_ids:?}"))
+            }
+            Commands::Disconnect {} => {
+                self.sdk()?.disconnect().await?;
```
roeierez, line 389: Now that it is not static anymore I think it would be nice to nullify the sdk instance allowing the user to re-connect again.

roeierez (approved):
LGTM!

JssDWt (approved):
LGTM!

PR #1226, opened 2025-06-05
Lock reqwest to v0.12.18

danielgranhao:
reqwest v0.12.19 introduced a breaking change on Wasm as it excluded the `Error::is_status()` method. Upstream issue: https://github.com/seanmonstar/reqwest/issues/2719

This is breaking our CI job where sdk-common is taken into a fresh Wasm project (deps not locked). This PR locks the reqwest version to 0.12.18 to fix CI.

### DIFF
--- a/libs/Cargo.toml
+++ b/libs/Cargo.toml
@@ -33,7 +33,8 @@ log = "0.4"
 once_cell = "1"
 prost = "^0.11"
 regex = "1.8.1"
-reqwest = { version = "0.12.13", features = ["json"] }
+# Pin reqwest until https://github.com/seanmonstar/reqwest/issues/2719 is solved
+reqwest = { version = "=0.12.18", features = ["json"] }
 rusqlite = { version = "0.29", features = [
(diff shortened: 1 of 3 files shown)


libs/Cargo.toml
```diff
@@ -33,7 +33,7 @@ log = "0.4"
 once_cell = "1"
 prost = "^0.11"
 regex = "1.8.1"
-reqwest = { version = "0.12.13", features = ["json"] }
+reqwest = { version = "=0.12.18", features = ["json"] }
```
JssDWt, line 36:
```suggestion
# Pin reqwest until https://github.com/seanmonstar/reqwest/issues/2719 is solved
reqwest = { version = "=0.12.18", features = ["json"] }
```

JssDWt (commented):
Looks good! Let's add a comment in the toml.

dangeross (approved):

JssDWt (approved):

PR #1227, opened 2025-06-05
Only attempt to parse BIP353 in Nodeless

dangeross:
This PR solves the issue in the Native/Greenlight SDK where if a LN address is registered with an offer (on Nodeless), then parsing the same address will first be parsed using BIP353, and as it's also a valid BIP353 address the returned BOLT12 Offer will not be recognized as a valid input.

Fixes https://github.com/breez/misty-breez/issues/562


dangeross:
> BIP353 may also return an LNURL Pay. Shouldn't we support that?

Maybe we should, but we currently do not which is causing the regression

roeierez (approved):
LGTM

PR #1239, opened 2025-07-08
Update RN package to 0.76.9

dangeross:
This PR:
- Migrates the React Native package to 0.76.9, but keeps RN bridge definition (vs. bridgeless JSI architecture)
- Rewrites the example app for 0.76.9

It has been tested using locally built SDK for Android/iOS. It still runs into the "no events on local built SDK" issue, so needs testing with a released package


danielgranhao (approved):
Must have been a pain... LGTM

PR #1241, opened 2025-07-14
Refactor node_api to use tokio `Stream` instead of tonic `Streaming`

andrei-21:
Such that gRPC internals of Greenlight implementation are not leaked to the interface.


JssDWt (changes requested):
Looks totally workable. Let's detach the interface from greenlight specifics some more by introducing our own types for the streams.

JssDWt (approved):

roeierez (approved):

PR #1245, opened 2025-07-17
Add LDK feature and stub implementation of NodeAPI

andrei-21:


libs/sdk-core/src/breez_services.rs
```diff
@@ -2413,6 +2412,19 @@ impl BreezServicesBuilder {
...
+                self.config.clone(),
+                &self.seed.clone().unwrap(),
+                restore_only,
+            )
+            .await;
+            node_api = Some(Arc::new(ldk));
+            if backup_transport.is_none() {
+                backup_transport = Some(Arc::new(LdkBackupTransport {}));
+            }
+        }
```
JssDWt, line 2427: This is using feature flags to dictate program flow. There are a few gaps I'd like to clear up here.

- There should be an option to remove the greenlight dependency if you're running LDK-only.
- Having ldk in the binary doesn't mean it has to be used. If the caller passes a Greenlight configuration object, you'd expect to go into a greenlight flow.
- The configuration enum should be extended with an LDK configuration.

How about instead of using the feature flags to determine program flow, we use the enum in the configuration. We add an LDK variant to the enum and match on it here. We could make the enum variants in the config tied to features. And make the enum non-exhaustive. Then the enum matches here will also get the feature flags and create the respective node. Return an error for the non-exhaustive case. This way you can enable greenlight-only, ldk-only, or both at the same time. Just if you enable none, you'll get a runtime error (we could make that a compile-time error).

JssDWt: I agree. I think we'll need 2 UDLs

JssDWt: That makes sense, yes.

Let's do that when we are confident we have a working LDK MVP. For now I think let's add `ldk` and `greenlight` feature flags where `greenlight` is enabled by default and ldk is not. And let's not worry about UDL bindings yet, given we can exclude things from compilation with the flags.

roeierez (approved):
LGTM

PR #1247, opened 2025-07-23
Add regtest environment type

andrei-21:


JssDWt:
I think it's best to hook into the `EnvironmentType`. Add a `Regtest` variant there. 

Let `CliConfig::default` take a network parameter: https://github.com/breez/breez-sdk/blob/221dc188a3ec2d8e84158aa9a1c863d779866f25/tools/sdk-cli/src/config.rs#L10-L17

Add a `Config::regtest` function to use in the match here: https://github.com/breez/breez-sdk/blob/860af986602ff14cbac546774c93ee1d38e77f87/libs/sdk-core/src/breez_services.rs#L1419-L1420

roeierez (approved):

PR #1251, opened 2025-08-09
Add `greenlight` as a default feature

andrei-21:


libs/sdk-core/src/node_builder.rs
```diff
@@ -0,0 +1,29 @@
...
+        let greenlight = Arc::new(greenlight);
+        let backup_transport = GLBackupTransport {
+            inner: greenlight.clone(),
+        };
+        let backup_transport = Arc::new(backup_transport);
+        Ok((greenlight, backup_transport))
+    }
+
+    #[cfg(not(feature = "greenlight"))]
+    compile_error!("Misconfigration: one feature must be enabled (greenlight).");
```
JssDWt, line 28: Instead of this compiler error, I would 
- Make the `NodeConfig` enum non_exhaustive
- go for a `match` on the `node_config`
- add a default case to the match statement where you return an error like 'unknown implementation'.

PR #1254, opened 2025-08-21
Add `LockingStore` and `TimeLock` for distributed locking mechanism

andrei-21:
Run unit tests with `cargo test --no-default-features --features ldk -- ldk`

### DIFF
--- a/libs/sdk-core/src/ldk/store/locking_store.rs
+++ b/libs/sdk-core/src/ldk/store/locking_store.rs
@@ -1,0 +1,213 @@
+use crate::ldk::store::time_lock::{LockData, LockedBy, TimeLock};
+use crate::ldk::store::versioned_store::{Error, VersionedStore};
+use std::time::{Duration, SystemTime};
+use tokio::sync::Mutex;
+use tonic::async_trait;
+
+struct VersionedTimeLock {
+    tl: TimeLock,
+    version: i64,
+}
+
+/// A wrapper around a `VersionedStore` that provides distributed locking capabilities.
+///
+/// `LockingStore` ensures that only one instance can access the underlying store at a time
+/// by implementing a distributed locking mechanism. This is useful in scenarios where multiple
+/// processes or instances might try to access the same storage concurrently.
+/// It provides *exclusive* access to the store, meaning no one else can acquire/release
+/// the lock while the `LockingStore` is managed.
+/// *Warning:* If the lock is somehow lost, `LockingStore` cannot recover,
+/// and all operations (except `unlock()`) will result in a `Conflict` error.
+///
+/// The locking mechanism works by:
+/// 1. Acquiring a lock when the store is created
+/// 2. Refreshing the lock periodically to maintain ownership
+/// 3. Releasing the lock when explicitly unlocked
+///
+/// If another instance already holds the lock, attempting to create a new `LockingStore`
+/// will result in a `Conflict` error.
+pub struct LockingStore<S: VersionedStore + Send + Sync> {
+    inner: S,
+    versioned_tl: Mutex<VersionedTimeLock>,
+}
+
+#[allow(dead_code)]
+impl<S: VersionedStore + Send + Sync> LockingStore<S> {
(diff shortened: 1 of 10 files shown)


libs/sdk-core/src/ldk/store/locking_store.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(())
+    }
+}
+
+/// Implementation of `VersionedStore` for `LockingStore`.
+///
+/// All store operations are wrapped with lock validation to ensure that
+/// only the instance holding the lock can perform operations on the underlying store.
+#[async_trait]
+impl<S: VersionedStore + Send + Sync> VersionedStore for LockingStore<S> {
```
roeierez, line 130: Curious why do we need this trait implementation? Isn't the intention that the app will use directly the LockingStore with lock/unlock/refresh_lock methods?

libs/sdk-core/src/ldk/store/locking_store.rs
```diff
@@ -0,0 +1,206 @@
...
+        let locked_until = self.lock().await?;
+        Ok(locked_until - Duration::from_secs(Self::REFRESH_WINDOW_SECS))
+    }
+
+    async fn lock(&self) -> Result<SystemTime, Error> {
+        let lock_data = self.tl.lock().await.next_lock();
+        let value = lock_data
+            .encode()
+            .map_err(|()| Error::Internal("Failed to encode lock_data".to_string()))?;
+        self.inner
```
roeierez, line 80: Should we also consider the case where  a lock exists with new version but it is expired? At this point shouldn't we fetch the new version and attempt to update from there?

libs/sdk-core/src/ldk/store/time_lock.rs
```diff
@@ -0,0 +1,357 @@
...
+    pub instance_id: String,
+}
+
+#[derive(Debug)]
+pub enum DecodeError {
+    InvalidUtf8(FromUtf8Error),
+    InvalidTimestamp(ParseIntError),
+}
+
+impl LockData {
```
roeierez, line 135: What is the advantage of using custom encode/decode over json for example?

roeierez (changes requested):
Looks good!

libs/sdk-core/src/ldk/store/locking_store.rs
roeierez: I see. IIUC the timelock will be created  when sdk connects. I am just worried if for example the app is going to the background without inet connection for some minutes/hours and meantime other instance acquired the lock and fail to release. In that case it seems we don't recover? This is not worrying you?

roeierez (approved):

roeierez:
@JssDWt can you add your review please?

libs/sdk-core/src/ldk/store/locking_store.rs
```diff
@@ -0,0 +1,213 @@
...
+                .map_err(|e| Error::Internal(format!("Failed to encode lock_data: {e}")))?;
+            self.inner
+                .put(Self::KEY.to_string(), value, versioned_tl.version)
+                .await?
+        }
+        Ok(())
+    }
+
+    async fn ensure_locked(&self) -> Result<(), Error> {
+        if !self.versioned_tl.lock().await.tl.is_locked() {
```
JssDWt, line 125: This is called for every get/put/delete/list. And `lock` will always `put` the lock on the remote storage. I think this can be much more time efficient if the lock is only refreshed when necessary, i.e. the time lock is about to expire (with the refresh window as leeway).

Also this now blocks parallel calls, because each call will fight for the same mutex to lock the database.

JssDWt (commented):
The additions are great. Added one more issue.

libs/sdk-core/src/ldk/store/locking_store.rs
JssDWt: Ah!

JssDWt (approved):
LGTM!

PR #1258, opened 2025-08-22
Build basic LDK Node

andrei-21:
Node can start and stop:
```
sdk> connect
Node was connected successfully
sdk> node_info
{
  "id": "",
  "block_height": 0,
  "channels_balance_msat": 0,
  "onchain_balance_msat": 0,
  "pending_onchain_balance_msat": 0,
  "utxos": [],
  "max_payable_msat": 0,
  "max_receivable_msat": 0,
  "max_single_payment_amount_msat": 0,
  "max_chan_reserve_msats": 0,
  "connected_peers": [],
  "max_receivable_single_payment_amount_msat": 0,
  "total_inbound_liquidity_msats": 0
}
sdk> disconnect
Node was stopped successfully
sdk>
Node was not initialized
```

### DIFF
--- a/libs/sdk-core/src/ldk/node_api.rs
+++ b/libs/sdk-core/src/ldk/node_api.rs
@@ -174,7 +174,7 @@ impl NodeAPI for Ldk {
         Err(NodeError::generic("LDK implementation not yet available"))
     }
 
-    async fn start_signer(&self, mut shutdown: mpsc::Receiver<()>) {
+    async fn start(&self, mut shutdown: mpsc::Receiver<()>) {
         debug!("Starting LDK Node");
         if let Err(e) = self.node.start() {
             error!("Failed to start LDK Node: {e}");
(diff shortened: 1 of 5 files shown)


libs/sdk-core/src/ldk/node_api.rs
```diff
@@ -127,92 +156,82 @@ impl NodeAPI for Ldk {
...
+            return;
+        }
+        debug!("LDK Node started");
+
+        _ = shutdown.recv().await;
+        debug!("Stopping LDK Node");
+        if let Err(e) = self.node.stop() {
+            error!("Error on stopping LDK Node: {e}");
+        }
+        debug!("LDK Node stopped");
```
roeierez, line 190: No external signer in ldk, does it sounds OK to wire the node start here?
Perhaps we use the start/stop from the breez services?
Or as an alternative perhaps just rename this trait function to "start" instead of "start_signer" which allows each node implementation adapt it to its own startup needs?

roeierez (commented):
Structure looks good and clear.

JssDWt:
Looking good

PR #1259, opened 2025-08-29
Add `LspMode` to `LspInformation`

andrei-21:
This information is needed for `PaymentReceiver` to differentiate behavior between Breez LSP mode and LSPS5.


libs/sdk-bindings/src/breez_sdk.udl
```diff
@@ -514,7 +514,13 @@ dictionary OpeningFeeParamsMenu {
     sequence<OpeningFeeParams> values;
 };
 
+enum LspMode {
+    "Breez",
+    "LSPS5",
+};
+
 dictionary LspInformation {
+    LspMode mode;
```
roeierez, line 523: Thinking out loud as I don't know yet how this is going to be used. Another alternative is to convert the LspInformation into enum that has either BreezInfo or LSPS5Info as I imagine they would contain different data.
Not an action item for this PR but when the usage is implemented it will become more clear on how to structure this.

PR #1260, opened 2025-09-02
Stream incoming payments

andrei-21:
Handle `PaymentReceived` event, convert to `IncomingPayment` and stream to the consumer

### DIFF
--- a/libs/sdk-core/src/ldk/node_api.rs
+++ b/libs/sdk-core/src/ldk/node_api.rs
@@ -14,11 +14,14 @@ use rand::Rng;
 use sdk_common::ensure_sdk;
 use sdk_common::prelude::Network;
 use serde_json::Value;
-use tokio::sync::{mpsc, watch};
-use tokio_stream::Stream;
+use tokio::sync::{broadcast, mpsc, watch};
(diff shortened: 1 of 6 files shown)


libs/sdk-core/src/ldk/node_api.rs
```diff
@@ -209,8 +224,15 @@ impl NodeAPI for Ldk {
...
-        // TODO: Implement.
-        Ok(Box::pin(futures::stream::pending()))
+        let incoming_payments_rx =
+            self.incoming_payments_rx
+                .lock()
+                .await
+                .take()
+                .ok_or(NodeError::generic(
+                    "Incoming payments stream has already started",
+                ))?;
```
JssDWt, line 234: What do you think about simplifying this by using a broadcast channel? No need for the option and the mutex on incoming_payments_rx. You would be able to subscribe multiple times if you want.

JssDWt (commented):
Looking good. Just one comment.

JssDWt (approved):
Let's try this. I see how you can miss the first events from the node now, because of the race in the beginning. But I think it's ok since it's only payment received events.

roeierez (approved):
LGTM

PR #1267, opened 2025-09-11
Implement mirroring store

andrei-21:


roeierez:
@andrei-21  can you elaborate on where/how this should be used? I think it is better to review after it is integrated it into the sdk.

JssDWt:
> It depends how we reconcile. If two instances are dirty, then the second will download from the remote and override the local state, see the branch:

I see now, thanks

libs/sdk-core/src/ldk/store/mirroring_store.rs
```diff
@@ -0,0 +1,429 @@
...
+        let parts: Vec<&str> = full_key.splitn(3, '/').collect();
+        let (primary, secondary, key) = match &parts[..] {
+            [p, s, k] => (p.to_string(), s.to_string(), k.to_string()),
+            _ => continue, // skip malformed keys
+        };
+
+        if let Some((value, version)) = remote.get(full_key).await? {
+            trace!("Got {} bytes @ {version}", value.len());
+            conn.execute(
+                "INSERT INTO store (primary_ns, secondary_ns, key, value, local_version, remote_version) VALUES (?1, ?2, ?3, ?4, ?5, ?5)",
```
roeierez, line 223: Shouldn't we only insert if remote version is larger than local? Do you have any concern of deleting all items and just copy the remote?

libs/sdk-core/src/ldk/store/mirroring_store.rs
```diff
@@ -0,0 +1,429 @@
...
+                info!("Local store is *dirty*, nothing new on remote. Uploading to remote...");
+                upload(conn, &*remote).await?;
+            }
+            (PreviousHolder::RemoteInstance, false) => {
+                info!("Local store is clean, something new on remote possible. Downloading from remote...");
+                download(conn, &*remote).await?;
+            }
+            (PreviousHolder::RemoteInstance, true) => {
+                info!("Local store is *dirty*, something new on remote possible. Downloading from remote...");
+                download(conn, &*remote).await?;
```
roeierez, line 84: In this case don't we loose local changes?

roeierez (commented):
Looks good. dropped some questions.

PR #1269, opened 2025-09-16
Build `LockingStore` and start refreshing the lock

andrei-21:


libs/sdk-core/src/ldk/store_builder.rs
```diff
@@ -0,0 +1,102 @@
...
+                    match locking_store.unlock().await {
+                        Ok(()) => info!("Remote lock was released"),
+                        Err(e) => error!("Failed to release remote lock: {e}"),
+                    };
+                    break;
+                },
+                }
+            }
+            Err(e) => warn!("Failed to refresh remote lock: {e:?}"),
+        }
```
JssDWt, line 98: A nesting suggestion. May be ignored, because it's a preference.

```suggestion
        let until = match locking_store.refresh_lock().await {
            Ok(until) => until,
            Err(e) => {
                warn!("Failed to refresh remote lock: {e:?}");
                // TODO: sleep?
                continue;
            }
        };

        trace!("Remote lock was refreshed");
        let duration = until.duration_since(SystemTime::now()).unwrap_or_default();
        tokio::select! {
            _ = tokio::time::sleep(duration) => (),
            _ = shutdown_rx.recv() => {
                info!("Releasing remote lock");
                match locking_store.unlock().await {
                    Ok(()) => info!("Remote lock was released"),
                    Err(e) => error!("Failed to release remote lock: {e}"),
                };
                break;
            }
        }
```

libs/sdk-core/src/ldk/store_builder.rs
```diff
@@ -0,0 +1,102 @@
...
+                    info!("Releasing remote lock");
+                    match locking_store.unlock().await {
+                        Ok(()) => info!("Remote lock was released"),
+                        Err(e) => error!("Failed to release remote lock: {e}"),
+                    };
+                    break;
+                },
+                }
+            }
+            Err(e) => warn!("Failed to refresh remote lock: {e:?}"),
```
JssDWt, line 97: This needs a little sleep to avoid retrying this instantly.

PR #1270, opened 2025-09-16
Compute node state

andrei-21:


libs/sdk-core/src/ldk/node_state.rs
```diff
@@ -0,0 +1,65 @@
...
+            .list_peers()
+            .iter()
+            .filter(|p| p.is_connected)
+            .map(|p| p.node_id.to_string())
+            .collect();
+
+        let channels = node.list_channels();
+        let max_payable_msat = channels
+            .iter()
+            .map(|c| c.next_outbound_htlc_limit_msat)
```
JssDWt, line 26: I love this field `next_outbound_htlc_limit_msat`

libs/sdk-core/src/ldk/node_state.rs
```diff
@@ -0,0 +1,65 @@
...
+            .sum();
+        let inbound_capacity_msats = channels.iter().map(|c| c.inbound_capacity_msat).sum();
+
+        Self {
+            id: node.node_id().to_string(),
+            block_height: node.status().current_best_block.height,
+            channels_balance_msat: balances.total_lightning_balance_sats * 1000,
+            onchain_balance_msat: balances.total_onchain_balance_sats * 1000,
+            pending_onchain_balance_msat: pending_onchain_balance_sats * 1000,
+            utxos: Vec::new(), // Not available in LDK Node.
```
JssDWt, line 40: Where do onchain funds from channel cloures go?

JssDWt (approved):

libs/sdk-core/src/ldk/node_state.rs
roeierez: I think we can start without these utxos and see where it puts us. From what I know we don't require them inside the sdk, Let's just verify that @andrei-21 
@JssDWt unless you are aware of some flow they are needed.

roeierez (approved):
LGTM. Left one comment.

libs/sdk-core/src/ldk/node_state.rs
JssDWt: I am also unaware of such flows.

PR #1273, opened 2025-09-30
Add esplora, vss, rgs, lsp endpoints to config

andrei-21:


JssDWt:
Apart from the failing CI checks, this looks good.

